PR #25, opened 2023-04-10
Cmake & minor fixes

klemens-morgenstern:


mrdox.yml
```diff
@@ -0,0 +1,4 @@
+project-name: MrDox
+namespaces:
+  include:
+    - ::N
```
vinniefalco, line 4: Now this is interesting! There's missing newlines by the way. And if you are going to add this mrdox.yml please put it in its own folder.

CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 #
 #-------------------------------------------------
 
-cmake_minimum_required(VERSION 3.25.2)
+cmake_minimum_required(VERSION 3.21)
```
vinniefalco, line 17: This seems ok

source/lib/Config.cpp
```diff
@@ -102,7 +103,7 @@ loadFromFile(
     auto fileText = llvm::MemoryBuffer::getFile(configPath);
     if(! fileText)
     {
-        R.failed("llvm::MemoryBuffer::getFile", fileText);
+        R.failed("llvm::MemoryBuffer::getFile", configPath);
```
vinniefalco, line 106: There's two "include" file filters now?

include/mrdox/Corpus.hpp
```diff
@@ -118,8 +118,8 @@ class Corpus
 
     /** Build the intermediate representation of the code being documented.
     */
-    static
     [[nodiscard]]
+    static
```
vinniefalco, line 122: A fine change

source/lib/Serialize.cpp
```diff
@@ -478,7 +478,7 @@ populateParentNamespaces(
             InfoType::IT_namespace);
 }
 
-void PopulateTemplateParameters(std::optional<TemplateInfo>& TemplateInfo,
+void PopulateTemplateParameters(llvm::Optional<TemplateInfo>& TemplateInfo,
```
vinniefalco, line 481: Why don't we just replace it everywhere?

include/mrdox/Generator.hpp
```diff
@@ -66,7 +66,8 @@ struct Generator
         llvm::StringRef rootPath,
         Corpus const& corpus,
         Config const& config,
-        Reporter& R) const;
+        Reporter& R,
+        const std::string & fileName) const;
```
vinniefalco, line 70: The `Reporter` always comes last in parameter lists but why are you changing the API of `Generator`?

source/tool/ToolMain.cpp
```diff
@@ -147,7 +155,7 @@ toolMain(
 
     // Run the generator.
     llvm::outs() << "Generating docs...\n";
-    if(! gen->build(config.OutDirectory, *rv, config, R))
+    if(! gen->build(config.OutDirectory, *rv, config, R, Filename))
```
vinniefalco, line 158: This is not right though, because the non-XML generators emit multi-file output

klemens-morgenstern: does the non-xml have a primary file I can use? I need some target for cmake.

vinniefalco:
I merged some of these commits, please rebase

source/lib/Serialize.cpp
klemens-morgenstern: I did.

vinniefalco:
Is this obsolete?

PR #27, opened 2023-04-10
Multi-config CI

alandefreitas:
Changes in this PR:

- `ci.yml`: remove GCC <=10 and Clang <=14 because `<source_location>` is not available so the library doesn't support them anymore. All tests are passing again.
- `CMakeLists.txt`:
    - `D_ITERATOR_DEBUG_LEVEL` matches the value of LLVM for multi-configurations. LLVM already uses the default debug level for MSVC, so this wouldn't be necessary in the general case. But a Debug build depending on LLVM compiled with RelWithDebInfo breaks it because the defaults are different. This change will check whatever debug level LLVM is using and apply it to the MrDox library.
    - Minimum version required is set to 3.13 since the most recent feature we use is from 3.13. Setting the `minimum_required` to the latest version instead of the real _minimum_ value required creates unnecessary errors and makes CI slower since now we need to needlessly install a more recent version of CMake. 
    - Returns with the `MRDOX_BUILD_TESTS` option. The convention everyone uses for its default value is [`BUILD_TESTING`](https://cmake.org/cmake/help/latest/module/CTest.html) but I just set it to `ON` so the VS workflow is not broken. This means, of course, anyone building it will build the tests unless they know we changed that convention.
    - Global options were moved to target options so they won't leak to targets where we don't want or can't have them.
- `Corpus.hpp` and `Error.cpp`: fixed small errors that broke GCC.
- `CMakeSettings.json.in`: see below

The current MSVC workflow described in `NOTES.adoc` creates a single multiconfig solution from CMake:

```mermaid
flowchart LR
    CMake[CMakeLists.txt] -->|configure| S[VS Solution]
    S -->|builds| Debug
    S -->|builds| DebugWithRel
    S -->|builds| RelWithDebInfo
    S -->|builds| ...
```

The main problem with that is LLVM-Config.cmake, like most libraries, does not support being imported into multi-config solutions with redirection of targets to the corresponding build type. Thus, all MrDox targets will attempt to link to the same LLVM configuration, which will either fail for conflicting definitions or at least not achieve what we wanted.

A trivial solution is to generate multiple solutions linking to different LLVM configurations and ignore any other potential configurations in the generated solution:

```mermaid
flowchart LR
    CMake[CMakeLists.txt] -->|configure with LLVM Debug| S1[VS Solution - Debug] -->|builds| Debug
    S1 -->|ignore| D1[...]
    CMake[CMakeLists.txt] -->|configure with LLVM RelWithDebInfo| S2[VS Solution - DebugWithRel] -->|builds| DebugWithRel
    S2 -->|ignore| D2[...]
    CMake[CMakeLists.txt] -->|configure with LLVM RelWithDebInfo| S3[VS Solution - RelWithDebInfo] -->|builds| RelWithDebInfo
    S3 -->|ignore| D3[...]
    CMake[CMakeLists.txt] -->|configure with LLVM ...| S4[VS Solution - ...] -->|builds| ...
    S4 -->|ignore| D4[...]
  
```

This works fine and the change in `CMakeLists.txt` to infer the proper debug level should also work fine. 

Some might consider this unnecessarily complex and it makes the workflow a little painful. Another solution is describing these effectively single-configuration solutions in a CMakeSettings.json file in VS and opening the CMake folder directly in VS, which will just do everything for us.

```mermaid
flowchart LR
    CMake[CMakeLists.txt] -->|VS configures it as in CMakeSettings.json and builds| Debug
    CMake[CMakeLists.txt] -->|VS configures it as in CMakeSettings.json and builds| DebugWithRel
    CMake[CMakeLists.txt] -->|VS configures it as in CMakeSettings.json and builds| RelWithDebInfo
    CMake[CMakeLists.txt] -->|VS configures it as in CMakeSettings.json and builds| ...

```

`CMakeSettings.json` describes exactly where each configuration should look for LLVM and will end up doing the same thing under the hood since they are all still using the Visual Studio 17 2022 generator. `CMakeSettings.json.in` includes a template to build the targets with multiple LLVM configurations in VS. It also allows for more configuration variants than a single solution.

`CMakeSettings.json.in` was left as a template to avoid breaking the current workflow. Renaming it `CMakeSettings.json` and adjusting `cmakeCommandArgs` to the correct `llvm-install` should already work. Other options that can be used in this file are [here](https://learn.microsoft.com/en-us/cpp/build/cmakesettings-reference?view=msvc-170).

`CMakeSettings.json` is part of the first VS push towards CMake. There's actually a second push where they [recommend](https://learn.microsoft.com/en-us/cpp/build/customize-cmake-settings?view=msvc-170) to customize the CMake via [`CMakePresets.json`](https://learn.microsoft.com/en-us/cpp/build/cmake-presets-json-reference?view=msvc-170) files. `CMakePresets.json` is more powerful and works outside VS, but it requires some investigation and we don't need any of its functionality at the moment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,12 +23,17 @@ jobs:
         include:
           - { name: "GCC 12 - C++20",    os: ubuntu-22.04, cc: gcc-12, cxx: g++-12, cxxstd: '20', install: g++-12 }
           - { name: "GCC 11 - C++20",    os: ubuntu-22.04, cc: gcc-11, cxx: g++-11, cxxstd: 20, install: g++-11 }
-          - { name: "GCC 10 - C++20",    os: ubuntu-22.04, cc: gcc-10, cxx: g++-10, cxxstd: 20, install: g++-10 }
-          - { name: "GCC 9  - C++20",    os: ubuntu-22.04, cc: gcc-9,  cxx: g++-9,  cxxstd: 20, install: g++-9 }
 
-          - { name: "Clang 14 - C++20",  os: ubuntu-22.04, cc: clang-14, cxx: clang++-14, cxxstd: 20, install: clang-14 }
-          - { name: "Clang 13 - C++20",  os: ubuntu-22.04, cc: clang-13, cxx: clang++-13, cxxstd: 20, install: clang-13 }
-          - { name: "Clang 12 - C++20",  os: ubuntu-22.04, cc: clang-12, cxx: clang++-12, cxxstd: 20, install: clang-12 }
+          # no <source_location>
+          # - { name: "GCC 10 - C++20",    os: ubuntu-22.04, cc: gcc-10, cxx: g++-10, cxxstd: 20, install: g++-10 }
+          # - { name: "GCC 9  - C++20",    os: ubuntu-22.04, cc: gcc-9,  cxx: g++-9,  cxxstd: 20, install: g++-9 }
+
+          - { name: "Clang 15 - C++20",  os: ubuntu-22.04, cc: clang-15, cxx: clang++-15, cxxstd: 20, install: clang-15 }
+
+          # no <source_location>
+          # - { name: "Clang 14 - C++20",  os: ubuntu-22.04, cc: clang-14, cxx: clang++-14, cxxstd: 20, install: clang-14 }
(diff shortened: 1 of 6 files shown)


klemens-morgenstern (approved):
lgtm as long as everything passes

PR #95, opened 2023-04-24
Add generate.py to generate enum declarations into tests/py/dcl.enum/

Ukilele:
As we discussed it in [cpplang.slack](https://cpplang.slack.com/archives/C21PKDHSL/p1682283694044249), here is the script to generate C++ declarations. So far it only generates enum declarations. The files will be written to _file-dir_/tests/py/dcl.enum/*.cpp, where _file-dir_ is the directory of the script generate.py. The script uses the python module [exrex](https://github.com/asciimoo/exrex), so please make sure to install it.

Feel free to add a code review or require other changes. In the meantime, I will look into further declarations and how to generate them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ temp/
 test/
 tests/**/*.adoc
 tests/**/*.bad.xml
+tests/py/
--- a/tests/generate.py
+++ b/tests/generate.py
@@ -1,0 +1,35 @@
+import exrex
+import os
+
+def ToplevelFolder():
+    scriptPath = os.path.realpath(__file__)
+    parentDirectory = os.path.dirname(scriptPath)
+    return os.path.join(parentDirectory, "py")
+
+def EnumDeclarationsFolder():
+    return os.path.join(ToplevelFolder(), "dcl.enum")
+
+
+def GenerateEnumDeclarations():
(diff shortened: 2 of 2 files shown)


vinniefalco:
clang does not like those lambda expressions one single bit :)

We need to fix these https://gist.github.com/vinniefalco/af93abfbc947dfb0ecacc2b530befaf4

Ukilele:
Whaaaat? But they are so common in C++...
Joke aside, I will remove them. But just to be clear, clang >= 5.0.0 supports them, if you add "-std=c++2a" (or "-std=c++20" if using a more recent version of clang). If you let me know which compilers/C++ standards we are aiming at, I can take that into account and try to avoid producing invalid or too recent declarations.

vinniefalco:
oh hmm.... well maybe my tests are broken! I am not setting the C++ standard version... I guess that would happen here? https://github.com/cppalliance/mrdox/blob/433c78d52000cdeb4c0dd730985e8de6a57096f0/source/tests/SingleFile.hpp#L35 You have to add those args somehow.

I am targeting the very latest C++ by the way (C++20 I guess, or even C++23 if that's possible).

vinniefalco:
The develop branch uses C++20 for tests now

Issue #58, opened 2023-04-17
Add plugins/generators via DLL


vinniefalco:
It should be possible for MrDox to load generators at runtime via DLL "plugins", so that authors of new formats do not have to build LLVM in order to write and test them.

There's no rush for this.

vinniefalco:
We have to attempt to make the metadata API such that generators do not need to include llvm headers, this will make it possible to write generators without installing and building clang/llvm.

alandefreitas:
For reference: https://github.com/cppalliance/mrdocs/issues/1113#issuecomment-3667556008

PR #98, opened 2023-04-25
has shared link option

### LINKED ISSUE #93: Build options for shared library
The CML needs to offer a shared library option, and it should be possible to build the test and tool executables using dynamic linking. And we can test those configurations in the actions.

klemens-morgenstern:
Closes #93.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,16 @@ project(
 #set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELWITHDEBINFO ON CACHE STRING "")
 
 option(MRDOX_BUILD_TESTS "Build tests" ON)
+option(MRDOX_BUILD_SHARED "Link shared" OFF)
+
+if (MRDOX_BUILD_SHARED)
+    set(MRDOX_LINK_MODE SHARED)
+    set(MRDOX_LINK_MODE_DEFINITION -DMRDOX_SHARED_LINK)
+else()
+    set(MRDOX_LINK_MODE_DEFINITION -DMRDOX_STATIC_LINK)
+endif()
+
+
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -34,6 +34,16 @@ project(
 #set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELWITHDEBINFO ON CACHE STRING "")
 
 option(MRDOX_BUILD_TESTS "Build tests" ON)
+option(MRDOX_BUILD_SHARED "Link shared" OFF)
+
+if (MRDOX_BUILD_SHARED)
```
vinniefalco, line 39: is this how its usually done with cmake? I mean, is there not some kind of standard switch or something?

vinniefalco:
I think maybe I did not write the issue up correctly... MrDox has to have a plugin system (i.e. load DLLs / shared libraries dynamically at runtime) so that people can author their own generators without having to build MrDox (and LLVM with it). Or at the least so they don't need to fork the program. 

For this to work they need to be able to call in to MrDox functions and I am assuming this has to happen via shared linking...

klemens-morgenstern:
#58 is the one.

vinniefalco:
> That's a different issue altogether I think. I'll look into it.

Yes, building the plugin architecture is a issue number 58, but this issue is about setting up the build system for it (and having it tested). Most of this pull request looks useful; we need those macros properly set for non-windows platforms, and we need the switches in the CML..

When mrdox is installed from a package the tool should come as a single executable not as an executable with `main` plus the mrdox-api as a shared DLL. But it should still have exports or whatever, so it can load the plugins at runtime, and so those plugins can import from the executable and call into mrdox.

For the tests, who cares. Maybe we should make the mrdox-test link to mrdox dynamically so that CI can test it?

PR #100, opened 2023-04-25
add_fail_test function

### LINKED ISSUE #29: fail-tests (and warnings)
Eventually mrdox will have to emit errors of some kind for example if a javadoc is malformed or if there are duplicate javadoc definitions. For this we will need fail-tests.

We might also want warnings, such as when parameters are not documented in the source inputs.

klemens-morgenstern:
Closes #29

### DIFF
--- a/tests/FailTest.cmake
+++ b/tests/FailTest.cmake
@@ -1,0 +1,37 @@
+if(WIN32)
+    set(MRDOX_FT_RUNNER ${CMAKE_CURRENT_LIST_DIR}/fail_test.bat)
+else()
+    set(MRDOX_FT_RUNNER ${CMAKE_CURRENT_LIST_DIR}/fail_test.sh)
+endif()
+
+
+function(add_fail_test)
+    cmake_parse_arguments(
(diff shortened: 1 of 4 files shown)


tests/FailTest.cmake
```diff
@@ -0,0 +1,37 @@
+if(WIN32)
+    set(MRDOX_FT_RUNNER ${CMAKE_CURRENT_LIST_DIR}/fail_test.bat)
+else()
+    set(MRDOX_FT_RUNNER ${CMAKE_CURRENT_LIST_DIR}/fail_test.sh)
```
vinniefalco, line 4: dafuq is "CMAKE_CURRENT_LIST_DIR" ? what is a listfile?

klemens-morgenstern: The directory of the current file

vinniefalco:
Is that a merge commit?

vinniefalco:
This needs to be rebased, and the scripts going to the right place (tools/ or scripts/)

PR #105, opened 2023-04-25
Extend python test generation

Ukilele:
Adds:

- empty declaration (`;`)
- namespace, nested namespace, inline namespace
- a broad range of class definitions a la class.pre
- a small set of non member functions


vinniefalco:
I merged this, but I also refactored the layout of the directories in the repo. The script is in tools/generate.py while the test files are in test-files/py. Could you please submit a patch to allow a target directory to be passed to generate.py as a command line argument? I would use it this way:
```
$ cd tools
$ ./generate.py ../test-files
```

Thanks! And.. that's quite a lot of output files :)

PR #110, opened 2023-04-26
Properly render C++20 and C++23 in one line in asciidoc

Ukilele:
Right now the readme says:

> This includes up to C20 and even experimental things in C23

This is due the fact that content between double pluses gets rendered differently. I have zero knowledge about asciiDoctor, but according to https://github.com/asciidoctor/asciidoctor/issues/1864, using `{cpp}` is the according fix.

The "C++" in line 42 gets rendered correctly. But for the sake of consistency, I also replaced it.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -39,10 +39,10 @@ cmake --install .
 
 === MrDox:
 
-* narrow and deep focus on C++ only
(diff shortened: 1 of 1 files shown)


vinniefalco:
image

PR #112, opened 2023-04-26
mrdox-test supports yml configs.

klemens-morgenstern:
This supports multiple documents per config (because yaml) so one config can create multiple runs with a single yaml file.


source/test/Config.cpp
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) 2023 Klemens D. Morgenstern
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
```
vinniefalco, line 4: It seems I have to repeat myself. We are not using the Boost license, Please follow the example set in the other files.

source/test/Config.cpp
```diff
@@ -0,0 +1,68 @@
...
+                return makeError(ec.message(), " when parsing file '", filePath, "' ");
+        }
+        while (yin.nextDocument());
+        return res;
+    }
+    return std::vector<TestConfig>{TestConfig{}};
+}
+
+}
+}
```
vinniefalco, line 68: Missing newline

source/test/Config.hpp
```diff
@@ -0,0 +1,33 @@
...
+{
+  std::string cxxstd = "c++20";
+  std::vector<std::string> compile_flags;
+  bool should_fail = false;
+  std::string heuristics = "unit test";
+};
+
+llvm::Expected<std::vector<TestConfig>>
+           loadTestConfig(llvm::StringRef dir,
+                          llvm::StringRef file);
```
vinniefalco, line 27: Please follow the established style

source/test/Config.cpp
```diff
@@ -0,0 +1,68 @@
...
+    {
+        filePath = dir;
+        filePath += "/mrdox-test.yml";
+    }
+
+    if (fs::exists(filePath))
+    {
+        auto fileText = llvm::MemoryBuffer::getFile(filePath);
+        if(! fileText)
+            return makeError(fileText.getError().message(), " when loading file '", filePath, "' ");
```
vinniefalco, line 51: I believe you can just put `fileText` here instead of `fileText.getError().message()`. If not, we can add a `nice` overload for it.

source/test/Config.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <llvm/Support/YAMLParser.h>
+#include <llvm/Support/YAMLTraits.h>
+
+
+template<>
+struct llvm::yaml::MappingTraits<
+    clang::mrdox::TestConfig>
+{
+  static void mapping(IO &io,
+                      clang::mrdox::TestConfig& f)
```
vinniefalco, line 19: It would be best if we did not have two files with the same name anywhere, we already have Config.cpp in api

source/test/SingleFile.hpp
```diff
@@ -25,22 +26,29 @@ class SingleFile
...
+        llvm::StringRef file,
+        const TestConfig &tc)
     {
         std::vector<std::string> cmds;
         cmds.emplace_back("clang");
-        cmds.emplace_back("-std=c++20");
+        {
+          char buf[64];
+          snprintf(buf, 63, "-std=%s", tc.cxxstd.c_str());
+          cmds.emplace_back(buf);
```
vinniefalco, line 40: please use the existing style

source/test/Config.cpp
klemens-morgenstern: Sorry, IDO template

test-files/old-tests/attributes_1.yml
```diff
@@ -0,0 +1,4 @@
+cxxstd: c++20
+...
+---
+cxxstd: c++17
```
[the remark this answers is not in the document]
klemens-morgenstern: That's not a new yaml format, that's standard yaml.

source/test/Config.cpp
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    if (fs::exists(filePath))
+    {
+        auto fileText = llvm::MemoryBuffer::getFile(filePath);
+        if(! fileText)
+            return makeError(fileText.getError().message(), " when loading file '", filePath, "' ");
+        llvm::yaml::Input yin(**fileText);
+
+        std::vector<TestConfig> res;
```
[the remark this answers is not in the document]
klemens-morgenstern: The latter part does happen. Allowing multiple configs per yml doesn't cost us much, but allows us to quickly add more tests. This is particularly helpful is a certain functionality doesn't work with a given flag OR on a standard. 

An issue reporter can then just add his case for an existing test.

source/test/SingleFile.hpp
```diff
@@ -15,6 +15,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include "TestConfig.hpp"
```
vinniefalco, line 18: includes should be ordered most specific -> most general

source/test/TestConfig.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <mrdox/Error.hpp>
+#include <llvm/Support/FileSystem.h>
+#include <llvm/Support/Path.h>
+#include <llvm/Support/YAMLParser.h>
+#include <llvm/Support/YAMLTraits.h>
+
+
+template<>
+struct llvm::yaml::MappingTraits<
+    clang::mrdox::TestConfig>
```
vinniefalco, line 19: Wait a minute I think there's some confusion here, the tests are supposed to use the same YML configuration as the tool. Not a whole new YML configuration.

source/test/SingleFile.hpp
```diff
@@ -25,22 +26,29 @@ class SingleFile
...
-        llvm::StringRef file)
+        llvm::StringRef file,
+        const TestConfig &tc)
     {
         std::vector<std::string> cmds;
         cmds.emplace_back("clang");
-        cmds.emplace_back("-std=c++20");
+        {
+            char buf[64];
+            snprintf(buf, 63, "-std=%s", tc.cxxstd.c_str());
```
vinniefalco, line 39: hmm this will create warnings on msvc.. prefer using `llvm::raw_string_ostream`

source/test/TestConfig.cpp
klemens-morgenstern: It's a different config object.

klemens-morgenstern:
@Vinnie rebased.

test-files/old-tests/attributes_1.yml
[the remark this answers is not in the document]
vinniefalco: Yes I see that but still.... where are you

PR #119, opened 2023-04-27
Add explicit specifier

Ukilele:
Emit `<explicit/>` when the `explicit` specifier is present at a constructor, deduction guide or conversion operator.

As discussed in slack, we will for now always emit `<explicit/>`, no matter if you write `explicit(false)` or `explicit(expr)`.

### DIFF
--- a/test-files/old-tests/explicit-conv-operator.cpp
+++ b/test-files/old-tests/explicit-conv-operator.cpp
@@ -1,0 +1,18 @@
+//Test conversion operators that can be "explicit", "explicit(true)", "explicit(false)" and "explicit(EXPR)"
+
+struct Explicit {
+    explicit operator bool();
+};
+
+struct ExplicitFalse {
+    explicit(false) operator bool();
+};
+
+struct ExplicitTrue {
+    explicit(true) operator bool();
+};
+
+template<bool B>
+struct ExplicitExpression {
+    explicit(B) operator bool();
+};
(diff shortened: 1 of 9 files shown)


test-files/old-tests/explicit-conv-operator.cpp
```diff
@@ -0,0 +1,18 @@
...
+};
+
+struct ExplicitTrue {
+    explicit(true) operator bool();
+};
+
+template<bool B>
+struct ExplicitExpression {
+    explicit(B) operator bool();
+};
```
vinniefalco, line 18: These each need to be in their own individual cpp file eventually

test-files/old-tests/explicit-conv-operator.xml
```diff
@@ -0,0 +1,38 @@
...
+      <file path="explicit-conv-operator.cpp" line="8"/>
+      <explicit/>
+      <return name="_Bool"/>
+    </function>
+  </struct>
+  <struct name="Explicit" id="bonobNKGOblPVcRjxpiPan4nYnc=">
+    <file path="explicit-conv-operator.cpp" line="3" class="def"/>
+    <function name="operator bool" access="public" id="64zY8rVu+TzEUayzMrnV65i9nFc=">
+      <file path="explicit-conv-operator.cpp" line="4"/>
+      <explicit/>
```
vinniefalco, line 17: nice, lol

PR #122, opened 2023-04-27
llvm version gets checked.

### LINKED ISSUE #115: check version of compiled LLVM
mrdox should emit a compilation error if the version of LLVM is too old At a minimum we should require LLVM-17.

klemens-morgenstern:
Closes #115.


include/mrdox/Platform.hpp
```diff
@@ -11,6 +11,8 @@
 #ifndef MRDOX_PLATFORM_HPP
 #define MRDOX_PLATFORM_HPP
 
+#include <llvm/Config/llvm-config.h>
```
vinniefalco, line 14: I rather not do this here, it is my hope and dream that the public API which is needed for `Generator` does not require LLVM.

include/mrdox/Platform.hpp
```diff
@@ -19,6 +21,15 @@
 namespace clang {
 namespace mrdox {
 
+//------------------------------------------------
+//
+// Check llvm version
+//
+//------------------------------------------------
+
+static_assert(LLVM_VERSION_MAJOR >= 15,
```
vinniefalco, line 30: Lets put this in Config.cpp

vinniefalco, line 30: Put this in Platform.hpp (it doesn't require llvm includes)
```
#define MRDOX_MINIMUM_LLVM_VERSION 15
```

And then use that in Config.cpp to do the do

vinniefalco:
Any word on this?

vinniefalco:
merged thanks!

PR #157, opened 2023-05-04
Fix function parameter types

### LINKED ISSUE #142: Ambiguity resolution for names should account for type decay
```
void f(int x);

void f(const int x)
{
}
```

Both declarations of `f` refer to the same function. mrdox produces the following output:

```
<?xml version="1.0" encoding="UTF-8"?>
<mrdox xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xsi:noNamespaceSchemaLocation="https://github.com/cppalliance/mrdox/raw/develop/mrdox.rnc">
<namespace name="">
  <function name="f" id="ackjGh57CJQLwtO8PihiSPPKIBY=">
    <file path="input.cpp" line="1"/>
    <param name="x" type="int"/>
  </function>
  <function name="f" id="6i4Aiw+mGDmAJKzUMLtNaLS8qjs=">
    <file path="input.cpp" line="3" class="def"/>
    <param name="x" type="const int"/>
  </function>
</namespace>
</mrdox>
```

Although the parameter types as declared are different, the determination of whether two function declarations refer to the same function only considers the parameter types *after* decaying array & function to pointers and removing top-level cv-qualifiers ; see http://eel.is/c++draft/basic.scope.scope#4 & http://eel.is/c++draft/dcl.fct#5

sdkrystian:
Closes #142

### DIFF
--- a/source/api/AST/ASTVisitor.cpp
+++ b/source/api/AST/ASTVisitor.cpp
@@ -272,8 +272,10 @@ parseParameters(
 {
     for(ParmVarDecl const* P : D->parameters())
     {
+        // KRYSTIAN NOTE: call getOriginalType if we want to preserve
+        // top-level cv-qualfiers/array types/function types
         FieldTypeInfo& FieldInfo = I.Params.emplace_back(
-            getTypeInfoForType(P->getOriginalType()),
+            getTypeInfoForType(P->getType()),
             P->getNameAsString());
         FieldInfo.DefaultValue = getSourceCode(
             D, P->getDefaultArgRange());
--- a/source/api/AST/ASTVisitor.cpp
+++ b/source/api/AST/ASTVisitor.cpp
@@ -1292,10 +1292,17 @@ ASTVisitor::
 WalkUpFromParmVarDecl(
     ParmVarDecl* D)
 {
-    // We do nothing here, to prevent ParmVarDecl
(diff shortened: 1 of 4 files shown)


sdkrystian:
Tests coming shortly...

source/api/AST/ASTVisitor.cpp
```diff
@@ -1292,10 +1294,17 @@ ASTVisitor::
 WalkUpFromParmVarDecl(
     ParmVarDecl* D)
 {
-    // We do nothing here, to prevent ParmVarDecl
-    // from appearing as VarDecl. We pick up the
-    // function parameters as a group from the
-    // FunctionDecl instead of visiting ParmVarDecl.
+    Assert(shouldTraversePostOrder() &&
+        "expected post-order traversal");
```
vinniefalco, line 1298: oh man I hate this "text" in the assert

sdkrystian: How about "modifying a ParamVarDecl here depends on post-order traversal"?

vinniefalco: I dont want any text at all, it is gratuituous because the assert will probably never  go off so we are paying for a string for nothing. I edited the code already to remove the text and just leave a comment.

PR #163, opened 2023-05-05
in/out @param javadoc

sdkrystian:


include/mrdox/Metadata/Javadoc.hpp
```diff
@@ -65,6 +65,18 @@ struct MRDOX_VISIBLE
         warning
     };
 
+    /** Parameter pass direction.
+    */
+    enum class ParamDirection : int
+    {
+        // KRYSTIAN NOTE: no idea why this is needed,
+        // does not seem to break when changed to 0
+        none = 1, // needed by bitstream
```
klemens-morgenstern, line 74: does that mean unspecified?

source/api/_XML/XMLWriter.cpp
```diff
@@ -615,8 +615,25 @@ XMLWriter::
...
+    switch(param.direction)
+    {
+    case Javadoc::ParamDirection::in:
+        direction = "in";
+        break;
+    case Javadoc::ParamDirection::out:
+        direction = "out";
+        break;
+    case Javadoc::ParamDirection::inout:
+        direction = "in,out";
```
klemens-morgenstern, line 628: I think `inout` is preferrable

source/api/_XML/XMLWriter.cpp
```diff
@@ -615,8 +615,25 @@ XMLWriter::
...
+        break;
+    default:
+        break;
+    }
     tags_.open("param", {
-        { "name", param.name, ! param.name.empty() }});
+        { "name", param.name, ! param.name.empty() },
+        { "direction", direction, !! direction }
```
klemens-morgenstern, line 635: I'd prefer `direction != direction::none`.

test-files/old-tests/param-direction.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+/** test brief
+ * @param x
+ * @param[in] y
+**/
+void g(int x, int y) { }
+
+/**
+ * @param x
+ * @param[in,out] y
```
klemens-morgenstern, line 14: Doesn't clang handle `@param[inout]` ?

klemens-morgenstern (changes requested):

include/mrdox/Metadata/Javadoc.hpp
vinniefalco: I think it does and I think we can use 0 here

test-files/old-tests/param-direction.xml
```diff
@@ -0,0 +1,76 @@
...
+  </function>
+  <function name="i" id="RmOJ/T+Nr/W5+22oXdFiYeoIDoA=">
+    <file path="param-direction.cpp" line="23" class="def"/>
+    <param name="x" type="int"/>
+    <param name="y" type="int"/>
+    <param name="z" type="int"/>
+    <doc>
+      <brief>
+      </brief>
+      <param name="x" direction="out">
```
vinniefalco, line 30: `class="out"`

test-files/old-tests/param-direction.cpp
sdkrystian: @vinniefalco @klemens-morgenstern it does not

source/lib/api/AST/ParseJavadoc.cpp
```diff
@@ -334,6 +334,21 @@ class JavadocVisitor
             param.name = ensureUTF8(C->getParamNameAsWritten().str());
         else
             param.name = "@anon";
```
vinniefalco, line 336: "@anon" ?

test-files/old-tests/param-direction.cpp
```diff
@@ -0,0 +1,72 @@
+/**
+ * @param x0
```
vinniefalco, line 2: When there are tests for javadocs, they should be strictly javadoc tests (I think yours are) and they should have a parent directory in the chain which is strictly javadoc tests

source/lib/api/AST/ParseJavadoc.cpp
sdkrystian: It's used for `@param` directives without any parameter name provided... only place I could see it being useful is for variadic function a la `void f(...)`.

source/api/_XML/XMLWriter.cpp
vinniefalco: Same.. !! ... ugh

PR #169, opened 2023-05-08
Tests print diffs on error, so CI is more informative.

klemens-morgenstern:


source/test/TestMain.cpp
```diff
@@ -271,15 +272,27 @@ handleFile(
...
+                    std::error_code ec;
+                    llvm::raw_fd_ostream os(bad, ec, llvm::sys::fs::OF_None);
+                    if (ec) {
+                        results_.numberOfErrors++;
+                        return makeError("raw_fd_ostream returned ", ec);
+                    }
+                    os << generatedXml;
+                }
+
+                auto diff = llvm::sys::findProgramByName("diff");
```
vinniefalco, line 287: Interesting :)

image

PR #211, opened 2023-05-24
ASTVisitor work

sdkrystian:


source/AST/ASTVisitor.cpp
```diff
@@ -46,12 +135,14 @@ ASTVisitor::
 ASTVisitor(
     tooling::ExecutionContext& ex,
     ConfigImpl const& config,
-    Reporter& R) noexcept
+    Reporter& R,
+    clang::CompilerInstance& compiler) noexcept
```
vinniefalco, line 139: Put the reporter after the CompilerInstance

include/mrdox/Metadata/Typedef.hpp
```diff
@@ -30,15 +33,29 @@ struct TypedefInfo : SymbolInfo
...
+    explicit
     TypedefInfo(
         SymbolID id_ = SymbolID::zero)
         : SymbolInfo(InfoType::IT_typedef, id_)
     {
     }
+
+private:
+    // KRYSTIAN NOTE: if Template is non-null,
+    // then IsUsing *must* be set; should we do it here?
```
vinniefalco, line 51: Maybe `IsUsing` should be a private data member and we would provide a public member function `isAlias()`

include/mrdox/Metadata/Template.hpp
```diff
@@ -14,12 +14,183 @@
...
-#include <mrdox/Metadata/TemplateParam.hpp>
+#include <mrdox/Metadata/Type.hpp>
+#include <string>
 
 namespace clang {
 namespace mrdox {
 
+enum class TParamKind : int
+{
+    None = 0, // for bitstream
```
vinniefalco, line 25: is "None" actually used or is this an invalid state?

sdkrystian: It's an intermediate state used when building when building the `TParam` (during AST traversal & deserialization)

sdkrystian:
@vinniefalco most recent commit add the declaration of `SpecializationInfo` you wanted to see last week

include/mrdox/Metadata/Template.hpp
vinniefalco: is it ever valid for a `Generator` to see `None` for a TParamKind ?

include/mrdox/Metadata/Template.hpp
```diff
@@ -191,6 +191,36 @@ struct TArg
...
+// member of an implicitly instantiated class template specialization.
+// this structure is stored in the TemplateInfo corresponding to the
+// outermost specialized template. if the explicitly specialized
+// member is itself a member of a nested template, the SpecializationInfo
+// which stores the arguments for the parent template will additionally
+// store a pointer to the SpecializationInfo for the nested template,
+// recursively. each SpecializationInfo node contains a vector of
+// SymbolIDs pairs `(specialized, primary)`, where `specialized` is the
+// replacement definition of `primary` for the given set of template
+// arguments of its parent template(s).
```
vinniefalco, line 204: ugh this is a really ugly comment can we please use /* */ for multiline exposition and also this should be a javadoc if it is going to be in a public API:
```
/** Metadata for explicit specializations of certain instantiated class template members

    This stores information pertaining to an explicit specialization of a
    member of an implicitly instantiated class template specialization.
    this structure is stored in the `TemplateInfo` corresponding to the
    outermost specialized template. if the explicitly specialized
    member is itself a member of a nested template, the `SpecializationInfo`
    which stores the arguments for the parent template will additionally
    store a pointer to the `SpecializationInfo` for the nested template,
    recursively. each `SpecializationInfo` node contains a vector of
    `SymbolID` pairs (specialized, primary), where `specialized` is the
    replacement definition of `primary` for the given set of template
    arguments of its parent template(s).
*/
struct SpecializationInfo;
```

include/mrdox/ADT/Optional.hpp
```diff
@@ -83,6 +106,16 @@ class Optional
...
     }
 
+    constexpr value_type* operator->() noexcept
+    {
+        return std::addressof(t_);
+    }
+
+    constexpr value_type const* operator->() const noexcept
+    {
+        return std::addressof(t_);
```
vinniefalco, line 116: Why do we need `addressof`?

include/mrdox/Metadata/Function.hpp
```diff
@@ -126,6 +130,15 @@ struct FunctionInfo
         : SymbolInfo(InfoType::IT_function, id_)
     {
     }
+
+private:
+    explicit
+    FunctionInfo(
+        std::unique_ptr<TemplateInfo>&& T)
```
vinniefalco, line 137: `noexcept`?

include/mrdox/Metadata/Record.hpp
```diff
@@ -19,10 +19,11 @@
 #include <mrdox/Metadata/Reference.hpp>
 #include <mrdox/Metadata/Scope.hpp>
 #include <mrdox/Metadata/Symbol.hpp>
-#include <mrdox/Metadata/Template.hpp>
 #include <mrdox/Metadata/Symbols.hpp>
+#include <mrdox/Metadata/Template.hpp>
 #include <clang/AST/Type.h>
 #include <llvm/ADT/SmallVector.h>
+#include <memory>
 #include <optional>
```
vinniefalco, line 27: is this include still needed?

include/mrdox/Metadata/Template.hpp
vinniefalco: If None is invalid then it must not be in the enumeration

include/mrdox/Metadata/Template.hpp
```diff
@@ -14,39 +14,254 @@
...
 
+enum class TParamKind : int
+{
+    None = 0, // for bitstream
+    // template type parameter, e.g. "typename T" or "class T"
+    Type,
+    // template non-type parameter, e.g. "int N" or "auto N"
+    NonType,
+    // template template parameter, e.g. "template<typename> typename T"
+    Template,
```
vinniefalco, line 34: Out of Type, NonType, or Template, figure out which one is statistically more likely to appear and make that the "zero" value.

include/mrdox/Metadata/Template.hpp
```diff
@@ -14,39 +14,254 @@
...
+};
+
+struct TParam;
+
+struct TypeTParam
+{
+    /** Default type for the type template parameter */
+    std::optional<TypeInfo> Default;
```
vinniefalco, line 42: unique_ptr?

sdkrystian: This should remain as an optional since `TypeInfo` is just a `Reference`. We can implement `OptionalTypeInfo` using `Optional` later on

include/mrdox/Metadata/Template.hpp
```diff
@@ -14,39 +14,254 @@
...
+                std::forward<T>(other).Variant_.Template);
+            break;
+        default:
+            break;
+        }
+    }
+
+public:
+    /** The kind of template parameter this is. */
+    TParamKind Kind = TParamKind::None;
```
vinniefalco, line 103: is TParamKind::None a valid value?

include/mrdox/Metadata/Template.hpp
```diff
@@ -14,39 +14,254 @@
...
-    // for primary templates:
-    //     - Params will be non-empty
-    //     - Args will be empty
-    // for explicit specializations:
-    //     - Params will be empty
-    // for partial specializations:
-    //     - Params will be non-empty
-    //     - each template parameter will appear at least
-    //       once in Args outside of a non-deduced context
+    /** For primary templates:
```
vinniefalco, line 242: The brief is different from the description. Add a proper brief and reformat the description to be a list (use `@li` for list items)

include/mrdox/Metadata/Template.hpp
```diff
@@ -14,39 +14,254 @@
...
+           - each template parameter will appear at least
+             once in Args outside of a non-deduced context
+    */
     std::vector<TParam> Params;
     std::vector<TArg> Args;
 
-    // stores the ID of the corresponding primary template
-    // for partial and explicit specializations
+    /** Stores the ID of the corresponding primary template
+        for partial and explicit specializations
```
vinniefalco, line 256: "Stores" wtf? No.. How about this
```
/** Primary template ID for partial and explicit specializations.

    If this is disengaged there is no primary template ID.
*/
```

include/mrdox/Metadata/Template.hpp
```diff
@@ -14,39 +14,254 @@
...
+    */
     OptionalSymbolID Primary;
 
+    /** Stores information for explicit specializations of members
+        of implicitly instantiated class template specializations
```
vinniefalco, line 261: this needs to be better. brief must be short and never on two lines.

include/mrdox/Metadata/Typedef.hpp
```diff
@@ -31,15 +36,29 @@ struct TypedefInfo
     //   typedef std::vector<int> MyVector;
     bool IsUsing = false;
 
+    std::unique_ptr<TemplateInfo> Template;
```
vinniefalco, line 39: is it just me or can almost everything be a template? Should we move this data member to a base?

include/mrdox/Metadata/Var.hpp
```diff
@@ -14,8 +15,10 @@
 #include <mrdox/Platform.hpp>
 #include <mrdox/Metadata/Field.hpp>
 #include <mrdox/Metadata/Symbol.hpp>
+#include <mrdox/Metadata/Template.hpp>
 #include <mrdox/Metadata/Type.hpp>
 #include <clang/Basic/Specifiers.h>
```
vinniefalco, line 20: Do we need this clang include still?

source/-XML/XMLWriter.cpp
```diff
@@ -386,7 +387,8 @@ writeRecord(
 
     writeSymbol(I);
 
-    writeTemplate(I.Template);
+    if(I.Template)
+        writeTemplate(*I.Template);
```
vinniefalco, line 391: consider if you want to do the emptiness check inside `writeTemplate`. For example:
```
void writeTemplate( std::unique_ptr<TemplateInfo> const& );
```

This makes the call site look nice and tidy

source/AST/ASTVisitor.cpp
```diff
@@ -46,12 +48,14 @@ ASTVisitor::
...
     ConfigImpl const& config,
+    clang::CompilerInstance& compiler,
     Reporter& R) noexcept
     : ex_(ex)
     , config_(config)
     , R_(R)
     , PublicOnly(! config_.includePrivate)
     , IsFileInRootDir(true)
+    , compiler_(compiler)
 {
```
vinniefalco, line 59: These warnings need fixing

include/mrdox/Metadata/Typedef.hpp
sdkrystian: Enums cannot be templates

source/AST/ASTVisitor.cpp
```diff
@@ -65,40 +69,49 @@ ASTVisitor(
...
+    id = SymbolID(llvm::SHA1::hash(
+        arrayRefFromStringRef(usr_)).data());
+    return true;
 }
 
-//------------------------------------------------
+SymbolID
+ASTVisitor::
+extractSymbolID(
+    const Decl* D)
```
vinniefalco, line 89: this is a weird overload..

source/AST/ASTVisitor.cpp
```diff
@@ -65,40 +69,49 @@ ASTVisitor(
...
+    return id;
+}
 
 bool
 ASTVisitor::
 shouldSerializeInfo(
     bool PublicOnly,
-    bool IsInAnonymousNamespace,
-    NamedDecl const* D) noexcept
+    bool IsOrIsInAnonymousNamespace,
```
vinniefalco, line 100: Is Or Is ?

source/AST/ASTVisitor.cpp
```diff
@@ -107,145 +120,90 @@ shouldSerializeInfo(
...
             Namespaces.emplace_back(
-                getSymbolID(N),
+                extractSymbolID(N),
                 N->getNameAsString(),
-                InfoType::IT_function);
+                InfoType::IT_enum);
         }
-        else if(auto const* N = dyn_cast<EnumDecl>(DC))
+        else if(const auto* N = dyn_cast<TranslationUnitDecl>(DC))
         {
```
vinniefalco, line 184: Fix the warning or disable it

source/AST/ASTVisitor.cpp
sdkrystian: "Is" or "IsIn"

source/AST/ASTVisitor.cpp
```diff
@@ -1083,12 +1005,30 @@ buildFriend(
...
             if(! constructFunction(I, FD))
                 return;
+#if 0
             SymbolID id;
             getParent(id, D);
-            RecordInfo P(id);
+#else
+            const DeclContext* DC = D->getDeclContext();
+            const auto* RD = dyn_cast<CXXRecordDecl>(DC);
+            Assert(RD && "the semantic DeclContext of a FriendDecl must be a class");
```
vinniefalco, line 1014: Please do not add strings to asserts. Instead use:
```
// The semantic DeclContext of a FriendDecl must be a class
Assert( RD );
```

source/-XML/XMLWriter.cpp
sdkrystian: This is going to be changed once implicit specializations are implemented fully, so I'd like to just leave it as-is for now

source/AST/ASTVisitor.cpp
```diff
@@ -1582,58 +1461,251 @@ TraverseFunctionTemplateDecl(
...
-    switch(FD->getKind())
+bool
+ASTVisitor::
+Traverse(
+    ClassScopeFunctionSpecializationDecl* D)
+{
+    if(! shouldExtract(D))
+        return true;
+
+    // for class scope explicit specializations of member function templates which
```
vinniefalco, line 1479: I would use /* */ to allow easier editing. And I would use proper grammar and capitalization since it is so much exposition.

source/AST/ASTVisitor.cpp
```diff
@@ -1582,58 +1461,251 @@ TraverseFunctionTemplateDecl(
...
+    auto Template = std::make_unique<TemplateInfo>();
+    parseTemplateParams(*Template, AD);
+
+    return Traverse(AD, std::move(Template));
+}
+
+template<typename... Args>
+auto
+ASTVisitor::
+Traverse(Args&&...)
```
vinniefalco, line 1514: you could =delete this instead and save the unreachable theatrics.

You might even make it not a template and instead write
```
auto Traverse(...) = delete
```

source/AST/ASTVisitor.cpp
```diff
@@ -1582,58 +1461,251 @@ TraverseFunctionTemplateDecl(
...
+        this->Traverse(static_cast<
+            CXXRecordDecl*>(D),
+            std::forward<Args>(args)...);
         break;
     case Decl::CXXMethod:
-        buildFunction(I, static_cast<
-            CXXMethodDecl*>(FD));
+        this->Traverse(static_cast<
+            CXXMethodDecl*>(D),
+            std::forward<Args>(args)...);
```
vinniefalco, line 1547: Unlike my other projects I'm ok with this entire statement being on one line. Up to you.

source/AST/ASTVisitor.cpp
```diff
@@ -1582,58 +1461,251 @@ TraverseFunctionTemplateDecl(
...
+    std::optional<llvm::StringRef> filePath =
+        Context.getSourceManager().getNonBuiltinFilenameForID(
+            Context.getSourceManager().getMainFileID());
+    if(! filePath)
+        return;
+
+    // Filter out TUs we don't care about
+    File = *filePath;
+    convert_to_slash(File);
+    if(! config_.shouldVisitTU(File))
```
vinniefalco, line 1693: broseph I do not think this is okay !

source/AST/ASTVisitor.hpp
```diff
@@ -50,6 +50,8 @@ class ASTVisitor
     ConfigImpl const& config_;
     Reporter& R_;
 
+    clang::CompilerInstance& compiler_;
+
     llvm::SmallString<512> File;
```
vinniefalco, line 55: Fix this warning

source/AST/ASTVisitor.hpp
```diff
@@ -219,38 +225,61 @@ class ASTVisitor
...
+    bool Traverse(TypeAliasDecl* D,
+        std::unique_ptr<TemplateInfo>&& Template = nullptr);
+    bool Traverse(VarDecl* D,
+        std::unique_ptr<TemplateInfo>&& Template = nullptr);
+    bool Traverse(FunctionDecl* D,
+        std::unique_ptr<TemplateInfo>&& Template = nullptr);
+    bool Traverse(CXXMethodDecl* D,
+        std::unique_ptr<TemplateInfo>&& Template = nullptr);
+    bool Traverse(CXXConstructorDecl* D,
+        std::unique_ptr<TemplateInfo>&& Template = nullptr);
```
vinniefalco, line 243: remove the parameter names and format all these on one line

include/mrdox/Metadata/Template.hpp
sdkrystian: Yes, it is used to indicate that the type of template parameter has not been determined. It ensures that we don't call the wrong destructor in case of an error when reading bitcode.

include/mrdox/Metadata/Record.hpp
sdkrystian: nope

source/AST/ASTVisitor.cpp
sdkrystian: No can do, that would cause a hard error in `TraverseDecl`. The catch-all function has to be selectable by overload resolution without causing an error (a deleted definition makes the program ill-formed if it is referenced)

source/AST/ASTVisitor.cpp
sdkrystian: Could you elaborate? I just moved this function -- this code was not changed by me

include/mrdox/Metadata/Template.hpp
sdkrystian: "disengaged"?

include/mrdox/Metadata/Template.hpp
sdkrystian: I'm going to change this to be a non-javadoc comment as it does not apply to a single data member.

include/mrdox/Metadata/Function.hpp
sdkrystian: I'm hesitant to add it here because its dependent upon whether any of the default constructors for the non-static data members of `FunctionInfo` throw. Your call

source/AST/ASTVisitor.cpp
vinniefalco: Nevermind... I thought it was not thread-safe.

source/AST/ASTVisitor.cpp
vinniefalco: boooo

source/AST/ASTVisitor.cpp
vinniefalco: What does "Is Or Is In" mean

sdkrystian: _is_ this declaration an anonymous namespace or _is_ this declaration _in_ an anonymous namespace?

PR #213, opened 2023-05-25
Refactor members & function parameters

sdkrystian:


include/mrdox/MetadataFwd.hpp
```diff
@@ -39,7 +39,7 @@ struct MemberTypeInfo;
 struct NamespaceInfo;
 struct RecordInfo;
 struct RecordScope;
-struct RefWithAccess;
+struct MemberRef;
```
vinniefalco, line 41: Alphabetize

source/AST/DecodeRecord.hpp
```diff
@@ -99,20 +99,20 @@ decodeRecord(
...
 inline
 llvm::Error
 decodeRecord(
     Record const& R,
-    std::vector<RefWithAccess>& f,
+    std::vector<MemberRef>& f,
     llvm::StringRef blob)
 {
-    constexpr auto RefWithAccessSize =
+    constexpr auto MemberRefSize =
```
vinniefalco, line 110: I would have just called it `Size` but w/e ...

include/mrdox/Metadata/FieldType.hpp
```diff
@@ -31,7 +31,7 @@ union FieldFlags
 };
 
 // Info for field types.
-struct FieldTypeInfo
+struct FieldInfo
```
vinniefalco, line 42: What is a `FieldInfo`? Can we please add a doc comment to this struct...

source/AST/DecodeRecord.hpp
sdkrystian: I just used replace all to change the names :) I don't mind what the name is either way

include/mrdox/Metadata/Function.hpp
```diff
@@ -71,12 +71,39 @@ union FnFlags1
     BitField<6, 7, FunctionKind> functionKind;
 };
 
+// KRYSTIAN TODO: attributes (nodiscard, deprecated, and carries_dependency)
+// KRYSTIAN TODO: flag to indicate whether this is a function parameter pack
```
vinniefalco, line 76: If a function parameter list can have only one parameter pack, then maybe we should store the index of the pack in the `FunctionInfo` instead of putting a bool on each and every `Param` - to optimize for space.

However, if we have to add flags to parameters then the calculus changes, once we have one mandatory flag on a Param then we can add another bool for `isParameterPack` basically for free.

include/mrdox/Metadata/Function.hpp
```diff
@@ -71,12 +71,39 @@ union FnFlags1
...
+/** Represents a single function parameter */
+struct Param
+{
+    /** The type of this parameter */
+    TypeInfo Type;
+
+    /** The name of this parameter, if any */
```
vinniefalco, line 83: /** The parameter name.

    Unnamed parameters are represented by empty strings.
*/

include/mrdox/Metadata/Function.hpp
```diff
@@ -71,12 +71,39 @@ union FnFlags1
...
+    }
+};
+
 // TODO: Expand to allow for documenting templating and default args.
 // Info for functions.
 struct FunctionInfo : SymbolInfo
 {
     TypeInfo ReturnType;   // Info about the return type of this function.
-    llvm::SmallVector<FieldTypeInfo, 4> Params; // List of parameters.
+    std::vector<Param> Params; // List of parameters.
```
vinniefalco, line 107: Good riddance to `SmallVector`

include/mrdox/MetadataFwd.hpp
```diff
@@ -30,7 +30,8 @@ enum class Access;
 struct BaseInfo;
 struct EnumValueInfo;
 struct EnumInfo;
-struct FieldTypeInfo;
+struct FieldInfo;
+struct Param;
```
vinniefalco, line 33: Alphabetize

include/mrdox/Metadata/FieldType.hpp
sdkrystian: I think you are looking at an outdated version of the PR... `Field.hpp` has the new declaration & specifies that `FieldInfo` is for non-static data memebers

source/-adoc/AdocWriter.cpp
```diff
@@ -106,7 +106,7 @@ namespace adoc {
 
 struct AdocWriter::FormalParam
 {
-    FieldTypeInfo const& I;
+    Param const& I;
```
vinniefalco, line 109: well this should be `P` not `I`?

source/llvm.natvis
```diff
@@ -187,14 +187,6 @@ For later versions of Visual Studio, no setup is required.
     <DisplayString>{Data}</DisplayString>
   </Type>
 
-  <Type Name="llvm::Optional&lt;*&gt;">
-    <DisplayString Condition="!Storage.hasVal">None</DisplayString>
-    <DisplayString Condition="Storage.hasVal">{Storage.value}</DisplayString>
```
vinniefalco, line 192: LOL don't change the LLVM natvis file, it is copied from the llvm repo. In fact we should remove the natvis file from our repo, and change the CMakeLists.txt to reference the copy of the natvis file that is in the llvm installation.

sdkrystian: oh LOL my bad

source/Metadata/Reduce.cpp
```diff
@@ -95,18 +96,28 @@ void mergeInfo(Info& I, Info&& Other)
...
 {
     Assert(canMerge(I, Other));
     if (!I.DefLoc)
         I.DefLoc = std::move(Other.DefLoc);
+    mergeInfo(I, std::move(Other));
+}
+
+static void mergeSymbolInfo(
+    SymbolRedeclInfo& I,
+    SymbolRedeclInfo&& Other)
```
vinniefalco, line 111: This function has the wrong name

PR #218, opened 2023-05-26
ci: all workflows support libxml2

alandefreitas:
This PR enables the `xml-lint` cmake test target that depends on LibXml2 tools and Java in all msvc, gcc, and clang workflows.


alandefreitas:
@sdkrystian Could you have a look at what I'm doing wrong here? The `xml-lint` test is not passing and I already rebased it on top of `develop`.

sdkrystian:
The schema fell behind. #230 updates it

PR #222, opened 2023-05-29
added check for rangs & format

klemens-morgenstern:


alandefreitas:
So...

The action is using these binaries for Clang 16.0.2: https://github.com/llvm/llvm-project/releases/download/llvmorg-16.0.2/clang+llvm-16.0.2-x86_64-linux-gnu-ubuntu-22.04.tar.xz

It comes with the corresponding libstdc++ which does have <format> and <ranges>. Everything works fine just with clang:
https://github.com/alandefreitas/mrdox/actions/runs/5113712752/jobs/9193229777

The latest Clang version is actually 16.0.4. If you set `check-latest: ${{ matrix.compiler == 'clang' }}` in the `setup-cpp` step, the action will download https://github.com/llvm/llvm-project/releases/download/llvmorg-16.0.4/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04.tar.xz instead, the latest in the `^16` semver range. It all works fine with 16.0.4 too https://github.com/alandefreitas/mrdox/actions/runs/5113792257/jobs/9193379591.

All the errors here are coming from the -stdlib=libc++ relationship, which I'm not even sure makes a lot of sense for this project given its relationship with LLVM.

vinniefalco:
Are we going to fix this or what?

vinniefalco:
was this merged?

klemens-morgenstern:
This would be obsolete due to the usage of fmt.

PR #225, opened 2023-05-29
ci: setup fmt

alandefreitas:


klemens-morgenstern:
definitely the better solution at this point in time.

alandefreitas:
> 228 is not ok, but after you merge the other 3 PRs and work on 228 to do an actual integration into mrdox (I will help) then it will be ok

PR #241, opened 2023-06-03
Adoc work

vinniefalco:


alandefreitas (approved):
Considering everything we already discussed in parallel, this looks good.

I would just make sure the test command can find handlebars.js so we don't break all tests.

PR #267, opened 2023-06-07
fix unintended fallthrough in javadoc post-processing

sdkrystian:


source/Metadata/Javadoc.cpp
```diff
@@ -94,6 +94,8 @@ postProcess()
                     std::move(static_cast<Returns &>(*it)));
                 it = blocks_.erase(it);
```
vinniefalco, line 95: well you need `continue;` here as well or else the comment below is wrong

PR #285, opened 2023-06-12
thread and command line work [with fixes]

klemens-morgenstern:


CMakeLists.txt
```diff
@@ -125,7 +125,7 @@ find_package(fmt REQUIRED CONFIG)
 configure_file(source/Version.hpp.in "${PROJECT_SOURCE_DIR}/include/mrdox/Version.hpp")
 file(GLOB_RECURSE SOURCES CONFIGURE_DEPENDS source/*.cpp source/*.hpp source/*.in source/*.natvis)
 file(GLOB_RECURSE INCLUDES CONFIGURE_DEPENDS include/*.hpp)
-
+file(COPY addons DESTINATION ${CMAKE_BINARY_DIR})
```
vinniefalco, line 128: We should not be copying the addons directory to the location of the executable via CMake.

You should do this only when we are building the release package (i.e. the installer).

source/Tool/ConfigImpl.cpp
```diff
@@ -78,11 +78,11 @@ ConfigImpl(
     namespace path = llvm::sys::path;
 
     if(! files::isAbsolute(workingDir_))
-        throw Error("path \"{}\" is not absolute", workingDir_);
+        throw Error("working path \"{}\" is not absolute", workingDir_);
     workingDir = files::makeDirsy(files::normalizePath(workingDir_));
 
     if(auto err = files::requireDirectory(addonsDir_))
-        throw Error("path \"{}\" is not absolute", addonsDir_);
+        throw Error("addons path \"{}\" is not absolute", addonsDir_);
```
vinniefalco, line 85: I don't understand...how does this fix a runtime errr?

source/Tool/ConfigImpl.hpp
```diff
@@ -255,7 +255,7 @@ loadConfigString(
     std::string_view addonsDir,
     std::string_view configYaml)
 {
-    return createConfigFromYAML(workingDir, configYaml, "");
+    return createConfigFromYAML(workingDir, addonsDir, configYaml);
```
vinniefalco, line 258: oops yeah

source/Tool/TestAction.cpp
```diff
@@ -145,6 +145,8 @@ makeConfig(
     std::error_code ec;
     auto config = loadConfigString(
         workingDir, toolArgs.addonsDir, configYaml);
+    if (!config)
+      reportError(config.getError().message());
```
vinniefalco, line 149: This should read thusly:
```
reportError( config.getError(), "load the configuration string" );
```

source/Tool/ConfigImpl.cpp
klemens-morgenstern: it helps me figure out which path is failing.

include/mrdox/Support/ExecutorGroup.hpp
```diff
@@ -0,0 +1,165 @@
...
+namespace mrdox {
+
+/** A set of execution agents for performing concurrent work.
+*/
+template<class Agent>
+class ExecutorGroup
+{
+    struct Impl
+    {
+        std::mutex mutex_;
```
klemens-morgenstern, line 32: this seems like an application for a shared_mutex to me. Optimization ofc.

klemens-morgenstern: I find the lock usage in this class confusing, but correct. Hard to read and see correctness though.

include/mrdox/Support/Thread.hpp
```diff
@@ -33,6 +38,74 @@ class TaskGroup;
...
+};
+
+//------------------------------------------------
+
+template<class>
+class any_callable;
+
+template<class R, class... Args>
+class any_callable<R(Args...)>
+{
```
klemens-morgenstern, line 66: what's wrong with std::function? Copyable requirement?

include/mrdox/Support/ThreadPool.hpp
```diff
@@ -0,0 +1,176 @@
...
+void
+ThreadPool::
+forEach(
+    Range&& range,
+    F const& f)
+{
+    TaskGroup taskGroup(*this);
+    for(auto&& value : range)
+        taskGroup.async(
+            [&f, &value]
```
klemens-morgenstern, line 166: Technically incorrect, since `auto && ` may mean it's a temporary when using some smart things like transforming ranges.

include/mrdox/Support/any_callable.hpp
```diff
@@ -0,0 +1,78 @@
...
+                return f_(std::forward<Args>(args)...);
+            }
+        };
+
+        p_ = std::make_unique<impl>(std::forward<Callable>(f));
+    }
+
+    R operator()(Args&&...args) const
+    {
+        return p_->invoke(std::forward<Args>(args)...);
```
klemens-morgenstern, line 71:
```suggestion
       Assert(p_); // make sure it wasn't moved from.
        return p_->invoke(std::forward<Args>(args)...);
```

source/-bitcode/BitcodeGenerator.cpp
```diff
@@ -14,7 +14,7 @@
 #include "Support/SafeNames.hpp"
 #include "AST/Bitcode.hpp"
 #include <mrdox/Support/Report.hpp>
-#include <mrdox/Support/Thread.hpp>
+#include <mrdox/Support/ThreadPool.hpp>
 #include <mrdox/Metadata.hpp>
```
klemens-morgenstern, line 19: Why are we mixing include styles?

source/Support/Thread.cpp
```diff
@@ -57,17 +57,23 @@ getThreadCount() const noexcept
...
 
 void
 ThreadPool::
-wait()
+post(
+    any_callable<void(void)> f)
 {
-    impl_->wait();
+    auto sp = std::make_shared<
+        any_callable<void(void)>>(std::move(f));
```
klemens-morgenstern, line 71: Can't be moved?

source/Support/Thread.cpp
```diff
@@ -57,17 +57,23 @@ getThreadCount() const noexcept
...
+    impl_->async(
+        [sp]
```
klemens-morgenstern, line 73:
```suggestion
        [sp=std::move(sp)]
```

Move, don't copy.

source/Support/Thread.cpp
```diff
@@ -90,17 +96,23 @@ TaskGroup(
...
 TaskGroup::
-wait()
+post(
+    any_callable<void(void)> f)
 {
-    impl_->wait();
+    auto sp = std::make_shared<
+        any_callable<void(void)>>(std::move(f));
+    impl_->async(
+        [sp]
```
klemens-morgenstern, line 112:
```suggestion
        [sp=std::move(sp)]
```

klemens-morgenstern (commented):
@vinniefalco please comment.

source/Support/Thread.cpp
vinniefalco: Nope. We have to reimplement `llvm::ThreadPool` to allow moving. Until then, this is the acceptable hack.

source/-bitcode/BitcodeGenerator.cpp
klemens-morgenstern: Yeah I figured that out. Headers in the source file didn't click immediately.

vinniefalco:
Need "chore:" or "feat:" in front of those commit messages. Word them in the present tense, e.g.
```
feat: plugins
```

include/mrdox/Support/ThreadPool.hpp
vinniefalco: how do we fix it

PR #287, opened 2023-06-12
eliminate Scope and Reference, store namespace member in single vector

sdkrystian:


source/CorpusImpl.cpp
```diff
@@ -291,12 +283,11 @@ class CorpusImpl::
...
-        canonicalize(I.Children.Namespaces);
-        canonicalize(I.Children.Records);
-        canonicalize(I.Children.Functions);
-        canonicalize(I.Children.Typedefs);
-        canonicalize(I.Children.Enums);
-        canonicalize(I.Children.Vars);
+        canonicalize(I.Members);
+        // KRYSTIAN FIXME: should we canonicalize specializations?
+        // we shouldn't canonicalize anything if we intend to
+        // preserve declaration order.
```
vinniefalco, line 289: I would say we canonicalize in the Generator not the Corpus but that's a bigger change

include/mrdox/Metadata/Type.hpp
```diff
@@ -34,24 +36,16 @@ namespace mrdox {
...
-
-    // Convenience constructor for when there is no symbol ID or info type
-    // (normally used for built-in types in tests).
-    explicit
-    TypeInfo(
-        std::string_view Name)
-        : Reference(
-            SymbolID::zero, Name, InfoKind::Default)
+        SymbolID ID = SymbolID::zero,
+        std::string_view name = "") noexcept
```
vinniefalco, line 46: the smallest nitpick imaginable: consider this instead:
```
std::string_view name = {}
```
What do you think?

sdkrystian: Yeah you're right that's better

source/-adoc/AdocSinglePageWriter.cpp
```diff
@@ -44,16 +44,24 @@ build()
 
 //------------------------------------------------
 
+// FIXME: we really should sort the members into
+// vectors of their respective types in one pass
```
vinniefalco, line 48: I have deleted all this code in my adoc branch so... this comment is doomed

include/mrdox/Metadata/Symbols.hpp
```diff
@@ -109,7 +109,6 @@ using OptionalSymbolID = Optional<SymbolID>;
 */
 enum class InfoKind
 {
-    Default,
     Namespace,
```
vinniefalco, line 112: This needs to explicitly name the 0 value for bitstream. And you should let `Function` have the value 0, because it is the most common.

include/mrdox/Metadata/Info.hpp
```diff
@@ -36,7 +35,19 @@ struct MRDOX_VISIBLE
...
+
+    /** Declaration access.
+
+        Class members use:
+        @li `AccessKind::Public`,
+        @li `AccessKind::Protected`, and
+        @li `AccessKind::Private`.
+
+        Namespace members use `AccessKind::None`.
+        Friends use `AccessKind::Friend`.
```
vinniefalco, line 48: umm... I'm not a fan of this. Why not just None

sdkrystian: I have to update this

include/mrdox/Platform.hpp
```diff
@@ -63,6 +64,16 @@ namespace mrdox {
 # error unknown platform for dynamic linking
 #endif
 
+//------------------------------------------------
+
+#define MRDOX_ASSERT(x) assert(x)
```
vinniefalco, line 69: `#ifndef MRDOX_ASSERT` first?

sdkrystian: quoth The Falco "this is not a boost library". it's already inside the header guard anyways

vinniefalco: but someone might want to change it

sdkrystian: very well

PR #274, opened 2023-06-09
Plugin

klemens-morgenstern:

PR #289, opened 2023-06-12
Plugin2

klemens-morgenstern:
Same as #274 but rebased onto #285.

### DIFF
--- a/include/mrdox/Config.hpp
+++ b/include/mrdox/Config.hpp
@@ -111,13 +111,11 @@ public:
 
     /** Destructor.
     */
-    MRDOX_DECL
     virtual
     ~Config() noexcept = 0;
 
     /** Return a pool of threads for executing work.
     */
-    MRDOX_DECL
     virtual
     ThreadPool&
     threadPool() const noexcept = 0;
--- a/include/mrdox/Metadata/DomMetadata.hpp
+++ b/include/mrdox/Metadata/DomMetadata.hpp
@@ -43,6 +43,12 @@ public:
     */
     virtual ~DomCorpus();
 
+    DomCorpus(DomCorpus const& corpus) = delete;
+    DomCorpus(DomCorpus&& corpus) noexcept;
+
+    DomCorpus& operator=(DomCorpus const& corpus) = delete;
+    DomCorpus& operator=(DomCorpus&& corpus) noexcept = delete;
+
     /** Constructor.
 
         Ownership of the specified Corpus is not
--- a/include/mrdox/Metadata/Javadoc.hpp
+++ b/include/mrdox/Metadata/Javadoc.hpp
@@ -605,6 +605,12 @@ public:
     */
     Javadoc() noexcept;
 
+    Javadoc(const Javadoc & ) = delete;
+    Javadoc& operator=(const Javadoc & ) = delete;
+
+    Javadoc(Javadoc && ) noexcept;
+    Javadoc& operator=(Javadoc && ) noexcept;
+
     /** Constructor
     */
     explicit
--- a/include/mrdox/Platform.hpp
+++ b/include/mrdox/Platform.hpp
@@ -38,6 +38,7 @@ namespace mrdox {
 
 // static linking
 #if defined(_MSC_VER)
+#pragma warning(disable:4251)
 # define MRDOX_SYMBOL_EXPORT __declspec(dllexport)
 # define MRDOX_SYMBOL_IMPORT __declspec(dllimport)
 # if defined(MRDOX_TOOL) // building tool
--- a/include/mrdox/Support/Handlebars.hpp
+++ b/include/mrdox/Support/Handlebars.hpp
@@ -635,7 +635,6 @@ public:
         @return The value of the property, or nullptr if the property does not exist
         @return `true` if the property was defined, `false` otherwise
      */
-    MRDOX_DECL
     std::pair<dom::Value, bool>
     lookupProperty(
         dom::Value const& context,
--- a/src/lib/Metadata/DomMetadata.cpp
+++ b/src/lib/Metadata/DomMetadata.cpp
@@ -861,6 +861,9 @@ public:
 DomCorpus::
 ~DomCorpus() = default;
 
+
+DomCorpus::DomCorpus(DomCorpus&& corpus) noexcept = default;
+
 DomCorpus::
 DomCorpus(
     Corpus const& corpus_)
--- a/src/lib/Metadata/Javadoc.cpp
+++ b/src/lib/Metadata/Javadoc.cpp
@@ -195,6 +195,9 @@ Javadoc(
 {
 }
 
+Javadoc::Javadoc(Javadoc && ) noexcept = default;
+Javadoc& Javadoc::operator=(Javadoc && ) noexcept = default;
+
 doc::Paragraph const*
 Javadoc::
 brief() const noexcept


CMakeLists.txt
```diff
@@ -212,6 +204,19 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 source_group(TREE ${PROJECT_SOURCE_DIR}/include/mrdox PREFIX "include" FILES ${INCLUDES})
 source_group(TREE ${PROJECT_SOURCE_DIR}/source PREFIX "source" FILES ${SOURCES})
 
+#-------------------------------------------------
+#
+# Plugins
+#
+#-------------------------------------------------
+
+file(GLOB plugins LIST_DIRECTORIES 1 CONFIGURE_DEPENDS plugins/*)
```
vinniefalco, line 214: but it is called addons/ ?

mrdox.cmake
```diff
@@ -57,6 +61,7 @@ function(mrdox)
                         ${MRDOX_COMPILE_COMMANDS}
                         --format=${MRDOX_TARGET_FORMAT}
                         --output=${MRDOX_TARGET_OUTPUT}
+                        --plugins-path=${MRDOX_TARGET_PLUGINS_PATH}
```
vinniefalco: but this is not right, the command line argument is "--addons"

source/Support/GeneratorsImpl.cpp
```diff
@@ -39,13 +45,42 @@ refresh_plist()
         plist_.push_back(g.get());
 }
 
+extern llvm::cl::opt<std::string>   PluginsPath;
```
vinniefalco, line 48: No we don't do options this way... see source/Tool/ToolArgs.hpp

source/Support/GeneratorsImpl.cpp
```diff
@@ -39,13 +45,42 @@ refresh_plist()
...
+
+      auto lib = llvm::sys::DynamicLibrary::getPermanentLibrary(itr->path().c_str(), &err_);
+      auto mkgen = lib.getAddressOfSymbol("makeMrDoxGenerator");
+
+      if (mkgen)
+        err = insert(reinterpret_cast<decltype(&makeAdocGenerator)>(mkgen)());
+    }
+
+    if (ec)
+      throw std::system_error(ec);
```
vinniefalco: prefer throwing `Error(ec)`

CMakeLists.txt
```diff
@@ -247,10 +247,10 @@ if (MRDOX_BUILD_TESTS)
         # test run
         include(mrdox.cmake)
 
-        mrdox(FORMAT adoc CONFIG docs/mrdox.yml SOURCE ${TEST_SOURCE} OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/test-docs)
-        mrdox(FORMAT xml  CONFIG docs/mrdox.yml SOURCE ${TEST_SOURCE} OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/test-docs)
-        mrdox(FORMAT test CONFIG docs/mrdox.yml SOURCE ${TEST_SOURCE} test_generator OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/test-docs
-              PLUGINS_PATH ${CMAKE_CURRENT_BINARY_DIR})
+        mrdox(FORMAT adoc CONFIG docs/mrdox.yml SOURCES ${TEST_SOURCE} OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/test-docs)
+        mrdox(FORMAT xml  CONFIG docs/mrdox.yml SOURCES ${TEST_SOURCE} OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/test-docs)
+        mrdox(FORMAT test CONFIG docs/mrdox.yml SOURCES ${TEST_SOURCE} test_generator
```
vinniefalco, line 252: I don't understand, what does this do?

include/mrdox/Plugin.hpp
```diff
@@ -12,16 +12,36 @@
...
+
+#include <string_view>
 
 namespace clang {
 namespace mrdox {
 
+class Generators;
+class Config;
+
+struct Plugin
```
vinniefalco, line 25: But what is this type? it should be `class` according to clang style. But why is it here? Is this really just a helper for a plugin? If so, then it should be in the sources of the example plugin and not part of mrdox public API

include/mrdox/Plugin.hpp
```diff
@@ -12,16 +12,36 @@
...
+    }
+};
+
 extern "C"
 {
 
-// implement this function if you want to provide your own generator
-std::unique_ptr<Generator> makeMrDoxGenerator();
+// implement this function if you want to provide your own plugin
+Plugin & mrDoxMakePlugin(const Config & cfg);
```
vinniefalco, line 44: This function should not be necessary and it sure wouldn't be public?

plugins/test/generator.cpp
```diff
@@ -9,38 +9,50 @@
...
+    fileExtension() const noexcept override {
+        return "test";
+    }
+
+    clang::mrdox::Error
+    buildOne(
+            std::ostream &os,
+            clang::mrdox::Corpus const &corpus) const override
+            {
+        os << "It works!";
```
vinniefalco, line 37: cool heh

klemens-morgenstern: That'll cause an error of undefined symbols on startup.

include/mrdox/Plugin.hpp
klemens-morgenstern: That's the plugin entry point, you described as `MrDoxPluginMain`.

include/mrdox/Plugin.hpp
klemens-morgenstern: Sorry, old habit.

CMakeLists.txt
klemens-morgenstern: Run mrdox with the test plugin, so we check the plugin system when we build.

source/Support/GeneratorsImpl.cpp
```diff
@@ -39,13 +45,42 @@ refresh_plist()
...
+    std::error_code ec;
+    for (auto itr = llvm::sys::fs::directory_iterator(plugin_path, ec);
+         itr != llvm::sys::fs::directory_iterator() && !ec;
+         itr.increment(ec))
+    {
+      if (!itr->path().ends_with(".dll")
+       && !itr->path().ends_with(".so"))
+        continue;
+
+      auto lib = llvm::sys::DynamicLibrary::getPermanentLibrary(itr->path().c_str(), &err_);
```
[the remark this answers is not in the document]
klemens-morgenstern: you mean `&err_` ?

mrdox.cmake
```diff
@@ -30,7 +30,7 @@ function(mrdox)
     cmake_parse_arguments(
         MRDOX_TARGET
         "" # options
-        "OUTPUT;FORMAT;CONFIG" # one_value_keywords
+        "OUTPUT;FORMAT;CONFIG;PLUGINS" # one_value_keywords
```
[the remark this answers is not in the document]
klemens-morgenstern: Leftover from draft 1, obsolete.

CMakeLists.txt
klemens-morgenstern: Wait no, that's correct - the source for the plugin (as of now) is in `/plugins/...` while it's build artifact (i.e. the dll) gets put into `addons/plugins`.

klemens-morgenstern:
I can't figure out what msvc is doing.

source/Tool/GenerateAction.cpp
```diff
@@ -13,6 +13,8 @@
 #include "ConfigImpl.hpp"
 #include "CorpusImpl.hpp"
 #include "AST/AbsoluteCompilationDatabase.hpp"
+#include "Plugins.hpp"
```
vinniefalco, line 16: this is in the wrong order

source/Tool/GenerateAction.cpp
```diff
@@ -62,6 +64,11 @@ DoGenerateAction()
         return absPath.getError();
     auto workingDir = files::getParentDir(*absPath);
 
+    // load in plugins
```
vinniefalco, line 67: this comment adds no information

source/Tool/GenerateAction.cpp
```diff
@@ -62,6 +64,11 @@ DoGenerateAction()
...
+    if (auto err = loadPlugins(toolArgs.addonsDir); err.failed())
```
vinniefalco, line 68: you don't need the "; err.failed()" part

source/Tool/Plugins.cpp
```diff
@@ -0,0 +1,63 @@
+//
+// Copyright (c) 2023 Klemens Morgenstern (klemens.morgenstern@gmx.net)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
```
vinniefalco, line 4: wrong license

source/Tool/Plugins.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include <mrdox/Config.hpp>
+#include <mrdox/Generators.hpp>
+#include <mrdox/Plugin.hpp>
+#include <mrdox/Support/Path.hpp>
+#include <mrdox/Support/Report.hpp>
+#include <mrdox/Version.hpp>
+
+#include "Support/Path.hpp"
+#include "Support/Debug.hpp"
```
vinniefalco, line 18: These includes should come before the "<mrdox/..." ones

source/Tool/Plugins.cpp
```diff
@@ -0,0 +1,63 @@
...
+//
+
+#include "Plugins.hpp"
+
+#include <mrdox/Config.hpp>
+#include <mrdox/Generators.hpp>
+#include <mrdox/Plugin.hpp>
+#include <mrdox/Support/Path.hpp>
+#include <mrdox/Support/Report.hpp>
+#include <mrdox/Version.hpp>
```
vinniefalco, line 15: This should come before the "<mrdox/Support/..." includes

source/Tool/Plugins.cpp
```diff
@@ -0,0 +1,63 @@
...
+Error
+loadPlugins(
+        std::string_view addonsDir
+        )
+{
+
+    std::string err_;
+
+    SmallPathString pluginPath{addonsDir};
+    pluginPath.append("/plugins");
```
vinniefalco, line 35: use the functions in <mrdox/Support/Path.hpp> please

source/Tool/Plugins.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+    std::error_code ec;
+    for (auto itr = llvm::sys::fs::directory_iterator(pluginPath, ec);
+         itr != llvm::sys::fs::directory_iterator() && !ec;
```
vinniefalco, line 39: Add this to the top of the function
```
namespace fs = llvm::sys::fs;
```

source/Tool/Plugins.cpp
```diff
@@ -0,0 +1,63 @@
...
+         itr.increment(ec))
+    {
+        if (!itr->path().ends_with(".dll")
+         && !itr->path().ends_with(".so"))
+            continue;
+
+        auto lib = llvm::sys::DynamicLibrary::getPermanentLibrary(itr->path().c_str(), &err_);
+        if (!lib.isValid())
+            return Error("couldn't load {} because {}", itr->path(), err_);
```
vinniefalco, line 48: Well no this message is not correct because the caller will end up printing this:
```
Could not initialize plugins because couldn't load {} because {}"
```

The text in the `Error` is the "because" so you might write this instead:

```
return Error( "getPermanentLibrary(\"{}\") returned \"{}\"" );
```

source/Tool/Plugins.cpp
```diff
@@ -0,0 +1,63 @@
...
+        auto mkgen = lib.getAddressOfSymbol("MrDoxPluginMain");
+
+        if (!mkgen)
+            return Error("plugin library {} does not provide the plugin entry point MrDoxPluginMain",
+                         itr->path());
+
+        (*reinterpret_cast<decltype(&MrDoxPluginMain)>(mkgen))(projectVersionMajor, projectVersionMinor);
+    }
+    if (ec)
+        return Error("Iterating through the plugin director failed ", ec);
```
vinniefalco, line 58: needs work

source/Tool/Plugins.cpp
```diff
@@ -0,0 +1,63 @@
...
+    return {};
```
vinniefalco, line 59: Since when do fmt-style substitutions work in regular C++?

source/Tool/Plugins.hpp
```diff
@@ -0,0 +1,27 @@
+//
+// Copyright (c) 2023 Klemens Morgenstern (klemens.morgenstern@gmx.net)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
```
vinniefalco, line 4: license is wrong

source/Tool/Plugins.hpp
```diff
@@ -0,0 +1,27 @@
...
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+
+#ifndef MRDOX_PLUGINS_HPP
+#define MRDOX_PLUGINS_HPP
```
vinniefalco, line 9: Well this is not the right include guard name

source/Tool/Plugins.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include "mrdox/Support/Error.hpp"
```
vinniefalco, line 11: needs angle brackets

source/Tool/GenerateAction.cpp
klemens-morgenstern: What's the right order?

vinniefalco: "this folder" comes before "other folder:"
``
#include "ConfigImpl.hpp"
#include "CorpusImpl.hpp"
#include "Plugins.hpp"
#include "AST/AbsoluteCompilationDatabase.hpp"
```

File: source/Tool/GeneratorsImpl.cpp [hunk 25]
```diff
@@ -73,17 +74,13 @@ insert(
...
-}
-
-Generators const&
+Generators &
 getGenerators() noexcept
 {
-    return getGeneratorsImpl();
+    auto & gens = toolWorld().generators;
+    if (gens == nullptr)
+        gens = std::make_unique<GeneratorsImpl>();
```
vinniefalco, line 82: why not do this in `ToolWorld::ToolWorld`?

source/Tool/ToolMain.cpp
```diff
@@ -55,14 +56,16 @@ extern Error DoGenerateAction();
...
-       << "\n    version: " << project_version
+    os << projectName
+       << "\n    " << projectDescription
+       << "\n    version: " << projectVersion
        << "\n    built with LLVM " << LLVM_VERSION_STRING;
 }
 
 int mrdox_main(int argc, char const** argv)
 {
+    clang::mrdox::ToolWorld globalState;
```
vinniefalco, line 67: this got renamed it seems

source/Tool/GeneratorsImpl.cpp
vinniefalco: I don't like this. Checking every time doesn't really make sense once it is already created. And I don't like that the lifetime is somewhat indeterminate - it depends on when `getGeneratorsImpl` is called. This is not ideal. Instead, we should do it in the constructor as I wrote it, because then the beginning of the lifetime is crystal clear to the reader.

source/Tool/GeneratorsImpl.hpp
```diff
@@ -8,12 +8,12 @@
 // Official repository: https://github.com/cppalliance/mrdox
 //
 
-#ifndef MRDOX_TOOL_SUPPORT_GENERATORSIMPL_HPP
-#define MRDOX_TOOL_SUPPORT_GENERATORSIMPL_HPP
+#ifndef MRDOX_TOOL_GENERATORSIMPL_HPP
+#define MRDOX_TOOL_GENERATORSIMPL_HPP
```
vinniefalco, line 12: actually the "correct" name is `MRDOX_TOOL_TOOL_GENERATORSIMPL_HPP` :)

include/mrdox/Support/Error.hpp
```diff
@@ -213,4 +212,4 @@ struct fmt::formatter<std::error_code>
     }
 };
 
-#endif
+#endif
```
vinniefalco, line 215: missing newline

klemens-morgenstern:
@vinniefalco any further change required?

include/mrdox/Plugin.hpp
```diff
@@ -12,16 +12,36 @@
...
+
+struct Plugin
+{
+    // MUST always come first
+    const std::string_view version;
+    Plugin(const std::string_view version) : version(version) {}
+
+    virtual
+    Error
+    registerGenerators(Generators & gen)
```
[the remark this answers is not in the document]
vinniefalco: The public API is designed so that there can be multiple invocations of each generator, each with a different Config and Corpus. For example to generate both single-page and multipage output in one run of the tool. The `Config` is tied to the `Corpus` and provided in the `Generator::build` member function call. At the time the plugin is loaded and initialized, there is no config yet.

include/mrdox/Plugin.hpp
vinniefalco: It is not a public API is what I am saying

vinniefalco:
We just need to make sure that the CMakeLists.txt is going to be right

klemens-morgenstern:
I don't know what "correct" means here.

source/Tool/GenerateAction.cpp
klemens-morgenstern: I don't? Isn't how do I negate it otherwise?

vinniefalco: `err` evaluates to `true` if there's an error (please double check this lol I get it confused sometimes)

alandefreitas:
@klemens-morgenstern Could you rebase that on top of `develop` so we can review "the real thing"?

klemens-morgenstern:
@vinniefalco redo.

src/tool/Plugins.cpp
```diff
@@ -0,0 +1,87 @@
...
+loadPlugins(
+    const std::string & addonsDir)
+{
+    auto pluginDir = files::appendPath(addonsDir, "plugins");
+    PluginEnvironmentImpl env;
+    auto err = forEachFile(
+        pluginDir,
+        [&](std::string_view path) -> Error
+        {
+            return loadOnePlugin(std::string(path), env);
```
vinniefalco, line 74: the error should be reported immediately, not added to a vector

include/mrdox/Plugin.hpp
```diff
@@ -0,0 +1,45 @@
...
+namespace clang {
+namespace mrdox {
+
+class Generator;
+class PluginEnvironment
+{
+public:
+    virtual
+    void
+    reportError(Error error) = 0;
```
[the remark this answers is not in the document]
klemens-morgenstern: Because the tool puts them into a warning. Because a plugin failure isn't critical. You reckon I should just remove this and let the plugin use `report::error` as it sees fit?

src/tool/Plugins.cpp
klemens-morgenstern: This and above removed.

include/mrdox/Config.hpp
```diff
@@ -35,7 +35,7 @@ class ThreadPool;
     particular directory from which absolute paths
     are calculated from relative paths.
 */
-class MRDOX_DECL
+class MRDOX_VISIBLE
```
alandefreitas, line 38: What is the rationale for changing these?

src/tool/Plugins.cpp
```diff
@@ -0,0 +1,72 @@
...
+    {
+        getGeneratorsImpl().insert(std::move(generator));
+    }
+};
+
+Error loadOnePlugin(
+    std::string pth,
+    PluginEnvironment & env)
+{
+    if (pth.ends_with(".so") || pth.ends_with(".dll"))
```
vinniefalco, line 38: is this case sensitive?

src/tool/Plugins.cpp
```diff
@@ -0,0 +1,72 @@
...
+    void
+    addGenerator(
+        std::unique_ptr<Generator> generator) override
+    {
+        getGeneratorsImpl().insert(std::move(generator));
+    }
+};
+
+Error loadOnePlugin(
+    std::string pth,
```
vinniefalco, line 35: either use one letter or spell the word, but "pth" is not a great abbreviation, no vowel, unpronounceable. `filePath` is what I use to indicate a path that must identify an ordinary file.

src/tool/Plugins.cpp
vinniefalco: But I'm curious what was the original motivation?

include/mrdox/Plugin.hpp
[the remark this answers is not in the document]
vinniefalco: Yes that makes more sense

include/mrdox/Plugin.hpp
```diff
@@ -0,0 +1,45 @@
...
+};
+
+}
+}
+
+// the version gets delivers as an argument, because env might incompatible.
+// this function must return false on version conflicts!
+extern "C"
+MRDOX_SYMBOL_EXPORT
+bool
```
vinniefalco, line 36: Why does it have to return false?

include/mrdox/Config.hpp
klemens-morgenstern: Oh. I thought they need to be exported to the dll in general.

src/tool/Plugins.cpp
klemens-morgenstern: probably not, good catch.

include/mrdox/Plugin.hpp
klemens-morgenstern: To indicate a version mismatch. If the version doesn't match I can't use any of the C++ APIs, but need to rely on a super primitive C API.

vinniefalco: But what does the caller of `MrDoxMain` do with the information? The tool doesn't call into the plugin, the plugin calls into the tool. So it is the plugin's job to know if the version is safe to call into. If the version is not safe then the plugin simply avoids installing itself. I dont understand how the bool is used.

src/tool/Plugins.cpp
```diff
@@ -0,0 +1,72 @@
...
+        auto res = llvm::sys::DynamicLibrary::getPermanentLibrary(pth.c_str(), &errMsg);
+        if (!res.isValid())
+            return formatError("Couldn't load {}, because '{}'", pth, errMsg);
+
+        auto f = res.getAddressOfSymbol("MrDoxMain");
+        if (f == nullptr)
+            return formatError("{}, doesn't export MrDoxMain symbol", pth);
+
+        auto func = reinterpret_cast<decltype(&MrDoxMain)>(f);
+        if (!func(projectVersionMajor, projectVersionMinor, env))
```
vinniefalco, line 50: This is not right. What if the plugin failed to load because a resource file it needs is missing? Should it return true or false? Having the plugin return the bool is too clever and needlessly couples the host and the plugin. Just return void from MrDoxMain and let the plugin do its own version checking and reporting.

include/mrdox/Plugin.hpp
klemens-morgenstern: So we can report that a plugin couldn't load for this reason, because it can't use `report:error`.

src/tool/Plugins.cpp
klemens-morgenstern: It uses `report::error`.

src/tool/Plugins.cpp
klemens-morgenstern: To report plugin load errors as warnings instead. Because you might still be able to run the tool if a plugin fails.

include/mrdox/Config.hpp
alandefreitas: Yes. MRDOX_DECL does that already. MRDOX_VISIBLE. In doubt, MRDOX_DECL is correct 99% of the time.

src/tool/Plugins.cpp
vinniefalco: This is your code:
```
        if (!func(projectVersionMajor, projectVersionMinor, env))
            return formatError("Couldn't load {} - version mismatch.", pth);
```

Now lets say MrDoxMain tries to load a resource file and it fails. But the version is ok. Should MrDoxMain return "true", or "false?" If it returns false then the error message will be incorrect. It will say "version mismatch" but it is really a "missing file" error.

If it returns true then the reader will infer that the plugin loaded successfully even though it could not carry out its installation.

Do you understand my point? Giving MrDoxMain a return value was not a good design choice. Do you remember the signature I gave you?

image

It was like that for a reason.

klemens-morgenstern:
Adopted to review comments. I will need to review the DECL -> VISIBLE changes.

CMakeLists.txt
```diff
@@ -34,7 +34,6 @@ include(GNUInstallDirs)
 
 option(MRDOX_INSTALL "Configure install target" ON)
 option(MRDOX_PACKAGE "Build install package" ON)
-option(MRDOX_BUILD_SHARED "Link shared" ${BUILD_SHARED_LIBS})
```
alandefreitas, line 37: If we're really messing seriously with shared libraries, should we be testing this in CI?

alandefreitas: For instance,

```
          factors: |
            gcc Shared
            msvc Shared
# ...
    uses: ... cmake-workflow
        # ...
        with:
            # ... 
           shared: ${{ matrix.shared }}
```

or a union of whatever factors are already there...

Here's an example: https://github.com/alandefreitas/cpp-actions/blob/develop/.github/workflows/ci.yml

CMakeLists.txt
```diff
@@ -77,23 +70,30 @@ string(REGEX REPLACE " /W[0-4]" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 
 # Duktape
 set(DUKTAPE_SOURCE_ROOT "" CACHE STRING "Root of the Duktape source tree")
-if (NOT DUKTAPE_SOURCE_ROOT)
+
+if (DUKTAPE_SOURCE_ROOT)
```
alandefreitas, line 74: What there are a reason to invert the logic here? I'm not sure I'm following...

CMakeLists.txt
```diff
@@ -77,23 +70,30 @@ string(REGEX REPLACE " /W[0-4]" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
...
+            "$<BUILD_INTERFACE:${DUKTAPE_SOURCE_ROOT}/src>"
+            "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>")
+
+    if (CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
+        target_compile_options(duktape PRIVATE -w)
+    elseif (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+        target_compile_options(duktape PRIVATE /w)
+    endif()
+    set(DUKTAPE_INSTALL duktape)
+elseif(DUKTAPE_PREBUILT)
```
alandefreitas, line 87: What does DUKTAPE_PREBUILT mean?

include/mrdox/Plugin.hpp
```diff
@@ -0,0 +1,45 @@
+//
+// This is a derivative work. originally part of the LLVM Project.
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (c) 2023 Vinnie Falco (vinnie.falco@gmail.com)
```
alandefreitas, line 7: Klemens

CMakeLists.txt
klemens-morgenstern: It is tested through the test-plugin.

CMakeLists.txt
klemens-morgenstern: it means I have it installed on my machine, in my case with `dnf install duktape-devel`.

include/mrdox/Plugin.hpp
klemens-morgenstern: hey it's my copyright, I am free to give it away.

CMakeLists.txt
alandefreitas: Oh... Sorry. The MRDOX_BUILD_SHARED support is actually being removed. OK.

CMakeLists.txt
alandefreitas: OK. I would say we have to document this but https://github.com/cppalliance/mrdox/pull/434, which includes the basic docs, is now hanging.

What is it you get with `dnf install duktape-devel`? Is it equivalent with what you would get from other package manager?

I'm still not sure I understand the logic. Maybe it's some pattern I don't know about. I understand

```
set_target_properties(duktape PROPERTIES IMPORTED_LOCATION ${DUKTAPE_PREBUILT})
```

but then

```
add_library(duktape SHARED IMPORTED src/test-plugin/test-plugin.cpp)
```

because it doesn't have anything to do with duktape and doesn't explain where the headers would come from. How does this become something equivalent to the path where we use DUKTAPE_SOURCE_DIR?

klemens-morgenstern: Well it's importing the lib. So I have an imported lib instead of a built one, so I need to set the import_location.

I am not saying that this is a great solution, nor that it needs to be in this PR. I can remove it and make a separate one if you like me to

alandefreitas: > Well it's importing the lib. So I have an imported lib instead of a built one, so I need to set the import_location.

Yes. This is precisely the part I mentioned I understand perfectly:

```
set_target_properties(duktape PROPERTIES IMPORTED_LOCATION ${DUKTAPE_PREBUILT})
```

It's the rest I don't understand. I don't understand is: 

1) how `src/test-plugin/test-plugin.cpp` is being imported since it's not compiled yet and I thought CMake didn't accept in source files in imported targets, 
2) what `src/test-plugin/test-plugin.cpp` semantically has to do with duktape as a target, 
3) how this imported library compares to the libraries we can get from other sources, 
4) how this can be equivalent with the alternative path based on DUKTAPE_SOURCE_DIR that doesn't include any of that and sets other properties 
5) where the headers come from, 
6) how can we achieve the same with the DUKTAPE_SOURCE_DIR pattern everyone else has been using since it doesn't include this file. 

Basically, all the semantics of what you're trying to achieve are unclear to me. I'm not saying it's wrong. I'm literally saying I don't understand it. But leaving modesty aside for a minute, if I don't understand it most people won't understand it either.

klemens-morgenstern: 1. That's a mistake.
2. Mistake.
3. No idea.
4. Because the includes are /usr/include
5. `/usr/include` in my case
6. Can't, because DUKTAPE_SOURCE_DIR is building.

klemens-morgenstern: I think I should remvoe this - can't we move duktape to FetchContent though?

alandefreitas: > That's a mistake. Mistake. No idea. Because the includes are /usr/include /usr/include in my case Can't, because DUKTAPE_SOURCE_DIR is building.

Oh... that's pretty bad then. Good thing I asked.

> I think I should remvoe this - can't we move duktape to FetchContent though?

I wrote FetchContent for fmt and duktape but Vinnie hated it. I think his internet was down and he deleted the build dir for some reason then didn't have duktape available. I'm always OK with fetchcontent as a second option, since the DUKTAPE_SOURCE_DIR option was still there, nothing changed in that sense, and I guess I would just tether some internet from my phone for a few seconds, download duktape and get back to it. 

But anyway... I don't think FetchContent helps much in this case. I'll just get the same content you already get from DUKTAPE_SOURCE_DIR. The instructions are in that other PR https://github.com/cppalliance/mrdox/pull/434 I mentioned before.

klemens-morgenstern: So should I remove this or do you think it's an ok option for linux builds?

alandefreitas: I'm using Linux most of the time and I don't ever see myself using that. I don't think the problem here is related to the OS at all. The problem is the pattern is not generalizable (although I understand the default hint paths cmake uses on linux do make a difference). `DUKTAPE_PREBUILT` could just be the usual `DUKTAPE_ROOT` cmake uses for all packages, if we had reliable packages. Then we'd have to agree on what these packages are, which is hard (because there are no reliable packages even from the original repo let alone CMake integration for them) but doable and fine. 

The problem at this point is not even about the packages anymore, which is a problem (`src/test-plugin/test-plugin.cpp` is not part of duktape, the duktape target should have the headers, if the library is prebuilt then cmake should be the one finding it, etc...) but that could easily be in another PR. The problem is the thing doesn't work the same when it takes the usual `DUKTAPE_SOURCE_ROOT` path everyone else is using.

I'd leave only `DUKTAPE_SOURCE_ROOT` here for now and if we want to have an alternative path that uses an installed version of duktape, then we'd choose which package that would be, document what it is, make the binary version the default in CML, fall back to `DUKTAPE_SOURCE_ROOT` ensuring both behave the same, and make the script look good by using the usual CMake conventions (the hint path is `DUKTAPE_ROOT` and then we use `find_package` or at least `find_library`/`find_header`). That could be on the same PR or a different one.

klemens-morgenstern: What if I can make it work with a `DUKTAPE_ROOT` where it checks for source and for a lib, and just picks the right thing?

alandefreitas: I guess that works. Just notice that `DUKTAPE_ROOT` is already the default cmake uses for the hint if I'm not wrong. If not, it's just a called to `find_*` using it as the hint. So you just really need to `find_library` and `find_path` (is it `find_path` the best command for finding headers(?)). We can leave only `DUKTAPE_SOURCE_ROOT` in the docs for now because `DUKTAPE_ROOT` is kind of unstable in terms of recommending to the user how to install it (also because the PR with these docs is not merged yet).

CMakeLists.txt
```diff
@@ -77,23 +70,37 @@ string(REGEX REPLACE " /W[0-4]" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
...
+else () # NOT DUKTAPE_SOURCE_ROOT
+    # try to find it on the machine
+    find_file(DUKTAPE_HEADER duktape.h)
+    find_library(DUKTAPE_LIBRARY duktape)
+    if (NOT DUKTAPE_HEADER STREQUAL DUKTAPE_HEADER-NOTFOUND
+        AND NOT DUKTAPE_LIBRARY STREQUAL DUKTAPE_LIBRARY-NOTFOUND)
+        add_library(duktape SHARED IMPORTED)
+        set_target_properties(duktape PROPERTIES IMPORTED_LOCATION ${DUKTAPE_LIBRARY})
+        find_library(DUKTAPE_LIBRARY duktape)
+    else()
```
klemens-morgenstern, line 96: @alandefreitas how's that? Just import if it can be found, without any argument.s

alandefreitas: Yes. This looks OK. More idiomatic.

```
find_library(DUKTAPE_LIBRARY duktape)
```

is being called twice though. 

And the header path/directory is being ignored rather than set in the imported target.

klemens-morgenstern: Ok, how about now?

klemens-morgenstern: Good idea, done. You think it's ready to merge now?

alandefreitas: To be honest, I just reviewed the CMake scripts and the MRDOX_DECL/VISIBLE macros... I had a look at the rest of the PR and it looks OK to me but I don't have any serious informed opinion about it because I don't know what requirements Vinnie gave you.

It appears that he's busy now and I'm quite unsure about the state of authority over these decisions in the meanwhile. I've only been merging handlebars stuff in the meanwhile because it doesn't touch MrDox at all. We could wait for him or just ask him.

vinniefalco:
We need at least some of the changes such as the visibility modifications and exports

alandefreitas:
Obsoleted

PR #293, opened 2023-06-12
CI runs mrdox in Debug configuration

sdkrystian:


alandefreitas:
The build type should usually go in

```yml
          build-type: ${{ matrix.build-type }}
```

By putting it in extra-args, the action might try to put it there twice in the configure step.

PR #295, opened 2023-06-13
replace `AnyList` with `std::vector<std::unique_ptr<Node>>`

sdkrystian:


sdkrystian:
@vinniefalco i'm not thrilled with the `equals` function i've implemented here... do you have any better ideas?

include/mrdox/Metadata/Javadoc.hpp
```diff
@@ -82,13 +91,31 @@ struct Node
...
+    }
+
+
+    bool isTextDerived() const noexcept
+    {
+        return kind == Kind::text ||
+            kind == Kind::styled;
+    }
+
+    bool isBlockDerived() const noexcept
```
vinniefalco, line 115: wut is this

include/mrdox/Metadata/Javadoc.hpp
```diff
@@ -148,21 +188,35 @@ struct Block : Node
...
+                return a->equals(*b);
+            });
+    }
 
 protected:
     explicit
     Block(
         Kind kind_,
-        AnyList<Text> children_ = {}) noexcept
+        doc::List<Text> children_ = {}) noexcept
```
vinniefalco, line 219: do you need to say `doc::` here?

include/mrdox/Metadata/Javadoc.hpp
```diff
@@ -175,19 +229,25 @@ struct Paragraph : Block
 {
```
vinniefalco, line 229: please fix these warnings

include/mrdox/Metadata/Javadoc.hpp
```diff
@@ -423,15 +519,18 @@ struct MRDOX_VISIBLE
     template<class T, class U>
     static
     void
-    append(AnyList<T>& list, AnyList<U>&& other) noexcept
+    append(doc::List<T>& list, doc::List<U>&& other) noexcept
     {
-        list.splice_back(std::move(other));
+        list.reserve(list.size() + other.size());
+        for(auto& p : other)
+            list.emplace_back(static_cast<T*>(p.release()));
```
vinniefalco, line 526: why not `std::move(p)`?

source/-XML/XMLWriter.cpp
```diff
@@ -586,12 +587,12 @@ template<class T>
...
 writeNodes(
-    AnyList<T> const& list)
+    doc::List<T> const& list)
 {
     if(list.empty())
         return;
-    for(auto const& node : list)
-        writeNode(node);
+    for(const auto& node : list)
+        writeNode(*node);
```
vinniefalco, line 595: you changed east const to west const.... I am going to mandate east const in this project

source/Metadata/Javadoc.cpp
```diff
@@ -17,19 +18,22 @@
...
+auto
+move_to(
+    doc::List<T>& dst,
+    doc::List<U>& src,
+    typename doc::List<U>::iterator it)
+{
+    auto elem = std::move(*it);
+    it = src.erase(it);
+    dst.emplace(dst.end(),
+        static_cast<T*>(elem.release()));
```
vinniefalco, line 34: why not `std::move(*it)`?

source/Metadata/Javadoc.cpp
```diff
@@ -60,9 +64,21 @@ Javadoc::
...
 {
-    return
-        std::tie(brief_, blocks_) ==
-        std::tie(other.brief_, other.blocks_);
+    if(! brief_ || ! other.brief_)
+    {
+        if(brief_ != other.brief_)
+            return false;
+    }
+    else if(! brief_->equals(*other.brief_))
```
vinniefalco, line 72: what about params and tparams? this op== needs to be on Javadoc?

source/Metadata/Javadoc.cpp
```diff
@@ -118,38 +135,39 @@ postProcess()
...
+        blocks_.erase(match);
     }
-    else
-    {
-        static std::shared_ptr<doc::Paragraph const> empty_para =
-            std::make_shared<doc::Paragraph>();
-        brief_ = empty_para;
-    }
+    // KRYSTIAN FIXME: should an empty paragraph be used
+    // as the brief when no written brief exists?
```
vinniefalco, line 170: My thinking is that the generator wants to know that the brief was left out, in order to have more information when rendering. But it is also a pain in the ass because the generator has to check "if( ! brief )` all the time, which is a hassle in the Handlebars template. We need more practice/experience writing generators to know the answer to this.

include/mrdox/Metadata/Javadoc.hpp
sdkrystian: because `T` might be a derived class of `U`

source/Metadata/Javadoc.cpp
sdkrystian: `dst` may be a `std::vector<std::unique_ptr<U>>` where `U` is derived from `T`

include/mrdox/Metadata/Javadoc.hpp
sdkrystian: nope

include/mrdox/Metadata/Javadoc.hpp
```diff
@@ -423,15 +519,18 @@ struct MRDOX_VISIBLE
...
     }
 
     template<class T, class Child>
     static
     void
-    append(AnyList<T>& list, Child&& child)
+    append(doc::List<T>& list, Child&& child)
     {
         list.emplace_back(
             std::forward<Child>(child));
```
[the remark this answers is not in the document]
sdkrystian: yeah, it should be. fixing

PR #301, opened 2023-06-14
Rename VarInfo to VariableInfo, rename SymbolInfo to SourceInfo

sdkrystian:


vinniefalco:
Please use the commit message format introduced by @alandefreitas

source/Support/Debug.cpp
```diff
@@ -14,84 +14,19 @@
...
 {
-#if 1
     return llvm::outs();
-#else
-    static debug_ostream stream(llvm::outs());
-    return stream;
-#endif
 }
 
 llvm::raw_ostream& debug_errs()
```
vinniefalco, line 27: we should never be using `debug_errs`

vinniefalco:
Please fix the warnings from CI

PR #302, opened 2023-06-14
ci: update cpp-actions to v1.5.0

alandefreitas:
Among [changes](https://github.com/alandefreitas/cpp-actions/releases/tag/v1.5.0) with many bug fixes and performance improvements in all actions, the new setup-cpp step now sets up official apt repositories whenever possible for the compiler version being set. 

In practice, this allows all the most recent compiler versions to be set up with APT rather than our custom binaries, which is not only faster but also fixes many of the unstable workarounds we had before, such as manually setting LD_LIBRARY_PATH and the standard library configuration flags.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,16 +79,15 @@ jobs:
           find "$llvm_dir" -type f
 
       - name: Setup C++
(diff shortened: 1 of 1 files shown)


alandefreitas:
Merging this since it's been here for a while and it's reasonably uncontroversial

PR #307, opened 2023-06-15
more doc work

vinniefalco:


source/AST/AnyBlock.hpp
```diff
@@ -98,6 +98,47 @@ class JavadocNodesBlock
...
+            if(node->kind != doc::Kind::param)
+                return Error("direction on wrong kind");
+            auto param = static_cast<doc::Param*>(node);
+            param->direction = direction;
+            return Error::success();
+        }
+
+        case JAVADOC_NODE_HREF:
+        {
+            switch(auto node = nodes.back().get();
```
sdkrystian, line 131: I'm not a fan of single case switch statements .... we should consider removing them

vinniefalco: Yeah ok, remove it after we ship mrdox 2.0, when we know that we will likely not add anything more to the switch

include/mrdox/Support/SharedPtr.hpp
```diff
@@ -0,0 +1,325 @@
...
+    template<class U>
+    friend class AtomicSharedPtr;
+
+    SharedPtr(Base*) noexcept;
+
+public:
+    ~SharedPtr();
+    SharedPtr() noexcept = default;
+    SharedPtr(SharedPtr const& other) noexcept;
+    SharedPtr& operator=(SharedPtr const& other) noexcept;
```
k3DW, line 93: Do you need this copy constructor and assignment operator? Could they removed in favour of the templated ones below?

include/mrdox/Support/SharedPtr.hpp
```diff
@@ -0,0 +1,325 @@
...
+class SharedPtrBase
+{
+protected:
+    struct Base
+    {
+        virtual ~Base() = default;
+        Base() noexcept = default;
+        Base* addref() noexcept;
+        void release() noexcept;
+        virtual void* get() noexcept = 0;
```
k3DW, line 41: I'm wondering if this needs to be a virtual function. Could you use a templated `Base` instead?

include/mrdox/Support/SharedPtr.hpp
```diff
@@ -0,0 +1,325 @@
...
+        InitFunction must have this equivalent signature:
+        @code
+        SharedPtr<T>(void)
+        @endcode
+    */
+    template<class InitFunction>
+    requires std::is_invocable_r_v<
+        SharedPtr<T>, InitFunction>
+    SharedPtr<T>
+    load(InitFunction const& init)
```
k3DW, line 303: Might be overkill, but I'd take this function as a `InitFunction&&`, then use it as `std::invoke(std::forward<InitFunction>(init))`

include/mrdox/Support/SharedPtr.hpp
vinniefalco: Its a good question. I am not sure. Peter might know :)

include/mrdox/Support/SharedPtr.hpp
vinniefalco: If `Base` was a class template then it would be undefined behavior to cast `Base<T>*` to `Base<U>*`, which would be needed for those functions.

include/mrdox/Support/SharedPtr.hpp
vinniefalco: Hmm interesting what is the difference?

include/mrdox/Support/SharedPtr.hpp
k3DW: Ah right, inheritance

include/mrdox/Support/SharedPtr.hpp
k3DW: The whole incantation with `invoke` probably doesn't matter, since it has to be invocable with no arguments anyway. `init()` should be fine. Personal preference, I just like the look of the uniform invocation with `invoke`.

I'd still use a forwarding reference for the function type, otherwise any rvalue would get stored in this lvalue-const-ref, and someone looking at this later might do something with it they shouldn't

vinniefalco: "otherwise any rvalue would get stored in this lvalue-const-ref"

No idea what this means, I don't speak nerd

k3DW: You might get weird errors one day if someone makes an unexpected change. For now it's fine

vinniefalco: Can you give an example of how this code my be invoked which could cause a problem? That would help me visualize it.

PR #312, opened 2023-06-16
C++ Handlebars

alandefreitas:


source/Support/Handlebars.cpp
```diff
@@ -0,0 +1,820 @@
+//
+// This is a derivative work. originally part of the LLVM Project.
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (c) 2023 Vinnie Falco (vinnie.falco@gmail.com)
```
vinniefalco, line 7: umm..

alandefreitas:
Dom types integrated. 

Moving to private data and block parameter functionality now.

vinniefalco:
IMPRESSIVE

include/mrdox/Support/Dom.hpp
```diff
@@ -96,6 +96,19 @@ class MRDOX_DECL
     */
     Array(impl_type impl) noexcept;
 
+    /** Assignment.
+
+        The newly constructed array will contain
+        references to its structured data.
+    */
+    Array&
+    operator=(Array const& other)
```
vinniefalco, line 105: UGH... is there any way to avoid this? There are consequences for scripts...

include/mrdox/Support/Handlebars.hpp
```diff
@@ -0,0 +1,1037 @@
...
+
+    It allows many types to be used as output streams, including
+    std::string, std::ostream, llvm::raw_string_ostream, and others.
+
+ */
+class MRDOX_DECL OutputRef
+{
+    friend class Handlebars;
+
+    using fptr = void (*)(void * out, std::string_view sv);
```
vinniefalco, line 91: This is one of the reasons I rejected Peter's Mustache submission...because it is push based rather than pull based. It can't handle the simple use-case of filling a buffer to capacity and then resuming later...

include/mrdox/Support/Handlebars.hpp
```diff
@@ -0,0 +1,1037 @@
...
+    HandlebarsCallback();
+
+public:
+    /** Render the block content with the specified context
+
+        This function renders the block content with the specified context.
+
+        For example, the following template calls the helper "list" with a block section:
+
+        @code{.handlebars}
```
vinniefalco, line 240: Does this work? .handlebars ?

include/mrdox/Support/Handlebars.hpp
```diff
@@ -0,0 +1,1037 @@
...
+    zero, or an non-empty string, array or object.
+
+    @param arg The value to test
+
+    @return True if the value is truthy, false otherwise
+
+ */
+MRDOX_DECL
+bool
+isTruthy(dom::Value const& arg);
```
vinniefalco, line 846: We can make this a member of `Value` if you like

include/mrdox/Support/Handlebars.hpp
```diff
@@ -0,0 +1,1037 @@
...
+   @param data The object to look up the property in
+   @param path The path to the property to look up
+
+   @return The value of the property, or nullptr if the property does not exist
+ */
+MRDOX_DECL
+dom::Value
+lookupProperty(
+    dom::Value const &data,
+    std::string_view path);
```
vinniefalco, line 880: What is a "path" ? This needs to be documented

include/mrdox/Support/Handlebars.hpp
```diff
@@ -0,0 +1,1037 @@
...
+registerBuiltinHelpers(Handlebars& hbs);
+
+/// "if" helper function
+/**
+ * You can use the if helper to conditionally render a block. If its argument returns false, undefined, null,
+ * "", 0, or [], Handlebars will not render the block.
+ */
+MRDOX_DECL
+dom::Value
+if_fn(
```
vinniefalco, line 916: This is a little weird because you have C++ helpers, which can only be called from C++ code directly. To call them from scripts they need to go through the registration interface and the glue for the scripting language. And for C++ to call helpers that are scripted, they have to go through the registration interface. In other words, C++ calling a scripted helper might look like this (exposition only):
```
handlebars.callHelper( "myhelper", args, cb );
```

Aside from some open questions like how do we know the cb is the right one (shouldn't it be known to the handlebars instance because it is topmost on the "stack" ?)... now there are two different ways for C++ to call helpers. I'm not sure this is the right choice. It would be better if there was uniform syntax for invoking helpers from C++. 

Maybe since we are using C++20 you could make a constexpr version of my `callHelper` above and dispatch to the native helpers if the string matches? Otherwise use the scripting language's call operation to get the parameters onto the js/lua stack and invoke the script.

test/unit/diff.hpp
```diff
@@ -0,0 +1,201 @@
...
+    // Reverse the diff lines to match the original order
+    std::reverse(diffLines.begin(), diffLines.end());
+
+    // Concatenate diff lines into a single string considering number of unchanged lines in the context
+    std::size_t unchanged = 0;
+    auto last_rendered = std::size_t(-1);
+    for (i = 0; i < diffLines.size(); ++i) {
+        auto& diffLine = diffLines[i];
+        if (diffLine.added || diffLine.removed) {
+            std::size_t context_begin = std::max(i - context_size, (last_rendered == std::size_t(-1) ? 0 : last_rendered + 1));
```
vinniefalco, line 161: You aren't protecting `max` from Windows macro shenanigans

test/unit/fixtures/handlebars_features_test.adoc.hbs
```diff
@@ -0,0 +1,395 @@
+== {{ page.name }}
```
vinniefalco, line 1: Do you really want to just put this in one giant file? Or would it be better with each test in its own file? If its easy to diagnose failures with the one big file, then fine

test/unit/handlebars.cpp
```diff
@@ -0,0 +1,460 @@
...
+    }
+};
+
+
+int
+main() {
+    /////////////////////////////////////////////////////////////////
+    // Fixtures
+    /////////////////////////////////////////////////////////////////
+    std::string_view template_path = MRDOX_UNIT_TEST_DIR "/fixtures/handlebars_features_test.adoc.hbs";
```
vinniefalco, line 55: What the hell...

test/unit/handlebars.cpp
```diff
@@ -0,0 +1,460 @@
...
+
+    HandlebarsOptions options;
+    options.noHTMLEscape = true;
+
+    /////////////////////////////////////////////////////////////////
+    // Context
+    /////////////////////////////////////////////////////////////////
+    dom::Object context;
+    dom::Object page;
+    page.set("kind", "record");
```
vinniefalco, line 75: Oh wow this whole test looks like a hot mess to maintain... why can't we do something like what we did for mrdox where you have some input data, a small template, and an expected output file then the test runs handlebars on the template and compares the actual output from the expected output file. Then you never need to change the C++ to add tests.

test/unit/handlebars.cpp
```diff
@@ -0,0 +1,460 @@
...
+    context.set("lookup_test2", lookup_test2);
+
+    /////////////////////////////////////////////////////////////////
+    // Register helpers
+    /////////////////////////////////////////////////////////////////
+    Handlebars hbs;
+    helpers::registerAntoraHelpers(hbs);
+    hbs.registerHelper("progress", [](dom::Array const& args) -> dom::Value {
+        if (args.size() < 3) {
+            return fmt::format("progress helper requires 3 arguments: {} provided", args.size());
```
vinniefalco, line 220: How does this error reporting work? Errors are strings? I don't understand this...

test/unit/handlebars.cpp
```diff
@@ -0,0 +1,460 @@
...
+    hbs.registerPartial("dynamicPartial", "Dynamo!");
+    hbs.registerPartial("lookupMyPartial", "Found!");
+    hbs.registerPartial("myPartialContext", "{{information}}");
+    hbs.registerPartial("myPartialParam", "The result is {{parameter}}");
+    hbs.registerPartial("myPartialParam2", "{{prefix}}, {{firstname}} {{lastname}}");
+    hbs.registerPartial("layoutTemplate", "Site Content {{> @partial-block }}");
+    hbs.registerPartial("pageLayout", "<div class=\"nav\">\n  {{> nav}}\n</div>\n<div class=\"content\">\n  {{> content}}\n</div>");
+
+    /////////////////////////////////////////////////////////////////
+    // Render and diff
```
vinniefalco, line 429: Oh... okay, so you are doing the render and then diff with the source data but where is the folder with all the triplets of files (input file, template file, expected output file)?

include/mrdox/Support/Dom.hpp
alandefreitas: Assigning to arrays is important because there are many situations where we need to (re)populate arrays to create the callbacks, call helpers, etc... and

```
implicitly-declared ‘clang::mrdox::dom::Array& clang::mrdox::dom::Array::operator=(const clang::mrdox::dom::Array&)’ is deprecated because ‘clang::mrdox::dom::Array’ has user-provided ‘clang::mrdox::dom::Array::Array(const clang::mrdox::dom::Array&)’
   89 |     Array(Array const& other);
```

include/mrdox/Support/Handlebars.hpp
alandefreitas: I didn't give this much thought. The previous version was appending to a string and this version was just an incremental improvement because I needed to append to other things at some points. 

What are the alternatives?

include/mrdox/Support/Handlebars.hpp
alandefreitas: In doxygen HTML `{.handlebars}` is basically just indicating this is not `cpp` because it can't highlight handlebars.

> If the contents of the code block are in a language that doxygen cannot parse, doxygen will just show the output as-is. You can make this explicit using .unparsed, or by giving some other extension that doxygen doesn't support, e.g.

Outside the context of doxygen or when using doxygen XML to generate something else, then it can be useful to render it as a handlebars source block instead of C++. The logic needs to be implemented but the information is there.

include/mrdox/Support/Handlebars.hpp
alandefreitas: > which can only be called from C++ code directly

I'm not sure I understand this problem. In theory these helpers *could* be called from C++ because they're valid, but they are not meant to be used directly. They are implementation details only meant to be called from templates. 

The templates use the `if_fn` because it's in a format in understands. But if I want an `if` statement in C++ or Javascript, I'll just use the `if` keyword without going through any `if_fn`. 

> now there are two different ways for C++ to call helpers

What two ways?

> uniform syntax for invoking helpers from C++.

I confirmed I'm not understanding any of this. 

There's no syntax to invoke helpers, unless you mean the helper signature handlebars expects, which is a single uniform syntax with two parameters: the array and the callback.

test/unit/diff.hpp
alandefreitas: Oh... I thought that wasn't a thing anymore.

test/unit/fixtures/handlebars_features_test.adoc.hbs
alandefreitas: I would have multiple files and unit tests if this was somewhat more independent from mrdox. But for now I just have a few macros I'm using, no test framework, and it doesn't scale very well.

test/unit/handlebars.cpp
alandefreitas: :man_shrugging:

test/unit/handlebars.cpp
alandefreitas: That's kind of what I would do if the project was more independent. I would have libraries to parse json, some nice unit testing framework, etc... I'm doing things this way to be less intrusive. I can't just add things to cmakelists.txt, etc.

test/unit/handlebars.cpp
alandefreitas: I haven't worked much on error reporting yet because it's some area where we probably want to deviate from javascript. 

Javascript always throws for any small thing and that's it. And the helpers can throw, so nothing the engine can do about that. Then there are hooks for missing variables and helpers: https://handlebarsjs.com/guide/hooks.html#helpermissing

But helpers can do anything. In this case, the "progress" helper just renders the error instead of throwing it because this is much more useful in tests. Instead of an exception is CI, this becomes just another line where the output was different when we diff.

test/unit/handlebars.cpp
alandefreitas: `test/unit/fixtures`

include/mrdox/Support/Handlebars.hpp
vinniefalco: Would you believe that I was thinking of `co_await` ? :)

include/mrdox/Support/Handlebars.hpp
vinniefalco: Oh.. if the helpers are not meant to be called directly from C++ then ignore. Just make sure those things are private.

test/unit/fixtures/handlebars_features_test.adoc.hbs
vinniefalco: Yeah that's fair. We will need to build up the tests (for mrdox and handlebars).

include/mrdox/Support/Handlebars.hpp
alandefreitas: Well... that's really another level of complexity. :)

include/mrdox/Support/Handlebars.hpp
alandefreitas: What I did is I include functions such as

```
MRODX_DECL
void
registerAntoraHelpers(Handlebars& hbs);
```

that will register groups of helper functions that should be optional or at least have the logic separated. The individual helpers are just another option, but in any case they would always be registered directly with the engine. There's no point in calling them outside the templates to performs tasks because it's always going to be better to perform these tasks directly.

In the future, I was thinking of other similar groups of helper functions like `registerStringHelpers` with things like `join`, `split`, etc... because most applications will need these small helpers for string manipulation anyway.

alandefreitas:
So... I believe this is ready to be used so we can further improve it with the real problems we find. This implementation should be able to handle any working handlebars template currently being used with handlebars.js. It includes an implementation that:

- replicates the complete javascript API, 
- supports all tag features (expressions, private data syntax, whitespace control, helpers, blocks, chained blocks, block parameters, partials, dynamic partials, inline partials, decorators, raw blocks, partial blocks, partial parameters, and hooks), 
- all built-in helpers from handlebars.js v4 (context switching, lookups, missing helpers defaulting to mustache behavior,  private data syntax, block expressions, and component paths for ".." segments), 
- integrates mrdox's dom types directly, 
- multiple helper signatures to emulate some JS behavior, 
- safe strings, and 
- default sets of optional helpers (antora helpers for logical expressions, string helpers, and helpers for composite data types (arrays and objects)).

These are the next features I'm considering for other PRs starting from the highest priority:

- **Complete specs tests**: Although all features are working correctly, the complete tests would focus on fixing whitespace and indentation issues in corner cases. All the rest works OK. The way it indents and removes whitespaces in corner cases is where most of the incompatibility will come from. This involves merging the test macros I wrote for handlebars with the unit test functionality from boost.url. Other HandlebarsOptions will only the activated at this step.
- **Review error propagation coherent with requirements**: handlebars.js just throws when there's an error. In our case, we need to differentiate between errors and exceptional errors. There are no errors to throw when rendering templates. Errors only come from the helpers, which are black-boxes to the handlebars environment. So we might need to come up with some convention helpers can use to propagate errors so that the engine can handle that at a higher level.
- **Helper "overloads"**: because we're replicating a JS API, all helpers starts with some boring and error-prone boilerplate checking the type of each argument. This is quite repetitive and very easy to automate if we know the types a helper requires. We already accept a few different helper signatures which are differentiated with `std::is_invocable`, which the environment needs to know what to do with the results from the helpers, but we don't yet have a system to identify which parameter types the helper expects as input. Determining the inputs is a little more complex than that. If the helpers don't have overloads, we can use something like Boost.CallableTraits' `args` to identify the input types, pre-validate them, and automatically return the errors with any convention required.
- **Incremental rendering**: Peter's Boost.Mustache implements incremental rendering in terms of the template input. You can provide template substrings until the process is complete. On the other hand, we might need an incremental implementation in terms of a buffer, so that the process will stop when the buffer is filled and we can take it from there. Both implementations are not very complex. The biggest complexity here is that these generators need to be recursive. Aside from that, it's all just a matter of storing the rendering state. The biggest problem we actually have is that handlebars.js not only allows helpers that work as black boxes but the environment depends on these helpers for most functionality. If the helpers are black boxes and most functionality is behind helpers, incremental rendering would be useless. The only solution would be to have a system where helpers can also opt-in to render content incrementally. Built-in block helpers would need to opt-in this system. Other helpers for simple tasks, to lowercasing strings, can still just be regular functions.
- **Compiling templates**: The benefit of compiled templates in C++ is very limited although they are not completely useless. Handlebars.js compiles templates to actual javascript code to render the compiled template. As rendering templates directly would be quite slow in JS, the compiled templates are very useful. In C++, we could, in principle, have functionality that would generate another C++ file to render the template, but that's outside the scope of any applications we have for now. Something closer to that would be a "pre-compiled" template, which would store the template information as opcodes. In this case, there's not a lot to gain, but there are some small gains such as trimming some whitespaces in advance and identifying block endings in the template without iterating the block twice. Since iterating sequential strings is already very cheap, the implementation would need to be very careful to ensure the cost of this data structure for the pre-compiled template does not outweigh the small gain we have on pre-processed strings. We can also adapt our OutputRef to store tags instead of strings and let it do both things: create pre-compiled templates or just render the string. It's important to note that pre-compiled templates cannot ensure there are not going to be errors when rendering the template because changing the context can trigger unforeseen errors, partials can be dynamic, and templates rely on block boxes helpers for most functionality.
- **Documentation**: I've tried to document whatever I could directly in the source code. So when mrdox runs on itself, there's going to be a lot of documentation available. Whenever this implementation has a better place for its documentation to go, then I'll move this content accordingly.
- **Make this a standalone library**: This is a low-hanging fruit since the logic is all here already. The only implication is we would need to readapt the value types to something more general that could type-erase maps, strings, etc... Also, any C++20 features we decide to use here would need to be removed and reimplemented, so some future decisions we make for mrdox might kind of invalidate the idea of a standalone library. In any case, this has very low priority and I'm mostly focused on supporting the mrdox requirements for now.

alandefreitas:
Merged

PR #332, opened 2023-06-21
feat: can filter namespaces.

klemens-morgenstern:
First draft, nested & non-relative namespaces don't work well yet.

### DIFF
--- a/lib/AST/ASTVisitor.cpp
+++ b/lib/AST/ASTVisitor.cpp
@@ -869,7 +869,89 @@ shouldExtract(
 
     IsFileInRootDir_ = true;
 
-    return true;
+    if (config_.excludeNamespaces.empty())
+        return true;
+
+    const Decl * scope = nullptr;
+
+    const auto isScope (D->getKind() == Decl::Namespace ||  D->getKind() == Decl::CXXRecord);
+    if (isScope)
+        scope = D;
+    else
+        scope = dyn_cast<Decl>(D->getDeclContext());
+
+    return !shouldExtractNamespace(scope);
+}
+
+bool
+ASTVisitor::
+shouldExtractNamespace(const Decl * D)
+{
+    // Step 1: Check if we're talking about a namespace
+    if ((D == nullptr) || (
+            D->getKind() != Decl::Namespace))
+        return false;
+
+    auto itr = namespaceFilter_.find(D);
+    if (itr != namespaceFilter_.end())
+        return itr->second;
+
+    // check if any parent is skipped already
+    auto nmd = dyn_cast<NamedDecl>(D);
+    bool filtered = false;
+
+    if (nmd != nullptr)
+    {
+        for (const auto & ft : config_.excludeNamespaces)
+        {
+            std::string_view ff{ft};
+            if (!ff.ends_with(nmd->getName()))
+                continue;
+
+            ff.remove_suffix(nmd->getName().size());
+            if (!ff.ends_with("::") && !ff.empty())
+                continue;
+            auto ctx = nmd->getDeclContext();
+            while(!ff.empty())
+            {
+                if (ff == "::") // absolute
+                {
+                    filtered = ctx->isTranslationUnit();
+                    break;
+                }
+
+                if (!ctx->isNamespace())
+                    break;
+
+                auto ns = cast<NamespaceDecl>(ctx);
+                ff.remove_suffix(2u); // remove trailing "::"
+                if (!ff.ends_with(ns->getName()))
+                    break;
+
+                ff.remove_suffix(ns->getName().size());
+                if (!ff.ends_with("::") && !ff.empty())
+                    break;
+
+                ctx = ctx->getParent();
+            }
+
+            filtered |= ff.empty();
+            if (filtered)
+                break;
+        }
+    }
+
+    if (!filtered)
+    {
+        auto ctx = D->getDeclContext();
+        if (ctx == nullptr)
+            return true;
+        auto outer = dyn_cast<Decl>(ctx);
+        filtered = shouldExtractNamespace(outer);
+    }
+
+    namespaceFilter_.emplace(D, filtered);
+    return filtered;
 }
 
 bool
--- a/test-files/filter/all_namespaces.cpp
+++ b/test-files/filter/all_namespaces.cpp
@@ -1,0 +1,22 @@
+struct thingy {};
+
+
+
+namespace detail
+{
+
+struct thingy_impl {};
+
+}
+
+namespace other
+{
+struct thingy { };
+
+
+namespace detail {
+
+struct thingy_impl {};
+
+}
+}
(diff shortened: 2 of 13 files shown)


source/Tool/CorpusImpl.cpp
```diff
@@ -185,11 +185,109 @@ build(
...
+                if (info != nullptr)
+                {
+                    if (info->isNamespace())
+                        Members = &static_cast<NamespaceInfo *>(info)->Members;
+                    else if (info->isRecord())
+                        Members = &static_cast<RecordInfo *>(info)->Members;
+                }
+                if (Members)
+                    std::erase(*Members, im.second->id);
+                InfoMap.erase(im.first());
```
vinniefalco, line 282: no you can't do this. the filtering needs to happen during top-level visitation. and even so,  sometimes things can slip through. For example consider the following:
```
namespace detail {
struct Q {};
};

struct P : detail::Q {};
```

In this scenario we should filter `detail::Q` when we see it at the top level during visitation. But when we extract "P", we cannot avoid extracting base classes unless they are private. So then we have to extract `detail::Q` even though it would fail the filter.

You are also not generally allowed to modify the InfoMap after it is created. This all has to be done in ASTVisitor.cpp

include/mrdox/Corpus.hpp
```diff
@@ -135,6 +135,10 @@ class MRDOX_VISIBLE
     getFullyQualifiedName(
         const Info& I,
         std::string& temp) const;
+
+    MRDOX_DECL
+    virtual
+    void filter() = 0;
```
vinniefalco, line 141: This should not be a public API in the first place.

source/AST/ASTVisitor.cpp
```diff
@@ -622,7 +622,92 @@ shouldExtract(
...
+        scope = D;
+    else
+        scope = dyn_cast<Decl>(D->getDeclContext());
+
+    return !calculateScopeFilter(scope);
+}
+
+bool
+ASTVisitor::
+calculateScopeFilter(const Decl * d)
```
vinniefalco, line 641: capital D, why don't you follow the style?

source/AST/ASTVisitor.cpp
```diff
@@ -622,7 +622,92 @@ shouldExtract(
...
+    if ((d == nullptr) ||
+        (d->getKind() != Decl::Namespace
+      && d->getKind() != Decl::Record))
+        return false;
+
+    auto itr = scopeFilter_.find(d);
+    if (itr != scopeFilter_.end())
+        return itr->second;
+
+    // check if any parent is skipped already
```
vinniefalco, line 653: I suspect this might not be needed, if we are in fact traversing top down as we should be.

source/AST/ASTVisitor.hpp
```diff
@@ -57,6 +57,10 @@ class ASTVisitor
         clang::SourceLocation::UIntTy,
         FileFilter> fileFilter_;
 
+    std::unordered_map<
+        const Decl*,
+        bool> scopeFilter_;
```
vinniefalco, line 62: Consider this instead:
```
std::unordered_map< NamespaceDecl const*, bool > shouldExtract_;
```

source/AST/ASTVisitor.hpp
```diff
@@ -218,6 +222,9 @@ class ASTVisitor
         TypedefInfo& I,
         DeclTy* D);
 
+    bool
+    calculateScopeFilter(const Decl * d);
```
vinniefalco, line 226: Consider
```
bool shouldExtractNamespace( NamespaceDecl const* D );
```

source/AST/ASTVisitor.cpp
```diff
@@ -622,7 +622,92 @@ shouldExtract(
...
+}
+
+bool
+ASTVisitor::
+calculateScopeFilter(const Decl * d)
+{
+    // Step 1: Check if we're talking about a scope type
+    if ((d == nullptr) ||
+        (d->getKind() != Decl::Namespace
+      && d->getKind() != Decl::Record))
```
vinniefalco, line 646: I don't like this new style of putting things in the left margin. Consider:
```
if ( ! D || (
        D->getKind() != Decl::Namespace &&
        D->getKind() != Decl::Record ) )
    return false;
```

source/AST/ASTVisitor.cpp
```diff
@@ -622,7 +622,92 @@ shouldExtract(
...
+    if (!filtered)
+    {
+        auto ctx = d->getDeclContext();
+        if (ctx == nullptr)
+            return true;
+        auto outer = dyn_cast<Decl>(ctx);
+        filtered = calculateScopeFilter(outer);
+    }
+
+    scopeFilter_.emplace(d, filtered);
```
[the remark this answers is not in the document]
klemens-morgenstern: It's also filtering records at this moment.

source/AST/ASTVisitor.cpp
klemens-morgenstern: Is that true if we renter a namespace later?

source/AST/ASTVisitor.hpp
klemens-morgenstern: Why?

klemens-morgenstern:
Addressed most issues.

source/AST/ASTVisitor.hpp
vinniefalco: is it not self-evident?

source/AST/ASTVisitor.cpp
vinniefalco: how can we re-enter a namespace?

vinniefalco: @sdkrystian can you please verify this

lib/AST/ASTVisitor.cpp
```diff
@@ -869,7 +869,89 @@ shouldExtract(
...
+}
+
+bool
+ASTVisitor::
+shouldExtractNamespace(const Decl * D)
+{
+    // Step 1: Check if we're talking about a namespace
+    if ((D == nullptr) || (
+            D->getKind() != Decl::Namespace))
+        return false;
```
vinniefalco, line 893: Should this be `MRDOX_ASSERT` instead?

test-files/filter/all_namespaces.cpp
```diff
@@ -0,0 +1,22 @@
...
+namespace detail
+{
+
+struct thingy_impl {};
+
+}
+
+namespace other
+{
+struct thingy { };
```
vinniefalco, line 14: no please no "thingies" in the test fixtures :)

test/TestAction.cpp
```diff
@@ -129,21 +130,27 @@ TestRunner::
...
+            "  xml:\n"
+            "    index: false\n"
+            "    prolog: true\n";
+
+        config = loadConfigString(
+            workingDir, testArgs.addonsDir, configYaml);
+    }
     if (!config)
-      reportError(config.error(), "load the configuration string");
+        reportError(config.error(), "load the configuration string");
```
vinniefalco, line 153: The message doesn't match the action if you go through `loadConfigFile`.

vinniefalco:
Please rebase this against develop and use the nested test config feature instead

alandefreitas:
Obsoleted by #432

PR #350, opened 2023-06-26
feat: warnings for duplicates

### LINKED ISSUE #266: Emit warnings for duplicate `@returns`/`@param`/`@tparam` etc.
If multiple `@returns` directives appear within a single javadoc, a warning should be emitted. The same applies for `@param` and `@tparam`. Additionally a warning should be issued when the parameter name specified in a `@param` directive does not match any of the function parameters. 

These warnings should not be emitted more than once for a given declaration; this would require tracking such warnings on a per-source location basis.

klemens-morgenstern:
closes #266

### DIFF
--- a/source/AST/ParseJavadoc.cpp
+++ b/source/AST/ParseJavadoc.cpp
@@ -422,6 +422,16 @@ visitBlockCommandComment(
         doc::Returns returns;
         Scope scope(returns, paragraph_);
         visitChildren(C->getParagraph());
+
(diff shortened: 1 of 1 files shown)


sdkrystian (approved):

vinniefalco:
This has to be rebased, the javadoc code has changed significantly as with the system of reporting warnings

alandefreitas:
Obsoleted by a6910529

PR #382, opened 2023-07-07
separate ASTVisitor from SemaConsumer

sdkrystian:


include/mrdox/Metadata/Function.hpp
```diff
@@ -137,8 +137,6 @@ struct FunctionInfo
     : IsInfo<InfoKind::Function>
     , SourceInfo
 {
-    friend class ASTVisitor;
```
vinniefalco, line 140: Good

include/mrdox/Metadata/Record.hpp
```diff
@@ -77,8 +77,6 @@ struct RecordInfo
     : IsInfo<InfoKind::Record>
     , SourceInfo
 {
-    friend class ASTVisitor;
```
vinniefalco, line 80: Good.

PR #386, opened 2023-07-09
chore: using fmt::format string to check formats at CT

klemens-morgenstern:


include/mrdox/Support/Error.hpp
```diff
@@ -39,7 +39,7 @@ struct Located
 
     template<class Arg>
     requires std::is_constructible_v<T, Arg>
-    Located(
+    consteval Located(
```
vinniefalco, line 42: fuuuuck... I was using `constexpr`? Is that no good anymore? wtf is `consteval` on a function?

vinniefalco:
so... this doesn't build

PR #396, opened 2023-07-10
fix(handlebars): inverse blocks with no helpers expand expressions

alandefreitas:
This fixes bugs such as

```
{{#with symbol}}
    {{> (lookup . 'kind') symbol=.}}
{{/with}}
```

where the partial resolves to

```
{{#if symbol.template}}
    {{>template-head symbol.template}}
    {{symbol.tag}} {{symbol.name~}}
    {{#if (neq symbol.template.kind "primary")~}}
        {{>template-args args=symbol.template.args}}
    {{/if}}
{{else}}
    {{symbol.tag}} {{symbol.name~}}
{{/if}}
```

which should expand the expressions `{{symbol.tag}} {{symbol.name~}}` but it currently rendering the block as is.

Inverse blocks are allowed to have their own helpers (for instance, `{{^ with ...}}`), but when this feature was implemented, the implementation wrongly assumed inverse blocks without a helper should be rendered as is, while it should actually be rendered as a block in the parent context.


vinniefalco:
I merged this from my own branch, it fixed things !

PR #398, opened 2023-07-10
test: integrate handlebars and boost.url test suite

alandefreitas:
This change integrates the test library developed for handlebars with the test library from boost.url that is being used for other tests.

With this integration, Handlebars tests are listed among any other tests in mrdox.

The features used for handlebars, such as the expression decomposer and the diff algorithm for golden master tests, have been integrated into the test library from Boost.URL. The diff algorithm is in an optional header.


lib/Support/Handlebars.cpp
```diff
@@ -131,7 +131,7 @@ escapeExpression(
     std::string_view str,
     HandlebarsOptions opt)
 {
-    if (opt.noHTMLEscape)
+    if (opt.noEscape)
```
vinniefalco, line 134: Please auto-close the issue as "fix" in the commit message

alandefreitas: OK. Is that all?

vinniefalco: for the tests yes but I need to look at the source tree refactoring

src/tool/GenerateAction.cpp
```diff
@@ -10,10 +10,10 @@
...
 
 #include "ToolArgs.hpp"
-#include "Lib/AbsoluteCompilationDatabase.hpp"
-#include "Lib/ConfigImpl.hpp"
-#include "Lib/CorpusImpl.hpp"
-#include "Lib/ToolExecutor.hpp"
+#include <lib/Lib/AbsoluteCompilationDatabase.hpp>
+#include <lib/Lib/ConfigImpl.hpp>
+#include <lib/Lib/CorpusImpl.hpp>
+#include <lib/Lib/ToolExecutor.hpp>
```
vinniefalco, line 16: I would use quotes for private includes but its just a cosmetic thing...

vinniefalco:
This is starting to look a LOT more organized than what we have in develop! I would like to see these tweaks though, because we are mixing up the "not-a-framework" files and the mrdox-test program files. 

TestArgs.cpp, TestArgs.hpp, TestMain.cpp, TestRunner.cpp, TestRunner.hpp are all considered sources for the mrdox-test.exe target not the "not-a-framework" so these should go here:

src/test/TestArgs.cpp
src/test/TestArgs.hpp
src/test/TestMain.cpp
src/test/TestRunner.cpp
src/test/TestRunner.hpp

detail/decomposer.cpp, detail/decomposer.hpp, test_suite.cpp, and test_suite.hpp are sources which belong to the "not-a-framework" unit testing helpers. I would imagine that we will make this into its own repository (which will remain _lean_ in the spirit of lightweight test but with a little more more horsepower to get the job done). And then we would use git-subtree or git-submodule to inline the sources into the target repository. In that case we want it to go to its own directory which has a unique prefix. I don't know what that prefix but for now I guess "test_suite/"? Maybe like this for now and we can decide if we want to change it later:

src/test_suite/test.hpp     // renamed from test_suite.hpp
src/test_suite/test.cpp     // renamed from test_suite.cpp
src/test_suite/detail/decomposer.hpp
src/test_suite/detail/decomposer.cpp

diff.cpp, and diff.hpp would be an implementation detail of the hypothetical "handlebars-test.exe" which we have not created yet and is currently fused to mrdox-test.exe so we could leave them in the locations below for now, until handlebars gets its own repository and then those files would be moved to become part of that project:

src/test/diff.hpp
src/test/diff.cpp

src/tool/GenerateAction.cpp
alandefreitas: OK. I'll undo that.

I usually use quotes when it's private and it's in the same directory because of how quotes invert the search procedure.

PR #403, opened 2023-07-12
handlebars basic spec tests

alandefreitas:
This PR replicates the basic spec tests from https:github.com/handlebars-lang/handlebars.js/blob/4.x/spec/basic.js

Fortunately the C++ implementation passed most tests without any changes. A few bugs in corner cases related to double escapes, specific types of whitespace control, and tag options were fixed.

On the other hand, Handlebars.js includes many undocumented features that are only being implemented now as I find out about them. For instance, `&` can be used in tags as an alternative for disabling HTML escaping. This does not seem to be documented anywhere.

There are also cases that can't be supported by the C++ implementation right now but are worth to thinking about. Especifically, Handlebars.js allows the context objects to include functions that are executed as helpers. This also doesn't seem to be documented anywhere, I have never seen that being used, and it's possible to use regular helpers instead. Nonetheless, it's a supported feature that is tested in the original test suite.

### DIFF
--- a/include/mrdox/Support/Handlebars.hpp
+++ b/include/mrdox/Support/Handlebars.hpp
@@ -18,6 +18,7 @@
 #include <functional>
 #include <type_traits>
 #include <vector>
+#include <variant>
 
 namespace clang {
 namespace mrdox {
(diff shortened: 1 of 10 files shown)


vinniefalco:
squash it all down and remove the text "(handlebars)" from the commit message

PR #410, opened 2023-07-14
SSH upload to website

sdarwin:
A functional example of uploading files to the server dev-websites.cpp.al which is hosting https://mrdox.com from the directory /var/www/mrdox.com/

The step "Add SSH key" should remain as-is. 
The step "Copy files" is uploading files to /var/www/mrdox.com/example1/ .   Replace with other commands. Upload to /var/www/mrdox.com/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -266,3 +266,34 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+  website:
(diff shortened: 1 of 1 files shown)


sdarwin:
The job is failing. The reason is most likely that secrets "... are not passed to workflows that are triggered by a pull request from a fork."  It should be moved to a CI workflow only triggered by branches and not PRs.

alandefreitas:
Integrated in #412

Issue #405, opened 2023-07-14
no fetchcontent


vinniefalco:
No CMakeLists.txt anywhere in Boost or CPPAlliance should require the Internet to execute.

alandefreitas:
I think this fixes itself when duktape is removed.

vinniefalco:
> I think this fixes itself when duktape is removed.

1. I'm not removing duktape, and
2. I want it removed from all CMakeLists.txt for any of our projects (URL, JSON, Beast, http-proto, http-io, buffers, mrdox, handlebars, static-string)

I am so sick of Visual Studio hanging for 5 minutes where I have no control

alandefreitas:
> I'm not removing duktape, and

Yes. I already fixed that in a PR I'm opening soon in any case.

> I want it removed from all CMakeLists.txt for any of our projects (URL, JSON, Beast, http-proto, http-io, buffers, mrdox, handlebars, static-string)

There's no fetch content in any other library. Any other problem your having is unrelated to this.

PR #412, opened 2023-07-15
WIP: Demos

alandefreitas:
This PR includes demos in the mrdox website. It also includes a number of fixes that are necessary to make this happen.

- [x] Documentation includes usage instructions
- [x] MrDox reference config is multipage
- [x] Installation target is fixed and also respects the FHS directory layout now
- [x] Include and install CMake module to create mrdox targets
- [x] Include a temporary non-intrusive heuristic to filter source files
- [x] Include mrdox self-reference in antora docs
- [ ] Integration tests using the binary packages
- [ ] Use this infrastructure to render the docs for other libs, bypass antora for these, and upload them to mrdox.com

The biggest problem here is the heuristic to filter source files. Until this is properly implemented, we are blocked. I included a temporary implementation to render fewer source files based on the sourceDir, but Antora is still having a hard time and even the GitHub interface is stuttering to show the content of these commits. An appropriate solution would depend on appropriate filters in the config file, like the doxygen filters, and the source files should have more utterable and representative names. We can wait for this better solution or move on with this.

Another smaller issue is the multipage documentation has no index to refer to in the antora nav.


alandefreitas:
Since there's no index.adoc to refer to in the antora nav, links to the reference can only be accessed through the direct links. For instance, https://412.mrdox.prtest2.cppalliance.org/mrdox/HEAD/reference/0DE5AF87692A841A9389BF991987BDFD1772CF5D.html, https://412.mrdox.prtest2.cppalliance.org/mrdox/HEAD/reference/2EC1D42B977544E441FD64B47925C52E9D3698D5.html, ...

vinniefalco:
Okay there is no intention to actually merge this right? Because this pull request is insane

vinniefalco:
Why does this commit the Duktape sources to the repo?

alandefreitas:
> Okay there is no intention to actually merge this right? 

It says WIP. 

> Because this pull request is insane

The only serious issue I see here is the necessity of filters to generate demos that make sense. We can't have demos without filters. That's a blocker. Antora can't render that monster it generates now.

But I wouldn't call anything else "insane". It's all justified.

> Why does this commit the Duktape sources to the repo?

Probably I misunderstood what "not requiring internet" means in your issue #405. There are only 3 options to integrate any dependency I know of:

1) dependency is bundled with the source code (directly or with gitmodules)
2) dependency is not bundled, and it comes from the environment (be it manually installed, via a package manager, or sources when there's no working package) at the configure step (that was DUKTAPE_SOURCE_ROOT)
3) dependency is not bundled, and it is optionally downloaded at configure time and cached (which is what happened when DUKTAPE_SOURCE_ROOT was not set)

We had 2 and 3, which you didn't like. So changed it to 1 here because I had to fix the install target but you also don't like it. I can also safely assume you also don't like 2 falling back to 3, since the issue is called "[no fetchcontent](https://github.com/cppalliance/mrdox/issues/405)" and fetchcontent was removed for fmt lately by someone else.

So I don't know of any other alternatives and I'm stuck because if I don't explain myself I will still not understand the requirements, but if I explain myself I'll probably hear something like "that's try-hard - just do the thing" and will still not know what the "thing" is.

vinniefalco:
The "insanity" is that this pull request changes over 2,500 files...

Duktape should be installed like everything else, vcpkg or manually, and then cmake finds it somehow.

alandefreitas:
> Duktape should be installed like everything else, vcpkg or manually, and then cmake finds it somehow

That's what we have now. Just set `DUKTAPE_SOURCE_ROOT` in your `cmakeCommandArgs` and you're good to go:

```
"cmakeCommandArgs": "-DCMAKE_PREFIX_PATH=\"C:\\Users\\vinnie\\src\\llvm-install\\RelWithDebInfo\" -DDUKTAPE_SOURCE_ROOT=\"C:\\Users\\vinnie\\src\\duktape"",
```

It doesn't expect a vcpkg package because the package is broken and unmaintained. `duktape` is not really well maintained anymore. It also doesn't expect a CMAKE_PREFIX_PATH like LLVM does because `duktape` is a C project that provides no CMake integration.

vinniefalco:
Which is more modern, CMakePresets or CMakeSettings?

vinniefalco:
Where do I get the latest duktape that can be consumed by setting `DUKTAPE_SOURCE_ROOT` then?

vinniefalco:
This seems to work but it looks like the directory that FetchContent downloaded (LOL)
```
                "DUKTAPE_SOURCE_ROOT": {
                    "value": "C:/Users/Vinnie/src/mrdox/build/Debug/_deps/duktape-src",
                    "type": "FILEPATH"
```

alandefreitas:
> Which is more modern, CMakePresets or CMakeSettings?
> Edit: Looks like CMakePresets is preferred...

Yes. I was going to suggest moving to CMakePresets. Changing it to CMakePresets is quite simple but I've been postponing this because all the work I would invest in convincing other people would deviate me from most urgent things.

If you convinced yourself already that's much easier. I can include it in this PR since it's related to what it's doing.

> Where do I get the latest duktape that can be consumed by setting DUKTAPE_SOURCE_ROOT then?

I'll document this when making the change to remove fetchcontent. But it's just

```
curl https://github.com/svaarala/duktape/releases/download/v2.7.0/duktape-2.7.0.tar.xz
```

> This seems to work but it looks like the directory that FetchContent downloaded (LOL)

Yes. Since I'm going to remove FetchContent, it will have to be installed somewhere else. Probably `C:\Users\vinnie\src\duktape` in your case.

PR #435, opened 2023-08-01
Handlebars whitespace control specs

alandefreitas:
https://github.com/handlebars-lang/handlebars.js/blob/4.x/spec/whitespace-control.js


vinniefalco:
squash it all down

alandefreitas:
> squash it all down

done

PR #436, opened 2023-08-02
feat: partials specs

alandefreitas:
WIP because these specs are huge. 

And also hugely important.

### DIFF
--- a/include/mrdox/Support/Handlebars.hpp
+++ b/include/mrdox/Support/Handlebars.hpp
@@ -55,6 +55,20 @@ struct HandlebarsOptions
      */
     bool ignoreStandalone = false;
 
+    /** Disables implicit context for partials
+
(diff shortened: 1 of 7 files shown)


alandefreitas:
@vinniefalco No longer WIP. 

This turned out to be quite complex.

vinniefalco:
What does "specs" mean? From the title it sounds like this is a documentation-only change...

alandefreitas:
> What does "specs" mean? From the title it sounds like this is a documentation-only change...

The specification. So that any template being used with handlebars.js has the same results in our implementation.

The handlebars.js repository has a set of tests/fixtures to check if an implementation matches the specification:

https://github.com/handlebars-lang/handlebars.js/tree/4.x/spec

These PRs are implementing each of these tests and fixing anything that needs to be fixed along the way.

PR #437, opened 2023-08-05
ToolExecutor cleanup, natvis improvements, enhanced dependency extraction

sdkrystian:


src/lib/AST/AnyBlock.hpp
```diff
@@ -14,13 +14,195 @@
...
+Error
+decodeRecord(
+    Record const& R,
+    OptionalLocation& Field,
+    llvm::StringRef Blob)
+{
+    if (R[0] > INT_MAX)
+        return formatError("integer value {} too large", R[0]);
+    Field.emplace((int)R[0], Blob, (bool)R[1]);
+    return Error::success();
```
alandefreitas, line 143: We don't have to worry about that in this PR, but the idiomatic way would be to return `Expected<void, Error>` and `return {}` on success instead of a somewhat contradictory and counterintuitive `Error::success()`.

src/lib/Lib/ConfigImpl.cpp
```diff
@@ -32,13 +32,42 @@ template<>
...
+template<>
+struct llvm::yaml::ScalarEnumerationTraits<
+    clang::mrdox::Config::ExtractOptions::Policy>
+{
+    using Policy = clang::mrdox::Config::ExtractOptions::Policy;
+
+    static void enumeration(IO& io,
+        Policy& value)
+    {
+        io.enumCase(value, "always", Policy::Always);
```
alandefreitas, line 51: Include in documentation

src/lib/Lib/ExecutionContext.cpp
```diff
@@ -5,30 +5,164 @@
...
+    InfoSet info = std::move(results);
+    // KRYSTIAN TODO: read stage will be required to
+    // update Info references once we switch to using Info*
+    #if 0
+    {
+        std::shared_lock read_lock(mutex_);
+    }
+    #endif
+
+    std::unique_lock write_lock(mutex_);
```
alandefreitas, line 69: Don't use template argument deduction for `std::unique_lock`

src/lib/Lib/ExecutionContext.hpp
```diff
@@ -28,23 +34,88 @@ namespace mrdox {
...
 */
 class ExecutionContext
-    : public tooling::ExecutionContext
 {
-    llvm::sys::Mutex mutex_;
-    Diagnostics diags_;
+protected:
+    const ConfigImpl& config_;
 
 public:
```
alandefreitas, line 41: Define `virtual ~ExecutionContext() = default`

src/lib/Lib/ExecutionContext.hpp
```diff
@@ -28,23 +34,88 @@ namespace mrdox {
...
+
+class InfoExecutionContext
+    : public ExecutionContext
+{
+    std::shared_mutex mutex_;
+    Diagnostics diags_;
+    InfoSet info_;
+
+public:
+    using ExecutionContext::ExecutionContext;
```
alandefreitas, line 73: Define `virtual ~InfoExecutionContext() = default`

src/lib/Lib/ExecutionContext.hpp
```diff
@@ -28,23 +34,88 @@ namespace mrdox {
...
+    : public ExecutionContext
+{
+    std::mutex mutex_;
+    Diagnostics diags_;
+
+    std::unordered_map<SymbolID,
+        std::vector<llvm::SmallString<0>>> bitcode_;
+
+public:
+    using ExecutionContext::ExecutionContext;
```
alandefreitas, line 99: Define virtual destructor

alandefreitas (changes requested):
Nice. That's going to be very useful.

include/mrdox/Config.hpp
```diff
@@ -42,25 +42,59 @@ class MRDOX_DECL
...
+            Dependency,
+            Never
+        };
+
+        /** Extraction policy for references to external declarations.
+
+            This determines how declarations which are referenced by
+            explicitly extracted declarations are extracted.
+
+            Given a function parameter of type `std::string`, `std::string`
```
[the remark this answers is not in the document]
sdkrystian: The naming might be a bit misleading... `ExtractOptions::Dependency` means "only extract if this was used as a base class, or if this is a local class which was deduced as the return type of a function with a placeholder for a deduced return type.

`ExtractOptions::Always` means to extract if it was referenced in any declaration.

src/lib/Lib/ExecutionContext.cpp
```diff
@@ -5,30 +5,164 @@
...
 namespace clang {
 namespace mrdox {
 
+namespace {
+
+// A standalone function to call to merge a vector of infos into one.
+// This assumes that all infos in the vector are of the same type, and will fail
+// if they are different.
+// Dispatch function.
+mrdox::Expected<std::unique_ptr<Info>>
```
vinniefalco, line 27: why the `mrdox` namespace qualifier?

src/lib/Lib/ExecutionContext.hpp
```diff
@@ -28,23 +34,88 @@ namespace mrdox {
...
-    void reportEnd(report::Level level);
+    virtual
+    void
+    report(
+        InfoSet&& info,
+        Diagnostics&& diags) = 0;
+
+    virtual
+    void
+    reportEnd(report::Level level) = 0;
```
vinniefalco, line 56: Why is this virtual now?

src/lib/Lib/ConfigImpl.cpp
```diff
@@ -32,13 +32,42 @@ template<>
...
+struct llvm::yaml::MappingTraits<
+    clang::mrdox::Config::ExtractOptions>
+{
+    static void mapping(IO &io,
+        clang::mrdox::Config::ExtractOptions& opts)
+    {
+        io.mapOptional("referenced-declarations", opts.referencedDeclarations);
+        io.mapOptional("anonymous-namespaces", opts.anonymousNamespaces);
+        io.mapOptional("inaccessible-members", opts.inaccessibleMembers);
+        io.mapOptional("inaccessible-bases", opts.inaccessibleBases);
```
vinniefalco, line 67: When you add config settings you have to also update the adoc files.

include/mrdox/Config.hpp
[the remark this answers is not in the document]
alandefreitas: I guess we need better names and docs then. I thought `Dependency` meant that we would also extract the implicit dependencies we discussed some other time. The actual meaning is quite specific. `Dependency` is actually what I thought `Always` was. Is there an option to extract everything it sees?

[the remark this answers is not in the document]
sdkrystian: > Is there an option to extract everything it sees?
Could you clarify? By "everything it sees" does this mean no dependencies, or does it mean everything declaration regardless of origin?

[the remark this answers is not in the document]
alandefreitas: Yes. Isn't that how it worked in the past?

src/lib/Lib/ExecutionContext.hpp
sdkrystian: This is only because on this branch there are two different `ExecutionContext` types: one that serializes to bitcode, and one which does not. Thus, I made `reportEnd` virtual just incase their behavior would differ (it does not). This is a intermediate solution which I will be getting rid of eventually.

src/lib/Lib/ExecutionContext.cpp
sdkrystian: `ExecutionContext.cpp` includes the bitcode serialization headers, and we use `llvm::Expected` in some places in them. I'm using a qualified-id more so for clarity than it actually being necessary

include/mrdox/Config.hpp
```diff
@@ -42,25 +42,59 @@ class MRDOX_DECL
     Config() noexcept;
 
 public:
-    struct Settings
+    struct ExtractOptions
```
vinniefalco, line 45: Should this be public? Would a plugin need to see it? I'm guessing no. Plugins should display what is extracted. I think. So these settings would be an implementation detail and visible to the AST traversal code.

include/mrdox/Config.hpp
```diff
@@ -42,25 +42,59 @@ class MRDOX_DECL
...
+
+            @li `Policy::Dependency`: `private` and `protected` members will only
+            be extracted via dependency.
+
+            @li `Policy::Never`: `private` and `protected` will never be extracted.
+        */
+        Policy inaccessibleMembers = Policy::Always;
+
+        Policy inaccessibleBases = Policy::Always;
+    };
```
vinniefalco, line 90: I'm not sure how I feel about creating a deep hierarchy of settings structs. Can't we just leave it flat? e.g.
```
Policy extractInaccessibleMembers;
Policy inaccessibleBases;
```
And so on?

sdkrystian: I personally prefer nesting because it reduces "config sprawl", especially once we add more of these extraction options. However, if you feel strongly about this I can flatten it.

vinniefalco: These structs aren't used in the plugins, and should probably not be public

vinniefalco: > if you feel strongly about this I can flatten it

I would leave it flat and only make it hierarchical when there is a well demonstrated benefit.

alandefreitas:
About the releases in https://mrdox.com/llvm+clang/:

- They should replace the older releases. Just a different ID. 
- There's no need to include "Minimal" because having it as minimal as possible is the original intention already as long as it works. We (and users) only need the minimal release for mrdox.
- We also need the same variants as before. Otherwise, people (including me) would need to rebuild everything from scratch.
- There's no need for two variants of the same release (tar.xz and tar.gz). We are using 7z for windows and tar.xz for Linux but that can change. But no duplicates because this is expensive. We already had to adapt the server twice to include the previous releases.

There are unresolved comments about the documentation. Please have a look.

PR #442, opened 2023-08-16
change names of CMake configure presets

sdkrystian:


alandefreitas:
For reference: https://cpplang.slack.com/archives/C0508A7LWUV/p1692127447084269

PR #459, opened 2023-10-05
refactor: MrDox uses C++ handlebars

alandefreitas:
This PR refactors the asciidoc and html generators to use the C++ implementation of handlebars. Javascript helpers are loaded with duktape.


sdkrystian:
Looks good to me

PR #466, opened 2023-10-17
[WIP] symbol name lookup

sdkrystian:


include/mrdocs/Metadata/Javadoc.hpp
```diff
@@ -81,14 +85,27 @@ enum class Admonish
...
-enum class ParamDirection : int
+enum class ParamDirection
 {
     none,
     in,
     out,
     inout
 };
 
+/** Which parts of the documentation to copy.
```
alandefreitas, line 96: The part to copy comes from the command (@copydoc, @copybrief, @copydetails). This is only an internal detail, right? Should it be in a .hpp file?

include/mrdocs/Metadata/Javadoc.hpp
```diff
@@ -620,8 +698,11 @@ class MRDOCS_DECL
 
     /** Return the brief, or nullptr if there is none.
     */
-    doc::Paragraph const*
-    brief() const noexcept;
+    const doc::Paragraph*
```
alandefreitas, line 701: Vinnie strongly prefers `T const*` over `const T*` and has given me a hard time because of that in the past.

alandefreitas (approved):
Looks good to me. 

Please ensure CI is passing and squash the fold commits, of course.

I didn't reverse engineer everything because that would be unreasonable. In general, I would ask for more javadocs and at least one test but I won't bother you too much with that since that's the direction we've been going already for Boost.URL.

include/mrdocs/Metadata/Javadoc.hpp
sdkrystian: @alandefreitas We don't have a way to hide implementation detail for `Info` and `doc::Node` derived types right now... they are directly used by generators

PR #474, opened 2023-10-25
Test asan

sdkrystian:


alandefreitas:
IIRC, our last conclusion was that this wasn't possible because Clang is not "ASan-free", right?

alandefreitas:
This is from Oct 25, 2023. Can we close it?

PR #496, opened 2023-11-24
feat: C++/JS proxy functions

alandefreitas:
Define JS proxy types for functions to allow complete interoperability between C++ and JS functions. 

When accessing a JS function in C++, the dom::Function contains a reference to the Scope variable with the JS function. The JS execution context will evaluate this function in JS and convert the result back to dom::Value.

When accessing a C++ function in JS, the JS object will contain a hidden buffer to store a copy of the original dom::Function. When JS executes the function, it will get the original dom::Function from this hidden buffer, convert all arguments to dom::Value, call the original dom::Function, and push the dom::Value result to the stack. When the function is garbage collected, the object finalization callback destroys the function stored in the buffer.


vinniefalco:
nitpick: Capitalization is inconsistent. There is "JavaScript.cpp" and "Javascript.cpp."

It looks like you have gained a deep understanding of `dom` and the duktape integration, as well as a strong grasp on how the architecture is supposed to work. This is evidenced by your ability to make functions "round trip" between C++ and JavaScript. That's great :)

PR #515, opened 2023-12-13
feat: Use compiler default include paths

fpelliccioni:


src/tool/GenerateAction.cpp
```diff
@@ -13,15 +13,108 @@
...
 #include <mrdocs/Generators.hpp>
 #include <mrdocs/Support/Error.hpp>
 #include <mrdocs/Support/Path.hpp>
 #include <clang/Tooling/JSONCompilationDatabase.h>
+
 #include <cstdlib>
+#include <unordered_map>
 
 namespace clang {
 namespace mrdocs {
```
alandefreitas, line 27: I'm under the impression that this is growing too much and we're breaking the separation of concerns here. Maybe all the functionality related to compilers can go into new .h/.cpp files under src/lib/Support. `Compiler.h/.cpp`?

src/lib/Lib/AbsoluteCompilationDatabase.cpp
```diff
@@ -51,14 +51,22 @@ static
...
-    const std::vector<std::string>& additional_defines)
+    const std::vector<std::string>& additional_defines,
+    std::unordered_map<std::string, std::vector<std::string>> const& includePathsByCompiler)
 {
     std::vector<std::string> new_cmdline;
     std::vector<std::string> discarded_cmdline;
     llvm::opt::InputArgList args;
     StringRef driver_mode;
-    if(! cmdline.empty())
+
```
alandefreitas, line 61: I think these extra whitespaces around systemIncludePaths got git confused

src/lib/Lib/AbsoluteCompilationDatabase.cpp
```diff
@@ -51,14 +51,22 @@ static
 std::vector<std::string>
 adjustCommandLine(
     const std::vector<std::string>& cmdline,
-    const std::vector<std::string>& additional_defines)
+    const std::vector<std::string>& additional_defines,
+    std::unordered_map<std::string, std::vector<std::string>> const& includePathsByCompiler)
 {
     std::vector<std::string> new_cmdline;
     std::vector<std::string> discarded_cmdline;
```
alandefreitas, line 58: Does `discarded_cmdline` affect the output at all? Looks like an unused variable. Unless I didn't understand it, references from cmdline are inserted there and there it goes out of scope unused.

(I know that's not you're code but it's worth commenting on this kind of thing in passing to improve things)

src/lib/Lib/AbsoluteCompilationDatabase.cpp
```diff
@@ -51,14 +51,22 @@ static
...
     llvm::opt::InputArgList args;
     StringRef driver_mode;
-    if(! cmdline.empty())
+
+    std::vector<std::string> systemIncludePaths;
+
+    if( ! cmdline.empty())
     {
+        if (auto it = includePathsByCompiler.find(cmdline[0]); it != includePathsByCompiler.end()) {
```
alandefreitas, line 66: Can `cmdline[0]` ever be different since CMake only accepts one toolchain at a time? I've never seen a case where it's different. 

In any case, it's probably fine to generalize this and parse this information for "all compilers we see" even if it's technically always the same compiler. Compilers can still technically be different according to the file format in compile_commands.json.

src/lib/Lib/AbsoluteCompilationDatabase.cpp
```diff
@@ -51,14 +51,22 @@ static
...
+            systemIncludePaths = it->second;
```
alandefreitas, line 67: There's no need to copy it because we're not mutating it. A reference or a pointer (depending on how it's implemented) should be enough here.

src/tool/GenerateAction.cpp
```diff
@@ -13,15 +13,108 @@
...
+    if (auto EC = llvm::sys::fs::createTemporaryFile("compiler-info", "txt", outputPath)) 
+    {
+        return std::nullopt;
+    }
+
+    std::optional<llvm::StringRef> redirects[] = {llvm::StringRef(), llvm::StringRef(), outputPath.str()};
+    std::vector<llvm::StringRef> args = {compiler, "-v", "-E", "-x", "c++", "-"};
+
+    llvm::ErrorOr<std::string> compilerPath = llvm::sys::findProgramByName(compiler);
+    if ( ! compilerPath)
```
alandefreitas, line 41: Will this return `true` when `compiler` is already the full path?

src/tool/GenerateAction.cpp
```diff
@@ -13,15 +13,108 @@
...
+    std::optional<llvm::StringRef> redirects[] = {llvm::StringRef(), llvm::StringRef(), outputPath.str()};
+    std::vector<llvm::StringRef> args = {compiler, "-v", "-E", "-x", "c++", "-"};
+
+    llvm::ErrorOr<std::string> compilerPath = llvm::sys::findProgramByName(compiler);
+    if ( ! compilerPath) 
+    {
+        return std::nullopt;
+    }
+
+    int result = llvm::sys::ExecuteAndWait(*compilerPath, args, std::nullopt, redirects);
```
alandefreitas, line 46: `compiler` is set as the first arg so I have the impression that the compiler path might be passed twice here.

`std::nullopt` makes it inherit the parent environment. Is that the same behavior as CMake (or the behavior we want)? Otherwise, we need to replace it with an empty `ArrayRef<StringRef>`.

Does LLVM provide a function where we can we capture the result directly in a string?

I'm also kind of impressed `std::optional<llvm::StringRef>` is implicitly convertible to `ArrayRef<std::optional<StringRef>>`.

src/tool/GenerateAction.cpp
```diff
@@ -13,15 +13,108 @@
...
+getCompilerInfo(llvm::StringRef compiler) 
+{
+    llvm::SmallString<128> outputPath;
+    if (auto EC = llvm::sys::fs::createTemporaryFile("compiler-info", "txt", outputPath)) 
+    {
+        return std::nullopt;
+    }
+
+    std::optional<llvm::StringRef> redirects[] = {llvm::StringRef(), llvm::StringRef(), outputPath.str()};
+    std::vector<llvm::StringRef> args = {compiler, "-v", "-E", "-x", "c++", "-"};
```
alandefreitas, line 38: `compiler` is an arg?

Does this make sense for MSVC? I have the impression the implicit paths for MSVC are always empty although I've never validated this.

src/tool/GenerateAction.cpp
```diff
@@ -13,15 +13,108 @@
...
 #include <clang/Tooling/JSONCompilationDatabase.h>
+
 #include <cstdlib>
+#include <unordered_map>
 
 namespace clang {
 namespace mrdocs {
 
+std::optional<std::string> 
+getCompilerInfo(llvm::StringRef compiler)
```
alandefreitas, line 29: Does the result of `c++ -v -E -x c++ -` have a more formal name than compiler info? I have the impression this name will be taken soon by a function that returns a complete `struct`.

src/tool/GenerateAction.cpp
```diff
@@ -13,15 +13,108 @@
...
+parseIncludePaths(std::string const& compilerOutput) 
+{
+    std::vector<std::string> includePaths;
+    std::istringstream stream(compilerOutput);
+    std::string line;
+    bool capture = false;
+
+    while (std::getline(stream, line)) 
+    {
+        if (line.find("#include <...> search starts here:") != std::string::npos)
```
alandefreitas, line 73: Is it the same format for all compilers we know about?

alandefreitas (commented):
This is very exciting. As soon as this is merged I'll remove all this logic from the script that generates the boost.url docs: https://github.com/alandefreitas/url/blob/eccc2b85bfecc0d897dca049409cf1b0ddfa2485/doc/generate-files.mjs#L116-L147

alandefreitas:
I forgot to mention, but the logic in these lines should be removed so that the tests ensure mrdocs is finding the appropriate include paths and to avoid including these paths twice.

https://github.com/cppalliance/mrdocs/blob/badb9d9e61ece24abf57efb261301eb290aaee2b/.github/workflows/ci.yml#L329C1-L340

src/tool/GenerateAction.cpp
fpelliccioni: > compiler is set as the first arg so I have the impression that the compiler path might be passed twice here.

You're right to double-check the arguments for llvm::sys::ExecuteAndWait. According to the [LLVM documentation](https://llvm.org/doxygen/namespacellvm_1_1sys.html#abb13fbc7f52c659966b05092b19e6e19), the first element in the Args array should indeed be the name of the program.

src/tool/GenerateAction.cpp
fpelliccioni: GCC and Clang has the same format, yes. This method does not work on MSVC, I will add a `#ifdef` to avoid calling this on MSVC. I guess we will not have that "issue" with that compiler.

src/tool/GenerateAction.cpp
fpelliccioni: there's no formal name for the combination of options used in the g++ -v -E -x c++ - command. It's more of a commonly used trick to extract detailed compiler configuration information. I will change the name of the function with a better name.

src/tool/GenerateAction.cpp
fpelliccioni: Yes, compiler is an argument. As I understand, `ExecuteAndWait` follows the same convention as in the C/C++ main function, where the first argument (`argv[0]`) is the name of the executable.

Regarding MSVC, the approach we use for GCC and Clang to fetch default include paths probably doesn't apply to MSVC. However, I suspect we might not face the issue on MSVC so I'm planning to skip querying the compiler for include paths in the case of MSVC.

src/tool/GenerateAction.cpp
fpelliccioni: > std::nullopt makes it inherit the parent environment. Is that the same behavior as CMake (or the behavior we want)? Otherwise, we need to replace it with an empty ArrayRef<StringRef>.

I've updated the code to use an empty llvm::ArrayRef<llvm::StringRef> instead, ensuring that the child process does not inherit the environment variables from the parent process.

fpelliccioni: > Does LLVM provide a function where we can we capture the result directly in a string?

After extensive testing with various approaches, I've found that the current implementation is the only one that consistently works as intended.

fpelliccioni: > I'm also kind of impressed std::optional<llvm::StringRef> is implicitly convertible to ArrayRef<std::optional<StringRef>>.

Indeed, the `redirects` variable is an array of three elements, each of type `std::optional<StringRef>`

src/tool/GenerateAction.cpp
alandefreitas: >  I will add a #ifdef

Note that this needs to be a runtime check.

src/tool/GenerateAction.cpp
alandefreitas: I imagined the documentation of the `-v -E -x c++` flags would have some name we could use.

src/tool/GenerateAction.cpp
fpelliccioni: Done

src/tool/GenerateAction.cpp
fpelliccioni: Yeah, I had a confusion when I wrote the message here but I did it Ok in the code 👍

fpelliccioni:
> I forgot to mention, but the logic in these lines should be removed so that the tests ensure mrdocs is finding the appropriate include paths and to avoid including these paths twice.
> 
> https://github.com/cppalliance/mrdocs/blob/badb9d9e61ece24abf57efb261301eb290aaee2b/.github/workflows/ci.yml#L329C1-L340

Done! removed

alandefreitas (commented):
Looks great. 

Now that I'm looking at `Compiler.h/cpp`, this looks kind of weird because it sounds like there are functions able to compile code. Maybe `CompilerInfo.h/cpp`? I'm just throwing ideas. I'm not sure. It's up to you.

I'm also a little worried about MSVC compiler identification, although this might work anyway as long as there's no `"#include <...> search starts here:"` in the MSVC output, which is kind of a way of identifying MSVC anyway.

These are just ideas. If you think it's good to go, then it's good to go.

fpelliccioni:
> Looks great.
> 
> Now that I'm looking at `Compiler.h/cpp`, this looks kind of weird because it sounds like there are functions able to compile code. Maybe `CompilerInfo.h/cpp`? I'm just throwing ideas. I'm not sure. It's up to you.
> 
> I'm also a little worried about MSVC compiler identification, although this might work anyway as long as there's no `"#include <...> search starts here:"` in the MSVC output, which is kind of a way of identifying MSVC anyway.
> 
> These are just ideas. If you think it's good to go, then it's good to go.

I removed the MSVC detection (using filename) because it will fail when called with the arguments `-v -E -x c++ -`  and as you said, the output will not contain the pattern we are looking for.

I also renamed `Compiler.h/cpp` -> `CompilerInfo.h/cpp`

alandefreitas:
Great! So this is good to go, right?

fpelliccioni:
Yes

alandefreitas:
Have you checked the output in the CI artifacts? I don't want to merge two things in a row with problems in the output 😆

sdkrystian:
Please squash before you merge!

fpelliccioni:
> Have you checked the output in the CI artifacts? I don't want to merge two things in a row with problems in the output 😆

@alandefreitas @sdkrystian 
I'm checking, I don't see any errors in the CI, it's passing Ok.
I understand that you want me to look at the entire output, but it seems strange to me that we are expecting a pattern in the output and it is not marked as an error. I mean maybe we have to modify the CI to make the error more obvious.
What pattern do you look for in the output to see if everything came out OK or not?
I would like to change the CI scripts in another PR, but I need help identifying the patterns.

fpelliccioni:
@alandefreitas 

I checked the output of the "Demos" job in Github Actions and everything seems fine

https://github.com/cppalliance/mrdocs/actions/runs/7262807028/job/19787446683?pr=515#logs

PR #516, opened 2023-12-14
docs: Corpus javadocs

alandefreitas:
Added Javadocs and did some cleanup incidentally while addressing other issues.

### DIFF
--- a/include/mrdocs/Metadata/Symbols.hpp
+++ b/include/mrdocs/Metadata/Symbols.hpp
@@ -26,7 +26,7 @@ namespace mrdocs {
 /** A unique identifier for a symbol.
 
     This is calculated as the SHA1 digest of the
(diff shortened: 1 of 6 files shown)


include/mrdocs/Metadata/Symbols.hpp
```diff
@@ -39,44 +39,74 @@ class SymbolID
...
+    /** Construct a SymbolID from a null-terminated string.
+
+        This function constructs a SymbolID from
+        a string. The string must be exactly 20
+        characters long.
+
+        @param src The string to construct from.
+    */
+    template <std::convertible_to<value_type> Char>
+    constexpr SymbolID(const Char* src)
```
sdkrystian, line 51: This isn't supposed to be public & will be removed eventually

PR #518, opened 2023-12-15
support @throw/@throws/@exception

sdkrystian:


alandefreitas:
Since we're here, it would be good to support @return already. It's a simple one.

sdkrystian:
@alandefreitas I thought we do?

alandefreitas:
> I thought we do?

I put something there but it's just a placeholder using the return type. It doesn't get anything from the javadoc.

alandefreitas:
> It doesn't get anything from the javadoc.

At least last time I tried.

PR #532, opened 2024-01-03
Execute cmake

fpelliccioni:


sdkrystian:
If we are invoking cmake, wouldn't it be better to simply obtain the implicit include directories from it instead? 

Also, it would be better if we searched for cmake in path rather than requiring the user to provide it explicitly.

Lastly, we might want to invoke with '-DBUILD_TESTING=OFF'.

alandefreitas:
> If we are invoking cmake, wouldn't it be better to simply obtain the implicit include directories from it instead?

Not really, because the process for getting the implicit include directories is a separate step in the workflow. We used to run cmake on an empty project and output these directories to a file from cmake. It's a workaround we don't need anymore since mrdocs can already run the compiler directly to find out what these paths are. The workaround would be a long way to achieve what the other PR by Fernando achieved.

> Also, it would be better if we searched for cmake in path rather than requiring the user to provide it explicitly.

Very true. Even providing some environment variable to fall back to if possible. Or even also try to fall back to some common locations, like `C:\Program Files\CMake`.

> Lastly, we might want to invoke with '-DBUILD_TESTING=OFF'.

We probably need some option in mrdocs.yml to describe the default cmake configure args. 

For instance, Boost.URL needs `-DBUILD_TESTING=ON` at least for now because if tests aren't compiled, header-only functions (like `boost::urls::format`) aren't compiled, and mrdocs won't document them. 

That's also a problem we have we filters right now. If `-DBUILD_TESTING=OFF`, some functions are not documented because they're not compiled. If `-DBUILD_TESTING=OFF` too many functions are documented because they're not filtered.

src/lib/Lib/MrDocsCompilationDatabase.cpp
```diff
@@ -165,6 +167,40 @@ adjustCommandLine(
     return new_cmdline;
 }
 
+std::optional<std::string> 
+executeCmakeExportCompileCommands(llvm::StringRef cmakePath, llvm::StringRef cmakeListsPath) 
+{
+    if ( ! llvm::sys::fs::exists(cmakePath)) {
```
alandefreitas, line 173: Return expected<std::string, Error> and describe the problem instead of `std::nullopt`?

src/lib/Lib/MrDocsCompilationDatabase.cpp
```diff
@@ -165,6 +167,40 @@ adjustCommandLine(
...
+        return std::nullopt;
+    }    
+    if ( ! llvm::sys::fs::exists(cmakeListsPath)) {
+        return std::nullopt;
+    }
+
+    llvm::SmallString<128> stdOutPath;
```
alandefreitas, line 180: These outputs are not being used?

src/lib/Lib/MrDocsCompilationDatabase.cpp
```diff
@@ -165,6 +167,40 @@ adjustCommandLine(
...
+    }
+
+    llvm::SmallString<128> stdErrPath;    
+    if (auto ec = llvm::sys::fs::createTemporaryFile("stderr", "txt", stdErrPath)) 
+    {
+        return std::nullopt;
+    }
+
+    std::optional<llvm::StringRef> const redirects[] = {llvm::StringRef(), stdOutPath.str(), stdErrPath.str()};
+    std::vector<llvm::StringRef> const args = {cmakePath, cmakeListsPath, "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"};
```
alandefreitas, line 193: I thought `cmakeListsPath` was the `CMakeLists.txt` path. It's actually its parent. This could be `projectPath` or something then.

alandefreitas: Projects usually have their requirements for CMake configure args. Only `DCMAKE_EXPORT_COMPILE_COMMANDS=ON` rarely works as configure args. These need to come from mrdocs.yml.

src/lib/Lib/MrDocsCompilationDatabase.cpp
```diff
@@ -165,6 +167,40 @@ adjustCommandLine(
...
+    int const result = llvm::sys::ExecuteAndWait(cmakePath, args, std::nullopt, redirects);
+
+    if (result != 0) 
+    {
+        return std::nullopt;
+    }
+
+    return "./compile_commands.json";
```
alandefreitas, line 201: What's the point of always returning the same string?

src/lib/Lib/MrDocsCompilationDatabase.hpp
```diff
@@ -19,6 +19,12 @@
 namespace clang {
 namespace mrdocs {
 
+/**
+ * Execute cmake to export compile_commands.json.
+*/
+std::optional<std::string> 
+executeCmakeExportCompileCommands(llvm::StringRef cmakePath, llvm::StringRef cmakeListsPath);
```
alandefreitas, line 26: This functionality seems unrelated. `MrDocsCompilationDatabase` requires an existing database it's going to convert. There can be no existing database if there's no `compile_commands.json` yet. So `MrDocsCompilationDatabase` can't be using this function.

src/tool/GenerateAction.cpp
```diff
@@ -74,12 +74,38 @@ DoGenerateAction()
     // Load the compilation database file
     //
     // --------------------------------------------------------------
-    MRDOCS_CHECK(toolArgs.inputPaths, "The compilation database path argument is missing");
-    MRDOCS_CHECK(toolArgs.inputPaths.size() == 1,
-        formatError(
-            "got {} input paths where 1 was expected",
-            toolArgs.inputPaths.size()));
-    auto compilationsPath = files::normalizePath(toolArgs.inputPaths.front());
+    if (detail::failed(toolArgs.inputPaths) && detail::failed(toolArgs.generateCompilationDatabase))
```
alandefreitas, line 77: `detail::failed` is not meant to be used like that. It's an abstraction for `MRDOCS_CHECK/TRY` so they can work with any type. You just want to use `if (x.empty())` here.

src/lib/Lib/MrDocsCompilationDatabase.cpp
alandefreitas: I'm not sure how cmake can run without the source and build paths as arguments.

And all the intermediary cmake build directories need to be in a temporary directory, where the compile_commands.json will also end up.

alandefreitas: If the default generator is Visual Studio, this needs to change to Ninja because Visual Studio can't generate compile_commands.json.

src/tool/GenerateAction.cpp
```diff
@@ -74,12 +74,38 @@ DoGenerateAction()
...
+    {
+        report::error("The compilation database path argument is missing or you can use --generate-compilation-database");
```
alandefreitas, line 79: I find this confusing. Even if `--generate-compilation-database`, mrdocs still needs to know the input path (which will now be the project path instead of compile_commands.json path).

I have the impression this is basically unrolling the `MRDOCS_CHECK` macros here (and forgetting to unroll one of them). This could be something like:

```
    MRDOCS_CHECK(!toolArgs.inputPaths.empty() || toolArgs.generateCompilationDatabase, "The input path argument is missing"); // <- not compilation database anymore
    MRDOCS_CHECK(toolArgs.inputPaths.size() == 1,
        formatError(
            "got {} input paths where 1 was expected",
            toolArgs.inputPaths.size()));
```

src/tool/ToolArgs.cpp
```diff
@@ -84,6 +84,20 @@ R"(
     "inputs",
     llvm::cl::Sink,
     llvm::cl::desc("The path to the compilation database."))
+
+, generateCompilationDatabase(
```
alandefreitas, line 88: Do we need that option? If the input path is a project directory instead of a compile_commands.json path, we can just assume we need to generate the compilation database. The cases are mutually exclusive:

1) `<dir>/compile_commands.json`: old behavior
2) `<dir>/` (a directory): run cmake with dir as source and (1)
3) `<dir>/CMakeLists.txt`: get parent <dir> and (2)
4) no input (?): get current working directory as <dir> and (2)

src/tool/GenerateAction.cpp
```diff
@@ -74,12 +74,38 @@ DoGenerateAction()
...
+        return {};
+    }
+
+    std::string inputPath;
+    if (toolArgs.generateCompilationDatabase)
+    {
+        MRDOCS_CHECK(toolArgs.cmakePath, "The cmake path argument is missing");
```
alandefreitas, line 86: If not provided, we need to fall back to some obvious defaults, such as environment variables and PATH.

src/tool/GenerateAction.cpp
```diff
@@ -74,12 +74,38 @@ DoGenerateAction()
...
+        MRDOCS_CHECK(toolArgs.cmakeListsPath, "The cmake-lists path argument is missing");
```
alandefreitas, line 87: This is just the input path.

src/tool/ToolArgs.cpp
```diff
@@ -84,6 +84,20 @@ R"(
...
+    "generate-compilation-database",
+    llvm::cl::desc("Generate the compilation database."),
+    llvm::cl::init(false))
+
+, cmakePath(
+    "cmake",
+    llvm::cl::desc(R"(The path to the cmake executable.)"))
+
+, cmakeListsPath(
```
alandefreitas, line 97: We just need the input path. There's no use for inputPath and cmakeListsPath at the same time because they're mutually exclusive.

src/tool/ToolArgs.cpp
fpelliccioni: I've implemented the requested refactoring for automatic compilation database generation, covering scenarios 1-3 as discussed.

Regarding scenario 4 (no input provided): I've held back on using the current working directory for output generation. I believe we should avoid cluttering the project directory with temporary files. Your input on this would be valuable.

Most of Alan's comments have been addressed, with a few remaining which I'm working on.

src/lib/Lib/MrDocsCompilationDatabase.cpp
fpelliccioni: You're right about the function currently always returning the same string. My intention was to encapsulate the responsibility of determining the path to "compile_commands.json" within this function, so the caller doesn't need to handle it. However, I agree that this implementation isn't entirely satisfactory.

Additionally, I'm still exploring whether it's feasible to generate "compile_commands.json" in a directory other than the working directory. I plan to address this soon and will add a TODO note to remind us of this pending improvement.

src/lib/Lib/MrDocsCompilationDatabase.cpp
fpelliccioni: Not yet, I am not sure if I will use them to return as error messages.

sdkrystian:
@fpelliccioni Please ensure that any files you create say  `Copyright (c) 2024 Fernando Pelliccioni [...]` instead of `Copyright (c) 2023 Vinnie Falco [...]` 😄. Also, you should not include `This is a derivative work. originally part of the LLVM Project.` unless the code came from LLVM.

src/lib/Lib/MrDocsCompilationDatabase.cpp
alandefreitas: It definitely *has* to be another directory. We can't mess with other people's source directories. It has to be in a build directory and a temporary build directory unless the user provides some directory considered safe.

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <llvm/Support/Path.h>
+#include <llvm/Support/Program.h>
+
+namespace clang {
+namespace mrdocs {
+
+Expected<std::string>
+getCmakePath() {
+    std::vector<std::string> paths = {
+        "/usr/bin/cmake",
```
alandefreitas, line 24: Except for "Program Files" and "opt" (to which we should fall back), all other directories should already be in PATH. 

If they're not, it's because the user made an explicit effort for them to not be in PATH so we shouldn't mess with that.

Also, it's (unfortunately) somewhat common to find tools in `/usr/opt` instead of only `/usr/local/opt`.

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,90 @@
...
+            return path;
+        }
+    }
+    return Unexpected(Error("cmake executable not found"));
+}
+
+Expected<std::string>
+executeCmakeExportCompileCommands(llvm::StringRef cmakeListsPath) 
+{
+    auto cmakePathRes = getCmakePath();
```
alandefreitas, line 53: I thought there would be a step in the calling code before `getCmakePath()` to get it from PATH. We need to look in PATH first. When the user puts it in PATH, the user is telling us it's where it wants his tools to be found. Even installing a CMake release version file that bundles it differently can impact where the tool will end up.

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,90 @@
...
+    if (! cmakePathRes) 
+    {
+        return cmakePathRes;
```
alandefreitas, line 56: This pattern is very verbose and repetitive. Use MRDOCS_TRY to eliminate that.

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,90 @@
...
+    }
+    auto cmakePath = *cmakePathRes;
+
+    if ( ! llvm::sys::fs::exists(cmakeListsPath)) 
+    {
+        return Unexpected(Error("CMakeLists.txt not found"));
+    }
+
+    llvm::SmallString<128> stdOutPath;
```
alandefreitas, line 65: Unused variable

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,90 @@
...
+    if (auto ec = llvm::sys::fs::createTemporaryFile("stdout", "txt", stdOutPath)) 
+    {
+        return Unexpected(Error("failed to create temporary file"));
+    }
+
+    llvm::SmallString<128> stdErrPath;
```
alandefreitas, line 71: Unused

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    llvm::SmallString<128> stdErrPath;    
+    if (auto ec = llvm::sys::fs::createTemporaryFile("stderr", "txt", stdErrPath)) 
+    {
+        return Unexpected(Error("failed to create temporary file"));
+    }
+
+    std::optional<llvm::StringRef> const redirects[] = {llvm::StringRef(), stdOutPath.str(), stdErrPath.str()};
+    std::vector<llvm::StringRef> const args = {cmakePath, cmakeListsPath, "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"};
```
alandefreitas, line 78: Missing args

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,90 @@
...
+    int const result = llvm::sys::ExecuteAndWait(cmakePath, args, std::nullopt, redirects);
+
+    if (result != 0) 
+    {
+        return Unexpected(Error("cmake execution failed"));
+    }
+
+    return "./compile_commands.json";
```
alandefreitas, line 86: Always same path. Should be the path to the file in the temporary build dir.

src/lib/Lib/CMakeExecution.cpp
alandefreitas: MSVC needs Ninja, etc...

just look at my old comments.

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,90 @@
...
+    auto cmakePathRes = getCmakePath();
+    if (! cmakePathRes) 
+    {
+        return cmakePathRes;
+    }
+    auto cmakePath = *cmakePathRes;
+
+    if ( ! llvm::sys::fs::exists(cmakeListsPath)) 
+    {
+        return Unexpected(Error("CMakeLists.txt not found"));
```
alandefreitas, line 62: MRDOCS_CHECK

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    llvm::SmallString<128> stdOutPath;
+    if (auto ec = llvm::sys::fs::createTemporaryFile("stdout", "txt", stdOutPath)) 
+    {
+        return Unexpected(Error("failed to create temporary file"));
```
alandefreitas, line 68: MRDOCS_CHECK

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    llvm::SmallString<128> stdErrPath;    
+    if (auto ec = llvm::sys::fs::createTemporaryFile("stderr", "txt", stdErrPath)) 
+    {
+        return Unexpected(Error("failed to create temporary file"));
```
alandefreitas, line 74: MRDOCS_CHECK

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    std::optional<llvm::StringRef> const redirects[] = {llvm::StringRef(), stdOutPath.str(), stdErrPath.str()};
+    std::vector<llvm::StringRef> const args = {cmakePath, cmakeListsPath, "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"};
+    int const result = llvm::sys::ExecuteAndWait(cmakePath, args, std::nullopt, redirects);
+
+    if (result != 0) 
+    {
+        return Unexpected(Error("cmake execution failed"));
```
alandefreitas, line 83: MRDOCS_CHECK

src/tool/GenerateAction.cpp
```diff
@@ -20,11 +21,43 @@
...
 namespace mrdocs {
 
+Expected<std::string>
+generateCompilationDatabaseIfNeeded(llvm::StringRef projectPath)
+{
+    namespace fs = llvm::sys::fs;
+    namespace path = llvm::sys::path;
+
+    fs::file_status fileStatus;
+    if (auto ec = fs::status(projectPath, fileStatus))
```
alandefreitas, line 37: MRDOCS_CHECK

src/tool/GenerateAction.cpp
```diff
@@ -20,11 +21,43 @@
...
+    {
+        return Unexpected(Error(ec));
+    }
+
+    if (fs::is_directory(fileStatus))
+    {
+        return executeCmakeExportCompileCommands(projectPath);
+    }
+    else if (fs::is_regular_file(fileStatus))
```
alandefreitas, line 46: Else not needed because first conditionals return.

Not sure if `fs::is_regular_file` is needed at this point. We already checked if it's a directory. We don't need to check if it's a symlink.

src/tool/GenerateAction.cpp
```diff
@@ -20,11 +21,43 @@
...
 #include <clang/Tooling/JSONCompilationDatabase.h>
 
+
 #include <cstdlib>
 
 namespace clang {
 namespace mrdocs {
 
+Expected<std::string>
+generateCompilationDatabaseIfNeeded(llvm::StringRef projectPath)
```
alandefreitas, line 31: Just `generateCompilationDatabase` should be fine. The combination of functions that already return `Expected` with `maybe_do_X`, `do_X_if_needed`, `try_to_do_X`, is dangerously viral. If it's not needed, we can just return no result.

Just include some javadocs around the function in the header files and we're fine.

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,90 @@
...
+                return Unexpected(Error("cmake execution failed"));
+            }
+            return path;
+        }
+    }
+    return Unexpected(Error("cmake executable not found"));
+}
+
+Expected<std::string>
+executeCmakeExportCompileCommands(llvm::StringRef cmakeListsPath)
```
alandefreitas, line 51: cmakeListsPath -> projectPath or sourcePath

src/lib/Lib/CMakeExecution.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <mrdocs/Support/Error.hpp>
+
+namespace clang {
+namespace mrdocs {
+
+/**
+ * Execute cmake to export compile_commands.json.
+*/
+Expected<std::string>
+executeCmakeExportCompileCommands(llvm::StringRef cmakeListsPath);
```
alandefreitas, line 26: cmakeListsPath -> projectPath or sourcePath

src/lib/Lib/CMakeExecution.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <string>
+
+#include <llvm/ADT/StringRef.h>
+#include <mrdocs/Support/Error.hpp>
+
+namespace clang {
+namespace mrdocs {
+
+/**
+ * Execute cmake to export compile_commands.json.
```
alandefreitas, line 23: Explain the expected input and output

alandefreitas (commented):
Nice work. The tool is going to be so much easier to use with this.

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -28,17 +29,70 @@ getCmakePath() {
...
+        }
+    }
+    return Unexpected(Error("Default CMake generator not found"));
+}
+
+Expected<bool>
+cmakeDefaultGeneratorIsVisualStudio(llvm::StringRef cmakePath) 
+{
+    MRDOCS_TRY(auto const defaultGenerator, getCmakeDefaultGenerator(cmakePath));
+    return defaultGenerator.find("Visual Studio") != std::string::npos;
```
alandefreitas, line 76: `.starts_with("Visual Studio")` or something equivalent?

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -28,17 +29,70 @@ getCmakePath() {
...
     MRDOCS_TRY(auto const cmakePath, getCmakePath());
     MRDOCS_CHECK(llvm::sys::fs::exists(projectPath), "CMakeLists.txt not found");
+    MRDOCS_TRY(auto const cmakeDefaultGeneratorIsVisualStudio, cmakeDefaultGeneratorIsVisualStudio(cmakePath));
 
     llvm::SmallString<128> tempDir;
     MRDOCS_CHECK(!llvm::sys::fs::createUniqueDirectory("compile_commands", tempDir), "Failed to create temporary directory");
 
     std::optional<llvm::StringRef> const redirects[] = {llvm::StringRef(), llvm::StringRef(), llvm::StringRef()};
-    std::vector<llvm::StringRef> const args = {cmakePath, "-S", projectPath, "-B", tempDir.str(), "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"};
+    std::vector<llvm::StringRef> args = {cmakePath, "-S", projectPath, "-B", tempDir.str(), "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"};
```
alandefreitas, line 90: -S and -B implies we require cmake >=3.13.5. 

When this feature is documented, this requirement needs to be included. 

I actually have implemented the logic to determine the cmake version and choose between the -S/-B and the older syntax in the cmake-workflows action but I don't think we need that here.

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -28,17 +29,70 @@ getCmakePath() {
...
+    if (cmakeDefaultGeneratorIsVisualStudio) {
+        args.push_back("-G");
```
alandefreitas, line 92: The user might have already foreseen the problem and already included `-G <some generator>` in their cmake-args. We probably need to look for `-G <something>` and (i) only change it if we need, and (ii) remove it if it's visual studio.

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include <llvm/Support/FileSystem.h>
+#include <llvm/Support/MemoryBuffer.h>
+#include <llvm/Support/Path.h>
+#include <llvm/Support/Program.h>
+
+namespace clang {
+namespace mrdocs {
+
+Expected<std::string>
+getCmakePath() {
```
sdkrystian, line 22: These helper functions should have internal linkage

include/mrdocs/Config.hpp
```diff
@@ -104,6 +104,10 @@ class MRDOCS_DECL
         */
         std::string extraYaml;
 
+        /** A string holding the complete CMake arguments.
+        */
+        std::string cmake;
```
alandefreitas, line 109: Everything should go into `ConfigImpl` unless external plugins need it (of we don't have evidence yet). 

But in that case you need neither `Config` or `ConfigImpl`. You can get this value directly from `Config::object`.

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,164 @@
...
+    llvm::SmallString<128> outputPath;
+    MRDOCS_CHECK(!llvm::sys::fs::createTemporaryFile("cmake-help", "txt", outputPath), 
+        "Failed to create temporary file");
+    std::optional<llvm::StringRef> const redirects[] = {llvm::StringRef(), outputPath.str(), llvm::StringRef()};
+    std::vector<llvm::StringRef> const args = {cmakePath, "--help"};
+    llvm::ArrayRef<llvm::StringRef> emptyEnv;
+    int const result = llvm::sys::ExecuteAndWait(cmakePath, args, emptyEnv, redirects);
+    MRDOCS_CHECK(result == 0, "CMake execution failed");
+
+    auto const bufferOrError = llvm::MemoryBuffer::getFile(outputPath);
```
alandefreitas, line 46: I don't know the type of `bufferOrError`, but if it's a common type this is a job for MRDOCS_TRY.

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,164 @@
...
+{
+    std::vector<std::string> parsedArgs;
+    std::string arg;
+    bool inQuotes = false;
+
+    for (char ch : cmakeArgsStr) 
+    {
+        if (ch == '"') 
+        {
+            inQuotes = !inQuotes;
```
alandefreitas, line 92: Is this failing to consider quotes can be escaped?

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,164 @@
...
+        } 
+        else if (std::isspace(ch) && !inQuotes) 
+        {
+            if ( ! arg.empty()) 
+            {
+                if (arg[0] == '-' && arg.size() == 2) 
+                    continue;
```
alandefreitas, line 99: Why put keys in the same parsedArg as values?

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,164 @@
...
+    return parsedArgs;
+}
+
+} // anonymous namespace
+
+Expected<std::string>
+executeCmakeExportCompileCommands(llvm::StringRef projectPath, llvm::StringRef cmakeArgs) 
+{
+    MRDOCS_TRY(auto const cmakePath, getCmakePath());
+    MRDOCS_CHECK(llvm::sys::fs::exists(projectPath), "CMakeLists.txt not found");
```
alandefreitas, line 124: This check should come before the call to getCmakePath

alandefreitas: The error message could be only that projectPath doesn't exist. It didn't check for CMakeLists.txt.

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,164 @@
...
+    MRDOCS_CHECK(!llvm::sys::fs::createUniqueDirectory("compile_commands", tempDir), "Failed to create temporary directory");
+
+    std::optional<llvm::StringRef> const redirects[] = {llvm::StringRef(), llvm::StringRef(), llvm::StringRef()};
+    std::vector<llvm::StringRef> args = {cmakePath, "-S", projectPath, "-B", tempDir.str(), "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"};
+
+    auto const additionalArgs = parseCmakeArgs(cmakeArgs.str());
+    bool visualStudioFound = false;
+    for (auto const& arg : additionalArgs) 
+    {
+        if (arg.starts_with("-G") && arg.find("Visual Studio", 2) != std::string::npos)
```
alandefreitas, line 136: Oh... so you put keys and values in the same arg to make it easier to find here. Not a good idea in terms of separation of concerns. Except for -D arguments, other arguments do need to be separated and the quotes can't be mixed when the API requotes them. The case where the key is -G can always be easily handled separately (if -G then check if continuation or next arg starts with Visual Studio).

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    llvm::SmallString<128> tempDir;
+    MRDOCS_CHECK(!llvm::sys::fs::createUniqueDirectory("compile_commands", tempDir), "Failed to create temporary directory");
+
+    std::optional<llvm::StringRef> const redirects[] = {llvm::StringRef(), llvm::StringRef(), llvm::StringRef()};
+    std::vector<llvm::StringRef> args = {cmakePath, "-S", projectPath, "-B", tempDir.str(), "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"};
+
+    auto const additionalArgs = parseCmakeArgs(cmakeArgs.str());
+    bool visualStudioFound = false;
+    for (auto const& arg : additionalArgs)
```
alandefreitas, line 134: If the user provides -DCMAKE_EXPORT_COMPILE_COMMANDS=ON we also need to remove that to avoid including it twice.

src/lib/Lib/ConfigImpl.cpp
```diff
@@ -105,6 +105,8 @@ struct llvm::yaml::MappingTraits<SettingsImpl>
         io.mapOptional("input",             cfg.input);
 
         io.mapOptional("filters",           cfg.filters);
+
+        io.mapOptional("cmake",             cfg.cmake);
```
alandefreitas, line 109: Not needed

src/lib/Lib/MrDocsCompilationDatabase.cpp
```diff
@@ -22,6 +22,8 @@
 #include <llvm/Option/OptSpecifier.h>
 #include <llvm/Option/OptTable.h>
 #include <llvm/Support/FileSystem.h>
+#include <llvm/Support/Program.h>
```
alandefreitas, line 25: Unused?

src/lib/Lib/CMakeExecution.cpp
alandefreitas: The executeCmakeExportCompileCommands javadoc says projectPath can be the source path or CMakeLists.txt. This function doesn't handle both cases. It handles only the source path case (without checking if CMakeLists.txt exists).

src/tool/GenerateAction.cpp
```diff
@@ -74,12 +120,21 @@ DoGenerateAction()
...
     // --------------------------------------------------------------
+    
     MRDOCS_CHECK(toolArgs.inputPaths, "The compilation database path argument is missing");
     MRDOCS_CHECK(toolArgs.inputPaths.size() == 1,
         formatError(
             "got {} input paths where 1 was expected",
             toolArgs.inputPaths.size()));
-    auto compilationsPath = files::normalizePath(toolArgs.inputPaths.front());
+
+    auto const inputPath = generateCompilationDatabase(toolArgs.inputPaths.front(), config->settings().cmake);
```
alandefreitas, line 130: This function name is ambiguous. It could be just generateCompileCommandsFile or something like that. CompilationDatabase here usually refers to an LLVM object type that can be generated in many ways (including reading a compile_commands.json file).

alandefreitas: Instead of `config->settings().cmake`, you can just read that property from the object if it exists or provide an empty string otherwise.

alandefreitas (commented):
CI needs to be updated to use the new feature.

This new behavior needs to be documented.

src/lib/Lib/CMakeExecution.cpp
fpelliccioni: I can't use `MRDOCS_TRY` because apparently it uses `std::move` and the type returned by `llvm::sys::ExecuteAndWait()` is not movable.

fpelliccioni: I will mark this comment as resolved.

src/lib/Lib/CMakeExecution.cpp
fpelliccioni: I'm combining the arguments (like `-D KEY=VALUE` or `-G VALUE`) because when passed as individual elements in the vector to `llvm::sys::ExecuteAndWait()`, each argument is sent to CMake in a way similar to what you described over Slack. As per your suggestion of formatting arguments like `"-DA=B C"`, keeping them combined ensures that they're processed correctly by CMake. While it might not be optimal in terms of software design purity, practically, it works well. I've tested various ways of specifying quotes (realized I missed escaping them, thanks) and this approach seems to function effectively. This piece of code is not intended to be a comprehensive CMake argument parser but rather to address our specific use case. 
What do you think?

fpelliccioni: Additionally, I'm aware that if the code starts to grow and we need to add more special conditions for argument parsing, it might become unwieldy or less clean. In that scenario, I agree that a refactoring of the parser would be necessary to maintain clarity and manageability. But as of now, it seems to serve our practical purposes well without overcomplicating things.

src/lib/Lib/CMakeExecution.cpp
fpelliccioni: Answered here: https://github.com/cppalliance/mrdocs/pull/532#discussion_r1447108981

src/lib/Lib/MrDocsCompilationDatabase.cpp
```diff
@@ -22,6 +22,7 @@
 #include <llvm/Option/OptSpecifier.h>
 #include <llvm/Option/OptTable.h>
 #include <llvm/Support/FileSystem.h>
+
```
fpelliccioni, line 25:
```suggestion
```

src/lib/Lib/CMakeExecution.cpp
alandefreitas: Oh... I see.

> As per your suggestion of formatting arguments like "-DA=B C", keeping them combined ensures that they're processed correctly by CMake

I think I miscommunicated that then. The comment was about args with literal `\"`s in it. Not about all combinations of key args (starting with `-`) and value args. 

For args in general, you just push them to the list of args as usual. There's no need to merge `-D` with the following arg. The logic you implemented is not only for `-D` but for all keys, and this merged arg will not always be valid. So you can just implement the simpler logic here.

So, about the point I was making, if you choose to implement it, is all about quotations. The problem is bash recognizes quotations around args (for instance, `"A C"` to represent a single arg) while cmake recognizes quotations inside args (for instance, `D="A C"` to represent a single pair). Note that make understands `D="A C"` as a single arg representing a pair while bash doesn't understand that. Bash understands `D="A C"` to be 2 args: `D="A` and `C"`. CMake uses two passes to convert the arguments bash understands to the arguments it understands. 

The problem this causes is that if your API to call processes attempt to quote everything again, when sending `D="A C"` to CMake, it will send something like `"D=\"A"` and `"C\""`, and CMake will not be able to reconstruct `D="A C"` from that. First you'd have to check if this is really a problem because the LLVM API for calling the process might not have this problem. Second, if this is a problem, you might not choose to fix it for now. Third, if you choose to fix the problem, I don't know what the best solution for this is but what I implemented in cpp-actions was logic to convert all args like `D="A C"` to `"D=A C"`, which is recognizable both by bash and cmake. [This](https://github.com/alandefreitas/cpp-actions/blob/89228895355dc399f2e1ef9d310c5f9695e30183/cmake-workflow/index.js#L1090) is the specific regex I used but, as I said, I'm not even sure that's the best solution.

> This piece of code is not intended to be a comprehensive CMake argument parser but rather to address our specific use case.

To be clear, the problem with that logic is not that it's incomplete. It's that it includes more complex logic that was effectively making things worse. Namely, your algorithm is joining any arg with the preceding arg if the preceding arg whenever it starts with `-`. At best, this makes things unnecessarily more complex. At worst, it breaks things because this is not valid for all args.

This is the main issue. The problem with arguments that contain quotes is an independent problem.

alandefreitas (commented):
Looking good. ci.yml needs to be updated to receive the project path only in the tests and adoc files need to be updated to document this behavior.

fpelliccioni:
> Looking good. ci.yml needs to be updated to receive the project path only in the tests and adoc files need to be updated to document this behavior.

CI and docs updated

PR #534, opened 2024-01-19
fix: LookupCallback has virtual destructor

alandefreitas:


sdkrystian:
I presume this is to fix a warning? Because a virtual destructor isn't needed here...

alandefreitas:
> I presume this is to fix a warning? 

Exactly.

PR #539, opened 2024-01-25
refactor: fallback to cmake --system-information

fpelliccioni:


src/lib/Lib/CMakeExecution.cpp
```diff
@@ -104,8 +131,54 @@ getCmakeDefaultGenerator(llvm::StringRef cmakePath)
...
+        }
+    }
+    return Unexpected(Error("Default CMake generator not found"));
+}
+
+Expected<std::string>
+getCmakeDefaultGenerator(llvm::StringRef cmakePath)
+{
+    Expected<std::string> const cmakeHelpExp = executeCmakeHelp(cmakePath);
+    if (!cmakeHelpExp)
```
alandefreitas, line 161: In this case, it's easier to define the case where `cmakeHelpExp` is true first.

```cpp
if (cmakeHelpExp)
{
     return parseCmakeHelpOutput(*cmakeHelpExp);
}
// fall back to system information
}

File: src/lib/Support/Path.cpp [hunk 2]
```diff
@@ -419,7 +419,7 @@ ScopedTempFile(
     ok_ = !llvm::sys::fs::createTemporaryFile(prefix, ext, tempPath);
     if (ok_)
     {
-        path_ = files::makeDirsy(path_.str());
+        path_ = tempPath;
```
alandefreitas, line 422: Oh...

.github/workflows/ci.yml
```diff
@@ -95,7 +95,7 @@ jobs:
           config_type="Release"
           filename="${{ runner.os }}-$config_type-29b20829.${{ ( runner.os == 'Windows' && '7z' ) || 'tar.xz' }}"
           url="https://mrdox.com/llvm+clang/$filename"
-          
+
```
alandefreitas, line 98: I think you can roll back the changes in this file.

src/lib/Lib/CMakeExecution.cpp
```diff
@@ -104,8 +131,55 @@ getCmakeDefaultGenerator(llvm::StringRef cmakePath)
...
+}
+
+Expected<std::string>
+getCmakeDefaultGenerator(llvm::StringRef cmakePath)
+{
+    Expected<std::string> const cmakeHelpExp = executeCmakeHelp(cmakePath);
+    if (cmakeHelpExp)
+    {
+        std::string const cmakeHelp = *std::move(cmakeHelpExp);
+        return parseCmakeHelpOutput(cmakeHelp);
```
alandefreitas, line 164: The code looks cleaner this way but now I'm wondering: executeCmakeHelp doesn't fail but parseCmakeHelpOutput does, shouldn't we move on and fall back to executeCmakeSystemInformation? And of course the same logic would apply to executeCmakeSystemInformation.

alandefreitas:
```cpp
    Expected<std::string> const cmakeHelpExp = executeCmakeHelp(cmakePath);
    if (cmakeHelpExp)
    {
        std::string const cmakeHelp = *std::move(cmakeHelpExp);
        auto r = parseCmakeHelpOutput(cmakeHelp);
        if (r) 
        {
            return *r;
        }
    }
    // ... let it fall back to the next case...
```

alandefreitas (commented):
This looks great.

.github/workflows/ci.yml
fpelliccioni, line 98:
```suggestion
          
```

.github/workflows/ci.yml
```diff
@@ -107,7 +107,7 @@ jobs:
               echo "Neither curl nor wget are available"
               exit 1
           fi
-          
+
```
fpelliccioni, line 110:
```suggestion
          
```

.github/workflows/ci.yml
```diff
@@ -302,6 +302,8 @@ jobs:
 
       - name: Generate demos
         run: |
+          set -x
+          cmake --help
```
fpelliccioni, line 306:
```suggestion
```

src/lib/Lib/CMakeExecution.cpp
fpelliccioni: Good suggestion

PR #543, opened 2024-02-14
refactor: format ASTVisitor

alandefreitas:
This PR simplifies some duplicated code, includes some javadoc, and "vinnie"-formats the ASTVisitor.


src/lib/AST/ASTVisitor.cpp
```diff
@@ -317,10 +318,10 @@ class ASTVisitor
             for(; NI != NE; ++NI, ++DI)
             {
                 // reached the end of the directory path
-                if(DI == DE)
+                if (DI == DE)
```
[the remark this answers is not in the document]
alandefreitas: From other projects, I thought you preferred it this way. It was probably just a coincidence then.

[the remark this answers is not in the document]
sdkrystian: Agree with Vinnie; please don't change whitespace on existing code (unless it's you're removing trailing whitespace or replacing tabs with spaces).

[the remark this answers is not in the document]
alandefreitas: Yes. I'll undo it. 

We have whitespace there in other projects so I thought it was intentional / something Vinnie cared about.

src/lib/AST/ASTVisitor.cpp
```diff
@@ -74,7 +74,9 @@ struct SymbolFilter
         current = node;
         detached = node_detached;
         if(node && node->Explicit)
+        {
```
sdkrystian, line 77: Please do not add braces to selection/iteration statements

src/lib/AST/ASTVisitor.cpp
```diff
@@ -760,40 +889,45 @@ class ASTVisitor
...
         was used as the pattern for instantiation.
     */
-    template<typename DeclTy>
-    DeclTy* getInstantiatedFrom(DeclTy* D);
+    template <class DeclTy>
+    DeclTy*
+    getInstantiatedFrom(DeclTy* D);
 
-    template<typename DeclTy>
+    template <class DeclTy>
```
sdkrystian, line 896: Please leave template-heads in their current style (no space between template and '<', and using `typename`)

vinniefalco: I originally used `class` but someone started using `typename`. Not clear why the longer version is preferred. Also Alan are you using clang-format? Because it sure looks like it.

alandefreitas: > I originally used class but someone started using typename. 

Yes. You told me to always use `class` because it was shorter.

> Not clear why the longer version is preferred. 

I don't think it was an explicit decision. I think it's just something Krystian likes. So I reverted it.

> Also Alan are you using clang-format? Because it sure looks like it.

Not at all. If I used clang-format at this point this commit would have 4000 changes.

PR #546, opened 2024-03-18
ci: statically link libc

alandefreitas:
Move configuration values to the matrix action and statically links both `libgcc` and `libstdc++`.

Since we are using the latest version of ubuntu in our workflow, this is intended to allow the release binaries to work on any earlier version of ubuntu. This is required by the Antora scripts in the Boost release tools.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,9 @@ jobs:
 
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.runs-on }}
-    container: ${{ (matrix.compiler == 'gcc' && 'ubuntu:22.04') || matrix.container }}
+    # Typically ubuntu-23.04 for GCC, which is needed later
+    # to test if the release binaries work on ubuntu-20.04.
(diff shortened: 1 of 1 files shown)


alandefreitas:
@sdkrystian I'm getting

```
    Target "mrdocs" links to target "Terminfo::terminfo" but the target was not
    found.  Perhaps a find_package() call is missing for an IMPORTED target, or
    an ALIAS target is missing?
```

in CI. Do we need to recompile LLVM to fix that?

alandefreitas:
Yes. I can't get this to work the simple way. If I use `-static` I get

```
/usr/bin/ld: attempted static link of dynamic object `/usr/lib/x86_64-linux-gnu/libtinfo.so'
```

On the other hand, `-static-libgcc` compiles in ubuntu 23.04 but it's not enough to make it work on ubuntu 20.04.

```
+ mrdocs --version
mrdocs: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.36' not found (required by mrdocs)
```

Issue #547, opened 2024-03-20
Statically linked executable


alandefreitas:
Because of how time-consuming it is to build mrdocs, the way mrdocs is supposed to be used is with scripts that download portable executables and run the tool. These scripts will usually be run in CI on some ubuntu-X container. So we need to provide portable executables people can use. 

The problem is we're using recent versions of ubuntu (ubuntu 23.04 now) because of the C++ features we need but users (including the boost release scripts) are usually using earlier versions of ubuntu (usually from 18.04 to 22.04) for a variety of reasons. And binaries built on ubuntu X (say 23.04) don't work under ubuntu <X (say 20.04).

I read about the solutions people proposed in [#boost](https://cpplang.slack.com/archives/C27KZLB0X) and wrote this summary of alternatives to the problem. I'm initially tempted to experiment with musl and a system libs wrapper.

Most solutions above would require us to rebuild all dependencies with `-static`/`musl`/`cosmopolitan`/`-ffreestanding` and might end up depending on other tools. Each solution is also quite complex and with lots of caveats. There's no way to know if it's going to work before testing.

1) Some people reported `-static`/`-static-libgcc` worked sometimes
    - It's also the first thing I tried but I couldn't get it to work (more details in the PR)
        - Compiling with `-static` gives us an error when trying to link `nlibtinfo.so`
        - `-static-libgcc` still asks for `GLIBC_2.36` in ubuntu 20.04 when looking for `libc.so.6`
    - Another example is https://www.github.com/muttleyxd/clang-tools-static-binaries
        - It uses `-static-libgcc` and builds LLVM with `-DDBUILD_SHARED_LIBS=OFF -DLLVM_BUILD_STATIC=ON`
        - I'm not sure `-static-libgcc` really works there or if it just works because they're compiling the tools with ubuntu-20.04 which should have no conflicts with later versions of ubuntu anyway
2) Use [musl](https://wiki.musl-libc.org/)
    - It's a more [static-friendly](https://superuser.com/a/1820299) C standard library implementation
    - Smaller than glibc
    - It's what the hdoc guys use
    - Documentation is not great
    - More references:
        - https://github.com/georgy7/build-with-musl
        - https://blog.gibson.sh/2017/11/26/creating-portable-linux-binaries/#some-general-suggestions
3) Use [cosmopolitan](https://github.com/jart/cosmopolitan)
    - Also a [libc](https://justine.lol/cosmopolitan/index.html) worth trying
[...]

PR #562, opened 2024-04-05
ci: create release for Ubuntu 20.04

alandefreitas:
This commit adds a new workflow that creates a Linux release on Ubuntu 20.04. This allows the release package to be used on more versions of Ubuntu. Since the C++ compilers we need are not available in Ubuntu 20.04, we use the clang executable from our own LLVM built from source to build the dependencies and MrDocs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,8 +33,16 @@ jobs:
             clang 17
             msvc 14.34
           standards: '20'
-          latest-factors: ''
+          # Set up a factor to build the main package on Ubuntu20.
+          # We use this factor with clang because it can be more easily installed
+          # from source on Ubuntu20.
+          # When this factor is set, we'll build the main package with our own clang-bin
+          # and libc++, which is then packaged for the release.
+          # This creates a version of MrDocs that works on Ubuntu >=20.04.
+          latest-factors: |
+            clang Ubuntu20-Release
           factors: ''
+
 
   build:
     needs: cpp-matrix
@@ -50,7 +58,7 @@ jobs:
 
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.runs-on }}
-    container: ${{ (matrix.compiler == 'gcc' && 'ubuntu:22.04') || matrix.container }}
+    container: ${{ (matrix.ubuntu20-release && 'ubuntu:20.04') || matrix.container }}
     env: ${{ matrix.env }}
     permissions:
       contents: write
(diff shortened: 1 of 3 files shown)


alandefreitas:
While working on many things related to Antora + Mrdocs, I've been exploring this solution in parallel for about 10 days now. As each attempt might take a couple of hours, I've been tweaking CI here and there and seeing how it goes in parallel. 

It's a good time to concede defeat. It's safe to say this is a bad idea to solve #547 because it's very unstable:

- It's overly complex / try-hard: To build MrDocs on Ubuntu 20.04 we need a custom-built compiler built with the tools Ubuntu 20.04 provides (clang + libcxx is the only option here) and then build LLVM again with this new custom toolchain. This involves many fine details that are hard to get right. It's also hard to maintain because we may need to adapt that workflow whenever MrDocs needs more new C++ features.
- It's slow: even with caching, building this custom toolchain and then building LLVM with it is much slower than the usual CI workflow. And every once in a while, the cache will need to be rebuilt, and building the toolchain twice might take over 2 hours.
- It's not general enough: This strategy allows us to support Ubuntu >=20.04 instead of Ubuntu >=22.04 or Ubuntu >=23.04. It doesn't generalize to other platforms, like the strategy CMake uses (APBuild, which is equivalent to a wrapper).
- It doesn't work properly. I've tried dozens of ways to make it work. If we [don't statically link libc++](https://github.com/cppalliance/mrdocs/actions/runs/8657616753/job/23747448921) then we have the same problem generating the demos because we now have a compatible libc version but not a compatible libc++ on the machine using the release. But [if we attempt to statically link libc++](https://github.com/cppalliance/mrdocs/actions/runs/8655543722/job/23734652584) we get other problems because clanglib doesn't support that properly.

Considering these limitations, the next idea I'll explore to solve #547 is [a wrapper overriding system libs](https://blog.gibson.sh/2017/11/26/creating-portable-linux-binaries/#a-wrapper-that-selectively-overrides-system-libs). This strategy is somewhat similar to what APBuild does but it's manually and more stable. It's similar to how CMake provides its binaries. This wrapper also seemed like a complex solution at first but it turns out to be much simpler and more effective than what we're trying to do here. I'll describe the proposed solution in #547.

alandefreitas:
While working on many things related to Antora + Mrdocs, I've been exploring this solution in parallel for about 10 days now. As each attempt might take a couple of hours, I've been tweaking CI here and there and seeing how it goes in parallel. 

It's a good time to concede defeat. It's safe to say this is a bad idea to solve #547 because it's very unstable:

- It's overly complex / try-hard: To build MrDocs on Ubuntu 20.04 we need a custom-built compiler built with the tools Ubuntu 20.04 provides (clang + libcxx is the only option here) and then build LLVM again with this new custom toolchain. This involves many fine details that are hard to get right. It's also hard to maintain because we may need to adapt that workflow whenever MrDocs needs more new C++ features.
- It's slow: even with caching, building this custom toolchain and then building LLVM with it is much slower than the usual CI workflow. And every once in a while, the cache will need to be rebuilt, and building the toolchain twice might take over 2 hours.
- It's not general enough: This strategy allows us to support Ubuntu >=20.04 instead of Ubuntu >=22.04 or Ubuntu >=23.04. It doesn't generalize to other platforms, like the strategy CMake uses (APBuild, which is equivalent to a wrapper).
- It doesn't work properly. I've tried dozens of ways to make it work. If we [don't statically link libc++](https://github.com/cppalliance/mrdocs/actions/runs/8657616753/job/23747448921) then we have the same problem generating the demos because we now have a compatible libc version but not a compatible libc++ on the machine using the release. But [if we attempt to statically link libc++](https://github.com/cppalliance/mrdocs/actions/runs/8655543722/job/23734652584) we get other problems because clanglib doesn't support that properly.

Considering these limitations, the next idea I'll explore to solve #547 is [a wrapper overriding system libs](https://blog.gibson.sh/2017/11/26/creating-portable-linux-binaries/#a-wrapper-that-selectively-overrides-system-libs). This strategy is somewhat similar to what APBuild does but it's manually and more stable. It's similar to how CMake provides its binaries. This wrapper also seemed like a complex solution at first but it turns out to be much simpler and more effective than what we're trying to do here. I'll describe the proposed solution in #547.

Issue #529, opened 2024-01-03
Command line option --output takes paths not relative to cwd


anarthal:
When running
```
mrdocs ... --output=dir
```

`dir` is interpreted relative to the config file, and not to `cwd`. This is a confusing setting, since no other tool does that.

alandefreitas:
Doesn't doxygen do that?

fpelliccioni:
> Command line option --output now takes paths relative to cwd #571

 I've tested doxygen and couldn't find a way to pass the output directory as an argument to the executable. What doxygen does support is specifying the `OUTPUT_DIRECTORY` in the config file. I can confirm that the output directory in doxygen is relative to the directory where doxygen is run, not the directory where the config file is located.

alandefreitas:
IIRC, the Doxygen command is `-s OUTPUT_DIRECTORY=...`, and it overrides whatever comes from the `Doxyfile.` In this case, any path specified in the command line simply overrides the value in the Doxyfile and should also be relative to it. 

Paths are relative to mrdocs.yml because mrdocs.yml is at a stable location in the project. Requiring all relative paths to be relative to cwd doesn't seem like a good idea because it would be extremely verbose, the relative values in mrdocs.yml wouldn't match the command line arguments which would keep changing according to cwd, and creating an exception for `output` would make it even more confusing. 

We can just close this issue, and this behavior should be better documented.

alandefreitas:
Obsoleted by #617

PR #571, opened 2024-05-15
Command line option --output now takes paths relative to cwd

fpelliccioni:
Fixes https://github.com/cppalliance/mrdocs/issues/529

### DIFF
--- a/src/tool/GenerateAction.cpp
+++ b/src/tool/GenerateAction.cpp
@@ -24,6 +24,7 @@
 
 
 #include <cstdlib>
+#include <filesystem>
 
 namespace clang {
 namespace mrdocs {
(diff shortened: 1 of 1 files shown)


alandefreitas:
The bigger question here is whether Doxygen does that. IIRC, doxygen renders the paths relative to the Doxyfile. We have to ensure what their behavior is. 

- Does doxygen allow these values to be passed on the command line? 
- Is our policy in the command line relative to the current working directory, and any setting in mrdocs.yml relative to mrdocs.yml?
- Are there other values we need to adjust? Many other paths can be passed in the command line.

fpelliccioni:
> * Does doxygen allow these values to be passed on the command line?

I tested doxygen, and it doesn't seem to support passing the output directory as an argument to the executable. Doxygen allows specifying the OUTPUT_DIRECTORY in the config file, and this directory is relative to where doxygen is run, not where the config file is located.

> * Is our policy in the command line relative to the current working directory, and any setting in mrdocs.yml relative to mrdocs.yml?

I'm not entirely sure I understand this question.

> * Are there other values we need to adjust? Many other paths can be passed in the command line.

I'll check this and report back on whether there are other values that need adjustment to ensure consistency.

PR #602, opened 2024-05-29
Create overload page for specializations

sdkrystian:


alandefreitas:
I don't know if it's just about the terminology or if it affects the output, but there are no specialization overloads in C++. There's the main template and then the specializations. The main page is dedicated to the main template, with the content from its doc comments, and it contains links to other pages for its specializations. Unfortunately, there's no example of that with Doxygen+Docca because that's a feature Doxygen doesn't support. It's another C++ construct we emulate that with macros. However, for instance, see std::vector and std::formatter in cppreference.

alandefreitas:
This is from May 29. Can we close it?

PR #605, opened 2024-05-30
configurable see-below and implementation-defined namespaces

sdkrystian:


alandefreitas (commented):
I saw there's a FOLD commit in there. So if it's still possible, it'd be good to find something less ambiguous than "Special" in "SpecialNamespace". Maybe "DetailNamespace"? I don't know.

Issue #299, opened 2023-06-13
pre/post doc node types


vinniefalco:
Doxygen has `@pre` and `@post` commands for Precondition and Postconditions we can extract them into typed nodes. Also "effects."

PR #609, opened 2024-05-30
support @pre and @post

sdkrystian:


sdkrystian:
Fixes #299

include/mrdocs/Metadata/Javadoc.hpp
```diff
@@ -709,6 +711,50 @@ struct Throws : Paragraph
...
+            std::move(details_.children))
+    {
+    }
+
+    bool operator==(const Precondition&)
+        const noexcept = default;
+
+    bool equals(const Node& other) const noexcept override
+    {
+        return kind == other.kind &&
```
vinniefalco, line 731: should this be
```
return static_kind = other.kind &&
```
?

sdkrystian: It _could_ used `static_kind`, but since every other javadoc node uses `kind` I'm not going to use `static_kind` for just these two. At some point we can just change all the nodes to use `static_kind`.

PR #622, opened 2024-06-06
Complete Documentation Pages

alandefreitas:
The PR completes the documentation to reflect the stage of this project properly. 

- All configuration options are documented, and a reference table is included.
- Demo variants that were causing bugs were fixed and restored
- The Antora UI was adjusted to get rid of components that don't make sense for MrDocs
- The home page content was simplified;
- An Antora extension was implemented to include direct links to all demos in mrdocs.com
- An Antora extension was implemented to include direct links to all releases on github.com
- The commands page includes supported commands and examples of documented code
- The design notes include the rationale for both Mr. Docs and automated documentation in general
- Complete sections in the contributor guide, including a codebase overview for new contributors
- README.adoc links to the documentation with a banner instead of attempting to replicate the content

This PR also includes several commits fixing smaller problems with CI, such as relevant information about errors not being output and wrong or insufficient error messages.

Unfortunately, GitHub updated the Windows runner while this PR was WIP, which broke CI for unrelated reasons:

- The new runner image can't build LLVM by itself anymore
- Our tests are now segfaulting. 

This new image affects both this PR and unrelated PRs, but I've been trying to fix CI in passing so that we don't push new PRs when CI is broken. Among several commits fixing smaller problems with CI, the solution to the first problem was to re-upload LLVM Windows binaries to mrdocs.com and make the LLVM step fall back to these binaries until we can build LLVM in CI again. 

However, I tried everything but couldn't fix the second problem. I can't even reproduce it locally and ran out of ideas. Any ideas are welcome here.


include/mrdocs/Metadata/Info.hpp
```diff
@@ -200,7 +132,7 @@ struct MRDOCS_VISIBLE
     class is a certain type.
 */
 template <InfoKind K>
-struct IsInfo : Info
+struct TypedInfo : Info
```
sdkrystian, line 135: I prefer we don't change the name of this, or use `Type` anywhere in `InfoNodes.inc`. We already have `TypeInfo`, so this makes it pretty confusing. Prefer `Kind` for the names of the macro parameter.

alandefreitas: > We already have TypeInfo, so this makes it pretty confusing. 

Mmm... Good thinking. I forgot about TypeInfo. That *is* confusing. You can also see why `IsInfo` is confusing: it's not a trait, and it has nothing to do with what it does. Is there another name you could suggest? 

 -InfoType
- InfoCategory
- ClassifiedInfo
- SpecificInfo
- CategorizedInfo
- TypeSpecificInfo
- TypedMetadata
- InfoDiscriminator
- InfoVariant
- TypeDefinedInfo
- ...

> Prefer Kind for the names of the macro parameter

I don't know what you mean here.

sdkrystian: @alandefreitas `IsInfoKind`? Regarding the name of the macro parameter, I'd use `KIND` rather than `TYPE`.

alandefreitas: I think the `Is*` prefix makes it look like a trait, though. Maybe that's a convention somewhere else, but I've never seen an `is*` class that's not a trait.

sdkrystian: `InfoBase`? `InfoCommon`? `InfoCommonBase`? Anything that doesn't contain "type" :)

alandefreitas: I think I like InfoCommonBase for now.

alandefreitas: Done

PR #636, opened 2024-06-26
detect types implementing SFINAE idiom

sdkrystian:


src/lib/AST/ASTVisitor.cpp
```diff
@@ -1456,6 +1456,318 @@ class ASTVisitor
...
+        return NTTPD;
+    }
+
+    std::optional<TemplateArgument>
+    tryGetTemplateArgument(
+        TemplateParameterList* Parameters,
+        ArrayRef<TemplateArgument> Arguments,
+        unsigned Index)
+    {
+        if(Index == -1)
```
alandefreitas, line 1508: Never true :)

PR #640, opened 2024-07-05
East-const style applied to documentation generation

### LINKED ISSUE #639: East const types
https://843.urlantora.prtest2.cppalliance.org/site/url/reference/boost/urls/grammar/ref.html

```cpp
template<class CharSet>
constexpr
implementation-defined
ref(const CharSet& cs) noexcept;
```

should be

```cpp
template<class CharSet>
constexpr
implementation-defined
ref(CharSet const& cs) noexcept;
```

fpelliccioni:
Fixes #639


vinniefalco:
Should this be configurable?

PR #648, opened 2024-07-24
Standard library incompatibility fix by packaging libc++

fpelliccioni:
Fixes https://github.com/cppalliance/mrdocs/issues/644


include/mrdocs/Config.hpp
```diff
@@ -104,6 +104,12 @@ class MRDOCS_DECL
         /// from CMakeLists.txt
         std::string cmake;
 
+        /// True if the compiler has to use just the system's standard library
+        bool useSystemStandardLib = false;
+
+        /// LibC++ include paths (private option, not exposed to the user)
+        std::vector<std::string> libCxxPaths;
```
alandefreitas, line 111: This needs to go to TestArgs, where private options go and where this is used. These configuration files are even going to be automatically generated in the future.

Then you'll need a way to pass this libCxxPaths to the generator without Config, but that's not a huge problem. Another place for private config variables is ConfigImpl, but I'm not sure it applies here. Another solution is to make it public because users will have access to the option anyway, and this comment won't stop them. 

Also, if this is public, we need to discuss the design in a way we don't need for private options. For instance, if this is just a standard library path, it's not a libCxxPath because it's supposed to work with any standard library.

src/lib/Lib/ConfigImpl.cpp
```diff
@@ -83,6 +84,39 @@ ConfigImpl(
     }
     settings_.configDir = files::makeDirsy(files::normalizePath(settings_.configDir));
 
+    if (settings_.libCxxPaths.empty())
```
alandefreitas, line 87: setupAddonsDir in Addons.hpp does that for addons. Does one of these designs need to be better?

src/lib/Lib/MrDocsCompilationDatabase.cpp
```diff
@@ -274,18 +276,32 @@ adjustCommandLine(
...
+    else
+    {
+        // ------------------------------------------------------
+        // Add LibC++ and Clang include directories
+        // ------------------------------------------------------
+
+        for (auto const& inc : libCxxIncludeDirs)
         {
-            new_cmdline.emplace_back(fmt::format("-I{}", inc));
+            new_cmdline.emplace_back(fmt::format("-stdlib++-isystem{}", inc));
```
alandefreitas, line 304: The first path uses `-isystem` and the second one uses `-stdlib++-isystem`?

src/test/TestArgs.hpp
```diff
@@ -47,6 +47,7 @@ class TestArgs
     llvm::cl::opt<bool>         unitOption;
     llvm::cl::list<std::string> inputPaths;
     llvm::cl::opt<std::string>  addons;
+    llvm::cl::list<std::string> libCxxPaths;
```
alandefreitas, line 50: There's a 1-to-1 relationship between Config and ToolArgs. Conceptually, they both derive from ConfigOptions.inc. We have yet to be able to automate all steps, but the logic relies on that premise. The thing can't be in Config and TestArgs because they mean different things.

alandefreitas (commented):
This is great! :)

This incompatibility has been causing us lots of problems.

.github/workflows/ci.yml
```diff
@@ -526,7 +591,7 @@ jobs:
                   cat "$previous_file" >> "$diff_output"
               fi
           done
-          
+
```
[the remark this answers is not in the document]
fpelliccioni: It's true it's annoying. But style changes are fine. In the future we should have (non-annoying) rules so that things like trailing spaces are removed, each of us in our own IDE.
One thing that is wrong is mixing different things in the same PR. So I'm going to create a small PR to modify these style changes, and when it's merged into develop this will no longer be in the way of this PR and will not hinder its review.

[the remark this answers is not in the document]
fpelliccioni: https://github.com/cppalliance/mrdocs/pull/651

CMakeLists.txt
```diff
@@ -92,6 +92,12 @@ elseif (LLVM_ROOT)
 endif()
 find_package(LLVM REQUIRED CONFIG)
 find_package(Clang REQUIRED CONFIG)
+
+if (LLVM_ROOT)
+    set(LIBCXX_DIR "${LLVM_ROOT}/../install-libcxx/include/c++/v1/")
```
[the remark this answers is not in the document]
fpelliccioni: Same as https://github.com/cppalliance/mrdocs/pull/648#discussion_r1691229674

CMakeLists.txt
```diff
@@ -319,6 +325,22 @@ if (MRDOCS_BUILD_TESTS)
...
+#-------------------------------------------------
+#
+# Copy libc++ and clang headers
+#
+#-------------------------------------------------
+
+set(MRDOCS_SHARE_LIBCXX_DIR "${CMAKE_CURRENT_BINARY_DIR}/../share/mrdocs/libcxx")
+message(STATUS "Copying libc++ headers to ${MRDOCS_SHARE_LIBCXX_DIR}")
+file(MAKE_DIRECTORY ${MRDOCS_SHARE_LIBCXX_DIR})
+file(COPY ${LIBCXX_DIR}/ DESTINATION ${MRDOCS_SHARE_LIBCXX_DIR})
```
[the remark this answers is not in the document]
fpelliccioni: The files are placed here because they're needed not only for `if (MRDOCS_INSTALL)` but also during the `if (MRDOCS_BUILD_DOCS)` step for documentation generation. We need them accessible in both contexts.

include/mrdocs/Config.hpp
fpelliccioni: I actually prefer to keep this option and make it public. With well-set defaults, I believe offering flexibility to the user is beneficial. 
What do you think?

.github/workflows/ci.yml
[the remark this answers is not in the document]
alandefreitas: > It's true it's annoying. But style changes are fine. 

I agree with annoying more than fine. lol

> In the future we should have (non-annoying) rules so that things like trailing spaces are removed, each of us in our own IDE.

We don't need rules for that. It's just a matter of not changing the lines, whatever they look like currently. The rule is "don't introduce unintended changes in formatting".

> One thing that is wrong is mixing different things in the same PR. So I'm going to create a small PR to modify these style changes, 

That would be one PR introducing the changes and another reverting the changes. That doesn't make sense because none of the PRs are merged yet. So we can just not introduce the changes in the first plane. You can do that by reverting the changes, putting them in a commit, and squashing them with the commit that introduced the changes. If the commit is far behind, you can interactively rebase the history to do it. If you're not used to doing that, just take your time because that's something we need to do all the time so it's a skill we need.

CMakeLists.txt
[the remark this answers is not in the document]
alandefreitas: OK. I'm assuming the MRDOCS_BUILD_DOCS uses the addons option. So let's just make the libcxx config option public then. This way we can avoid copying the files, which is bad workaround. 

The `if (MRDOCS_INSTALL)` step also needs a step to install these files, IIRC. If you're not installing these files, the demos can't find them and are probably falling back to something else.

include/mrdocs/Config.hpp
alandefreitas: No problem. 

If it's always a single path, it's just a string option then. Not a vector. And the variable name would not be in the plural: `libcxx` instead of `libcxxs`

Would this also work with other standard library implementation headers? I don't see why not. In that case, the option name can't be `libcxx`. We can use the same name compilers use for the standard library path.

.github/workflows/ci.yml
```diff
@@ -253,6 +261,63 @@ jobs:
...
+                  -D CMAKE_INSTALL_PREFIX="$llvm_project_root/install-libcxx"
+
+            ninja -C build-libcxx cxx cxxabi unwind
+            ninja -C build-libcxx install-cxx install-cxxabi install-unwind
+          else
+            cmake -G Ninja \
+                  -S runtimes \
+                  -B build-libcxx \
+                  -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi" \
+                  -DCMAKE_INSTALL_PREFIX="$llvm_project_root/install-libcxx" \
```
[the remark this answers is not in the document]
fpelliccioni: I’ll use `$llvm_project_root/install/libcxx` for installing libc++, to avoid conflicts with existing LLVM source directories

.github/workflows/ci.yml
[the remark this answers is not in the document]
fpelliccioni: > > It's true it's annoying. But style changes are fine.
> 
> I agree with annoying more than fine. lol

I meant that it's not ideal to mix style changes with functional changes in the same PR—it diverts focus from the main purpose. The current code does have trailing spaces, which isn’t great either. That’s why I proposed addressing it in a separate PR to keep things clean and focused.

> You can do that by reverting the changes, putting them in a commit, and squashing ...

 I realize there might have been some confusion about my approach. Here’s what I was thinking:
 
1.  #651 for Style Changes: I've already created #651, which exclusively contains the style changes that are also currently in #648.

2. Merge #651: The plan is to merge #651 first, which includes all the style changes.

3. Rebase #648: After #651 is merged, I'll rebase #648. This will effectively remove the style changes from #648’s diff, as those changes will now be part of the main branch.

This strategy ensures that #648 focuses strictly on functionality without the noise of style changes. Does this sound like a good approach to you?

CMakeLists.txt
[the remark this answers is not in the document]
fpelliccioni: 👍

include/mrdocs/Config.hpp
fpelliccioni: > No problem.
> 
> If it's always a single path, it's just a string option then. Not a vector. And the variable name would not be in the plural: `libcxx` instead of `libcxxs`

It should remain a vector to accommodate multiple paths.


> Would this also work with other standard library implementation headers? I don't see why not. In that case, the option name can't be `libcxx`. We can use the same name compilers use for the standard library path: `stdlib`.

I'll apply the name change to stdlib.

.github/workflows/ci.yml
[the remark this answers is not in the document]
alandefreitas: The only problem it caused was some annoyance when reviewing the PR. Apart from that, they're just trailing spaces. 

We're discussing this like it's rocket science. We don't need to fix anything; leaving them as they are is okay. The project can move on with or without these tailing spaces. 

You can remove them from this PR by squashing the changes or leaving them as they are here, as the PR has already been reviewed this way. Whatever you decide, we can keep it the new way.

include/mrdocs/Config.hpp
alandefreitas: > accommodate multiple paths

My question is whether this is ever multiple paths.

> I'll apply the name change to stdlib.

Yes. This makes more sense.

.github/workflows/ci.yml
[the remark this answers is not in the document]
alandefreitas: The question is: do any conflicts exist?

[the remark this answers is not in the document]
fpelliccioni: I do not know... I am doing some tests right now to confirm it.
I  also moved LLVM installation to `$llvm_project_root/install/llvm`

include/mrdocs/Config.hpp
fpelliccioni: > My question is whether this is ever multiple paths.

I need to add libC++ and some Clang headers, for example. I realized this after doing several tests. It is likely that some other path will be needed in the future.

.github/workflows/ci.yml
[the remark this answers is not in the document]
alandefreitas: You'll need to update CI and the documentation then. CMake LLVM_ROOT is set to llvm_root, and llvm_root is set to ../third-party/llvm-project/install. 

Regarding the other problem: 

> The question is: do any conflicts exist?
> I do not know...

What you are doing is creating two Filesystem Hierarchy Standard (FHS) trees: one for LLVM and one for libcxx. 
But when you install LLVM and libcxx in your machine, you're installing it in a single FHS tree: /usr/local or /usr.

So you do know. There are no conflicts. There can't be any conflicts or otherwise it wouldn't be possible to install them in a single tree.

In principle, there's no problem having separate trees. In practice, however, that means your CMake script now has too look for two things: LLVM and libcxx, because they're not in the same tree. If they are in the same tree, you can find libcxx relative to the LLVM include directory. If they're not in the same tree, you need to give CMake this tree: some kind of LIBCXX_ROOT and find libcxx from there. And there's no built-in package to do that.

src/test/TestArgs.hpp
fpelliccioni: I noticed that `addons` is included in both `Config` and `TestArgs`, similar to my addition of `stdLibPaths` (aka: `libCxxPaths`). Given your feedback on avoiding duplication between these, should we revisit the placement of `addons`?

share/cmake/MrDocs.cmake
```diff
@@ -146,6 +146,12 @@ function(add_mrdocs MRDOCS_TARGET_NAME)
         set(MRDOCS_TARGET_OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/mrdocs)
     endif()
 
+    #-------------------------------------------------
+    # LibC++ paths
+    #-------------------------------------------------
+    if (NOT DEFINED MRDOCS_LIBCXX_PATHS)
+        set(MRDOCS_LIBCXX_PATHS ${CMAKE_SOURCE_DIR}/share/libcxx)
```
alandefreitas, line 153: I think 1) this script is not being supported and 2) if it were, the logic above is wrong.

I'm not even sure we should be maintaining this script. Klemens wrote this once, but it's kind of a workaround, and we haven't been supporting (testing) it. It's a huge workaround because it only makes sense when the target is there. In other words, it only makes sense inside the MrDocs project. 

In practice, if people want to create a target that generates MrDocs documentation, I expect them to have MrDocs installed and create a custom target with whatever command they want. Then we could have a MrDocs.cmake script bundled with the package to help them create these custom targets (like Catch2 bundles a script to create test targets) and let people use it. 

But this would need to be tested. For now, I think we can think of this script as deprecated because we're not really supporting or testing it.

In any case, if some other project includes this script, it assumes the other project has a copy of libcxx in `${CMAKE_SOURCE_DIR}/share/libcxx`, which is not the case. Even MrDocs shouldn't have a copy of libcxx there.

src/lib/Lib/MrDocsCompilationDatabase.cpp
```diff
@@ -222,7 +222,9 @@ std::vector<std::string>
 adjustCommandLine(
     const std::vector<std::string>& cmdline,
     const std::vector<std::string>& additional_defines,
-    std::unordered_map<std::string, std::vector<std::string>> const& implicitIncludeDirectories)
+    std::unordered_map<std::string, std::vector<std::string>> const& implicitIncludeDirectories,
+    std::vector<std::string> const& stdLibIncludeDirs,
```
alandefreitas, line 226: Aren't these two kind of the same thing?

alandefreitas (commented):
Nice...

include/mrdocs/ConfigOptions.inc
```diff
@@ -69,6 +69,8 @@ COMMON_OPTION (compilationDatabase     , compilation-database    , Path to the c
 // Build options
 COMMON_OPTION (cmake                   , cmake                   , CMake arguments when generating the compilation database from CMakeLists.txt)
 COMMON_OPTION (defines                 , defines                 , Additional defines passed to the compiler)
+COMMON_OPTION (useSystemStdLib         , use-system-stdlib       , True if the compiler has to use just the system standard library)
+COMMON_OPTION (stdLibPaths             , stdlib-path             , Standard Library include paths)
```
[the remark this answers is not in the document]
fpelliccioni: I'm not sure I fully understand what you're proposing, however...
This PR already incorporates `-isystem` with the `-stdlib-path` option.

To maintain focus, I propose that this PR remains centered exclusively on resolving issue #644 with the current implementation.
We can explore the addition of `-I` in a future PR, discussing its integration and benefits in more depth.
What do you think?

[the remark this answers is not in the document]
alandefreitas: > This PR already incorporates -isystem with the -stdlib-path option.

Yes. But it's the other way around. This PR should already incorporates -stdlib-path with the -isystem option. Because -isystem is the general case. Not -stdlib-path.

Let's just this new options `stdlib-paths` make it represent `-isystem` instead of `-stdlib` because, in our case, 1) `-stdlib` is just a subset of `-isystem` (we only need the headers in MrDocs - there's no linking), and 2) the syntax of `-stdlib` is `-stdlib=<library>` but that's not our syntax, which is confusing.

In other words, let's rename this. We already have `defines` for what is `PREDEFINED` in Doxygen. Let's create an `includes` option, which we can apply in the database as `-I`, and another option we can apply in the database as `-isystem`, so we can call it `system-includes`.

So instead of `stdlib-paths`, we will have `includes` and `system-includes`. Instead of `stdlib-paths`, we will use `system-includes` for these stdlib paths. This way, we already implement non-ambiguous options we'll need anyway and already solve the problem about whether `stdlib-paths` should be public or not.

fpelliccioni:
Great! Thanks por the detailed explanation.

include/mrdocs/ConfigOptions.inc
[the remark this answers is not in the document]
fpelliccioni: Thank you for clarifying that. I believe the main point of confusion here is that what I implemented as `-stdlib-path` is actually a search path for standard library header files, which in GCC and Clang is equivalent to using `-isystem`. So, it's essentially the same functionality, just under a different name.

I will add support for `-includes` (similar to `-I` in GCC terminology) and will rename `-stdlib-paths` to `-system-includes` ...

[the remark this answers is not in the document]
fpelliccioni: @alandefreitas 

I've recently encountered some compilation issues in my PR, which you can see here:

https://github.com/cppalliance/mrdocs/actions/runs/10212908784/job/28257619626#step:10:64

I managed to resolve these issues by modifying how we include LibC++:

`new_cmdline.emplace_back(fmt::format("-stdlib++-isystem{}", inc));`

Instead of using:

`new_cmdline.emplace_back(fmt::format("-isystem{}", inc));`

Using the latter `-isystem` results in the errors linked above.

Given that in what we now call `system-includes` we do not distinguish between C++ and C libraries, and our default `system-includes` already include LibC++ (for C++) and Clang includes (for C), I am thinking that we might need to differentiate between C and C++:

1. `includes`: include path for header files from user libraries, applicable to both C and C++ (indistinct here).
2. `system-includes`: include path for system library header files for the C language.
3. `system-includes-cxx`: include path for system library header files for the C++ language.

What are your thoughts?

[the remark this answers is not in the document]
alandefreitas: Yes. The headers in `/usr/local/share/mrdocs/libcxx/...` end up including the headers in `/usr/include`. Should this be allowed to happen at all? 

Looking at https://libcxx.llvm.org/UsingLibcxx.html, did you give clang the options to remove these standard paths? Like `-nostdinc++` and `-nostdlib++`.

I also don't see that `-stdlib++-isystem` option documented anywhere. Where did you find it? What is it supposed to do that `-nostdinc++`, `-nostdlib++`, and `-isystem` can't?

>  I am thinking that we might need to differentiate between C and C++

Why? MrDocs has no linking anyway. Even C and C++ compilers are often symlinks to the same executables.

> system-includes-cxx: include path for system library header files for the C++ language

maybe `system-includes-libcxx`, but I'm still failing to see what this translates to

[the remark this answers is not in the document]
fpelliccioni: > Yes. The headers in `/usr/local/share/mrdocs/libcxx/...` end up including the headers in `/usr/include`. Should this be allowed to happen at all?

That is peculiar. Although it is not well-documented, I believe this is due to the difference between `-stdlib++-isystem` and `-isystem`.

> Looking at https://libcxx.llvm.org/UsingLibcxx.html, did you give clang the options to remove these standard paths? Like `-nostdinc++` and `-nostdlib++`.

Yes, I did that in the past, trying various combinations until I found one that works, which is the one in the current code.

> I also don't see that `-stdlib++-isystem` option documented anywhere. Where did you find it? What is it supposed to do that `-nostdinc++`, `-nostdlib++`, and `-isystem` can't?

That option is listed here in this link:

https://clang.llvm.org/docs/ClangCommandLineReference.html

It does not explain the difference between `-isystem` and `-stdlib++-isystem` but it is designed for the C++ standard library according to the documentation: "Use directory as the C++ standard library include path". That's why I am proposing we add a similar option in MrDocs, to differentiate C++ StdLib from other includes.

> Why? MrDocs has no linking anyway. Even C and C++ compilers are often symlinks to the same executables.

What I am proposing is not related to linking. I believe Clang treats the inclusion of the C++ standard library specially, and this might be related to `#include_next`. I think if Clang uses headers that are in the directory specified by `-stdlib++-isystem`, it stops searching even if there is an `#include_next`.

> maybe `system-includes-libcxx`, but I'm still failing to see what this translates to

I suggested `system-includes-cxx` rather than `system-includes-libcxx` because `-stdlib++-isystem` is used to specify the include path for any C++ standard library implementation, not just LibC++.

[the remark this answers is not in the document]
alandefreitas: > Although it is not well-documented

I think this is documented. https://libcxx.llvm.org/UsingLibcxx.html says you need to use -nostdinc++ -nostdlib++ when using a custom std library. So, I assume it will try to include the usual system paths otherwise (like it did in CI). If it relies on system includes, we lose the reproducibility this PR intends to achieve.

> I did that in the past, trying various combinations until I found one that works

But if we have a problem above where it includes /usr/include, then it's more like it didn't work because the PR didn't achieve its goal. It just worked in the sense that it's not using the mrdocs packaged std library. My intuition is this PR without -nostdinc++/nostdlib++ has to be wrong, because we're now depending on system headers that are not reproducible and the problem remains.

> It does not explain the difference between -isystem and -stdlib++-isystem

Yes. The explanation is unclear. Even -cxx-isystem<directory> makes more sense because we know it would skip the option in C files. But -stdlib++-isystem is quite confusing. Anyway, as long as it is a distinct option for clang, it should be OK to support it. Just keep these links so we can reference them in the documentation. It would be good to have more information on the option, so we know what we're doing, though.

.github/workflows/ci.yml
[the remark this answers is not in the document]
fpelliccioni: Done

alandefreitas:
So only the documentation is missing now, right?

fpelliccioni:
> So only the documentation is missing now, right?

Just updated

CMakeLists.txt
```diff
@@ -284,11 +290,11 @@ if (MRDOCS_BUILD_TESTS)
     endif ()
     target_compile_definitions(mrdocs-test PRIVATE -DMRDOCS_TEST_FILES_DIR="${CMAKE_CURRENT_SOURCE_DIR}/test-files")
     add_test(NAME mrdocs-unit-tests COMMAND mrdocs-test --unit=true)
-    add_test(NAME mrdocs-golden-tests COMMAND mrdocs-test --unit=false --action=test "${PROJECT_SOURCE_DIR}/test-files/golden-tests" --addons="${CMAKE_SOURCE_DIR}/share/mrdocs/addons")
+    add_test(NAME mrdocs-golden-tests COMMAND mrdocs-test --unit=false --action=test "${PROJECT_SOURCE_DIR}/test-files/golden-tests" --addons="${CMAKE_SOURCE_DIR}/share/mrdocs/addons" --system-includes="${LIBCXX_DIR}")
```
alandefreitas, line 293: `CLANG_STDLIB_INCLUDE_DIR` was set in line 98 but never used here.

CMakeLists.txt
```diff
@@ -442,6 +448,13 @@ if (MRDOCS_INSTALL)
     #-------------------------------------------------
     # share
     #-------------------------------------------------
+    install(DIRECTORY ${LIBCXX_DIR}
+            DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/mrdocs/libcxx
+            FILES_MATCHING PATTERN "*")
+    install(DIRECTORY ${CLANG_STDLIB_INCLUDE_DIR}
+            DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/mrdocs/clang
```
alandefreitas, line 455: `clang` is bad naming unless this directory includes, well... the clang compiler.

src/test/TestArgs.cpp
```diff
@@ -80,6 +80,16 @@ R"(
     "addons",
     llvm::cl::desc("The directory with the addons."),
     llvm::cl::cat(commonCat))
+
+, systemIncludes(
+    "system-includes",
+    llvm::cl::desc("A list of paths to the standard library."),
```
alandefreitas, line 86: Isn't system-includes corresponding to `-isystem`? 

- If so, it's "A list of paths to the system headers.". 
- If not, then the `isystem` option is missing.

CMakeLists.txt
alandefreitas: I'm very suspicious about this. If you really need `CLANG_STDLIB_INCLUDE_DIR`, it's not supposed to work without it. My intuition is you're not providing CLANG_STDLIB_INCLUDE_DIR and the thing is working in CI because it finds it in the system as usual. 

The main problem related to this blocking us now is we're trying to build documentation in Ubuntu 20.04 because that's what the boost release tools use. To emulate that and be sure the binaries works elsewhere the `website` workflow in ci.yml used to use Ubuntu 20.04 to emulate the conditions in the boost release tools. The reason we have to disable 20.04 and use 23.04 is this very issue.

So there's one thing we have to do to make sure this PR achieves its goal. ci.yml should use ubuntu 20.04 in the website step. 

https://github.com/cppalliance/mrdocs/blob/e460d86aceed8784d2cd3c7dd04daef51d74e3a8/.github/workflows/ci.yml#L323

Note that we should set the `container` and not `runs-on` to ensure we are running it in a clean container. If mrdocs builds in this environment, it means it's using its own libcxx because no other stdlib can be found in system paths.

We also have to remove steps this PR is supposed to make irrelevant:

We don't need to install clang because mrdocs is already supposed to have everything it needs:

https://github.com/cppalliance/mrdocs/blob/e460d86aceed8784d2cd3c7dd04daef51d74e3a8/.github/workflows/ci.yml#L327-L333

No need to install libstdc++-12 because mrdocs shouldn't be using the system stdlib:

https://github.com/cppalliance/mrdocs/blob/e460d86aceed8784d2cd3c7dd04daef51d74e3a8/.github/workflows/ci.yml#L339C20-L339C36

We might need to adjust it to install `build-essential` though. The reason is not related to mrdocs per se, but because cmake needs at least one compiler to run the configuration step. 

PS: The only reason we need some dummy compiler in the system is to run the cmake configuration step. It's technically possible to make MrDocs + CMake work without an existing compiler though: mrdocs would need to provide itself to cmake as the compiler when needed. That's more complex and outside the scope of this issue, but it's worth knowing.

CMakeLists.txt
alandefreitas: Also `STDLIB_INCLUDE_DIR` is confusing. Isn't `LIBCXX_DIR` already an implementation of the standard library? What's in `STDLIB_INCLUDE_DIR`? How does it differ from `LIBCXX_DIR`?

.github/workflows/ci.yml
```diff
@@ -319,26 +385,29 @@ jobs:
         shell: bash
     name: Website
     timeout-minutes: 30
-    runs-on: ubuntu-20.04
-    container: ubuntu:23.04
+    # runs-on: ubuntu-20.04
```
alandefreitas, line 388: Please replace the content instead of commenting on the old one.

.github/workflows/ci.yml
```diff
@@ -319,26 +385,29 @@ jobs:
...
+      #     check-latest: true
 
       - name: Install packages
         uses: alandefreitas/cpp-actions/package-install@v1.8.3
         id: package-install
         with:
-          apt-get: libstdc++-12-dev asciidoctor cmake bzip2 git
-          cc: ${{ steps.setup-cpp.outputs.cc }}
-          cxx: ${{ steps.setup-cpp.outputs.cxx }}
+          # apt-get: libstdc++-12-dev asciidoctor cmake bzip2 git
```
alandefreitas, line 407: Same

.github/workflows/ci.yml
```diff
@@ -319,26 +385,29 @@ jobs:
...
       contents: write
     steps:
-      - name: Setup C++
-        uses: alandefreitas/cpp-actions/setup-cpp@v1.8.3
-        id: setup-cpp
-        with:
-          compiler: clang
-          version: 16
-          check-latest: true
+      # - name: Setup C++
```
alandefreitas, line 395: Same

.github/workflows/ci.yml
```diff
@@ -319,26 +385,29 @@ jobs:
...
       - name: Install packages
         uses: alandefreitas/cpp-actions/package-install@v1.8.3
         id: package-install
         with:
-          apt-get: libstdc++-12-dev asciidoctor cmake bzip2 git
-          cc: ${{ steps.setup-cpp.outputs.cc }}
-          cxx: ${{ steps.setup-cpp.outputs.cxx }}
+          # apt-get: libstdc++-12-dev asciidoctor cmake bzip2 git
+          apt-get: build-essential asciidoctor cmake bzip2 git
+          # cc: ${{ steps.setup-cpp.outputs.cc }}
```
alandefreitas, line 409: package-install needs to know where the compilers are in case we include vcpkg dependencies

.github/workflows/ci.yml
```diff
@@ -361,8 +430,8 @@ jobs:
           mrdocs --version
           MRDOCS_ROOT=/usr/local
           echo -e "MRDOCS_ROOT=$MRDOCS_ROOT" >> $GITHUB_ENV
-          CXX=/usr/bin/clang-16
-          echo -e "CXX=$CXX" >> $GITHUB_ENV
+          # CXX=/usr/bin/clang-16
```
alandefreitas, line 433: Same

.github/workflows/ci.yml
```diff
@@ -378,6 +447,7 @@ jobs:
       - name: Generate Landing Page
         working-directory: docs/website
         run: |
+          gcc -v -E -x c++ -isystem/usr/local/share/mrdocs/libcxx/ -isystem/usr/local/share/mrdocs/clang/ -nostdinc++ -nostdlib -
```
alandefreitas, line 450: What's happening here?

.github/workflows/ci.yml
```diff
@@ -415,11 +485,11 @@ jobs:
 
       - name: Generate Demos
         run: |
-          CXX="${{ steps.setup-cpp.outputs.cxx }}"
-          export CXX
-          CC="${{ steps.setup-cpp.outputs.cc }}"
-          export CC
-          
+          # CXX="${{ steps.setup-cpp.outputs.cxx }}"
```
alandefreitas, line 488: Same

.github/workflows/ci.yml
```diff
@@ -253,6 +262,63 @@ jobs:
...
+          cd ..
+          llvm_project_root="$(pwd)/third-party/llvm-project"
+          if [ ! -d "$llvm_project_root/runtimes" ]; then
+              mkdir -p $llvm_project_root
+              cd $llvm_project_root
+              git config --global init.defaultBranch master
+              git config --global advice.detachedHead false
+              git init
+              git remote add origin https://github.com/llvm/llvm-project.git
+              git fetch --depth 1 origin ${{ steps.llvm-parameters.outputs.llvm-hash }}
```
alandefreitas, line 288: Why are we cloning the same repo again in a subdirectory?

CMakeLists.txt
```diff
@@ -284,11 +290,11 @@ if (MRDOCS_BUILD_TESTS)
     endif ()
     target_compile_definitions(mrdocs-test PRIVATE -DMRDOCS_TEST_FILES_DIR="${CMAKE_CURRENT_SOURCE_DIR}/test-files")
     add_test(NAME mrdocs-unit-tests COMMAND mrdocs-test --unit=true)
-    add_test(NAME mrdocs-golden-tests COMMAND mrdocs-test --unit=false --action=test "${PROJECT_SOURCE_DIR}/test-files/golden-tests" --addons="${CMAKE_SOURCE_DIR}/share/mrdocs/addons")
+    add_test(NAME mrdocs-golden-tests COMMAND mrdocs-test --unit=false --action=test "${PROJECT_SOURCE_DIR}/test-files/golden-tests" --addons="${CMAKE_SOURCE_DIR}/share/mrdocs/addons" --system-includes="${LIBCXX_DIR}" --system-includes="${STDLIB_INCLUDE_DIR}")
     foreach (action IN ITEMS create update)
         add_custom_target(
             mrdocs-${action}-test-fixtures
-            COMMAND mrdocs-test --unit=false --action=${action} "${PROJECT_SOURCE_DIR}/test-files/golden-tests" --addons="${CMAKE_SOURCE_DIR}/share/mrdocs/addons"
+            COMMAND mrdocs-test --unit=false --action=${action} "${PROJECT_SOURCE_DIR}/test-files/golden-tests" --addons="${CMAKE_SOURCE_DIR}/share/mrdocs/addons" --system-includes="${LIBCXX_DIR}"
```
alandefreitas, line 297: Why does this one only need LIBCXX_DIR?

src/lib/Lib/MrDocsCompilationDatabase.cpp
```diff
@@ -274,19 +277,42 @@ adjustCommandLine(
...
+    if (useSystemStdLib)
+    {
+        // ------------------------------------------------------
+        // Add implicit include paths
+        // ------------------------------------------------------
+        // Implicit include paths are those which are automatically
+        // added by the compiler. These will not be defined in the
+        // compile command, so we add them here so that clang
+        // can also find these headers.
+        if (auto const it = implicitIncludeDirectories.find(progName);
```
alandefreitas, line 289: Besides checking the use of SystemStdLib, the input here should be empty. We don't want to calculate it if we're not using it.

alandefreitas (commented):
Looking good

.github/workflows/ci.yml
fpelliccioni: But in this case it is not necessary, because we are only including apt-get dependencies. 
Is there any vcpkg dependency that is being installed by package-install internally that is not evident?

.github/workflows/ci.yml
fpelliccioni: Just temporary "testing" the defaults include paths give those flags. Removing ...

.github/workflows/ci.yml
fpelliccioni: It is cloning to the same directory: `third-party/llvm-project`

alandefreitas: Yes. But again?

PR #649, opened 2024-07-24
Support concepts and constraints

sdkrystian:


alandefreitas:
We must fix that `Build Warning - clang++-17 - [-Wsign-compare]`. Not because of the type mismatch but because the condition is always `false`, and I suspect this is hiding some bug in the condition you were expected to be true sometimes. It was introduced in your previous PR.

PR #659, opened 2024-08-15
Config target

alandefreitas:
This PR includes the reference for the config options in the documentation. It implements a solution similar to Tablegen to describe the configuration options. The same information is used as a source for the configuration files, the command line options, and the documentation. An Antora extension parses this source file and generates the corresponding reference sections. With this PR, the configuration options in the code and the documentation will always be in sync.


alandefreitas:
@sdkrystian @fpelliccioni Just a heads up. CMakeLists.txt will require python after this PR.

Issue #650, opened 2024-07-25
Tagfiles


alandefreitas:
Generators should generate tagfiles in the doxygen format. Tagfiles contain a list of symbols and their relative urls. They allow for documentation to refer to symbols in other documentation.

Here are some examples:

https://github.com/alandefreitas/url/blob/update-antora/doc/lib/cpp_tagfiles/cppreference-doxygen-web.tag.xml
https://github.com/alandefreitas/url/tree/update-antora/doc/tagfiles

We already have an Antora extension for the Antora documentation to use these tagfiles. I copied some logic from the collector extension to allow components to set their tagfiles besides tagfiles that might be set in the playbook.

PR #670, opened 2024-09-12
feat: tagfiles generation for cross-referencing in doxygen format

fpelliccioni:
Fixes https://github.com/cppalliance/mrdocs/issues/650

### DIFF
--- a/include/mrdocs/Corpus.hpp
+++ b/include/mrdocs/Corpus.hpp
@@ -143,6 +143,7 @@ public:
         The type of the first argument is determined
         by the `InfoKind` of the `Info` object.
 
+        @param I The Info to visit.
         @param info The Info to visit.
         @param f The function to invoke.
         @param args The arguments to pass to the function.
@@ -202,6 +203,14 @@ public:
     getFullyQualifiedName(
         const Info& I,
         std::string& temp) const;
+
+    std::string
+    getFullyQualifiedName(const Info& I) const
+    {
+        std::string temp;
+        return getFullyQualifiedName(I, temp);
+    }
+
 };
 
 //------------------------------------------------
--- a/include/mrdocs/Generator.hpp
+++ b/include/mrdocs/Generator.hpp
@@ -182,6 +182,29 @@ public:
         Corpus const& corpus) const;
 };
 
+/** Return the full path for single page output.
+
+    This function determines the full path for a single-page output file
+    based on the provided `outputPath` and file `extension`.
+
+    If the `outputPath` already exists:
+    - If it is a directory, appends the default file name with the provided extension.
+    - If it is a file, uses the provided `outputPath` directly.
+
+    If the `outputPath` does not exist:
+    - If it ends with a '/', assumes it is a directory and appends the default file name.
+    - Otherwise, it returns an error because the path is ambiguous.
+
+    @return The full path or an error if the `outputPath` is ambiguous.
+
+    @param outputPath The specified output path, which can be a directory or file.
+    @param extension The file extension to use for single-page output.
+*/
+Expected<std::string>
+getSinglePageFullPath(
+    std::string_view outputPath,
+    std::string_view extension);
+
 } // mrdocs
 } // clang
 
--- a/src/lib/Gen/hbs/HandlebarsGenerator.cpp
+++ b/src/lib/Gen/hbs/HandlebarsGenerator.cpp
@@ -15,7 +15,17 @@
 #include "Builder.hpp"
 #include "MultiPageVisitor.hpp"
 #include "SinglePageVisitor.hpp"
+
+#include "lib/Lib/TagfileWriter.hpp"
+#include "lib/Support/RawOstream.hpp"
+
+#include <llvm/ADT/SmallString.h>
+#include <llvm/Support/FileSystem.h>
+#include <llvm/Support/Path.h>
+
 #include <mrdocs/Support/Path.hpp>
+
+#include <fstream>
 #include <sstream>
 
 namespace clang {
@@ -79,7 +89,21 @@ build(
 {
     if (!corpus.config->multipage)
     {
-        return Generator::build(outputPath, corpus);
+        auto e = Generator::build(outputPath, corpus);
+        if (e.has_value() && !corpus.config->tagfile.empty())
+        {
+            // Generate tagfile if specified
+            auto const singlePagePath = getSinglePageFullPath(outputPath, fileExtension());
+            MRDOCS_CHECK_OR(singlePagePath, Unexpected(singlePagePath.error()));
+            HandlebarsCorpus hbsCorpus = createDomCorpus(*this, corpus);
+            MRDOCS_TRY(auto tagFileWriter, TagfileWriter::create(
+                    hbsCorpus,
+                    corpus.config->tagfile,
+                    files::getFileName(*singlePagePath)));
+            tagFileWriter.build();
+        }
+
+        return e;
     }
 
     // Create corpus and executors
@@ -93,6 +117,16 @@ build(
     // Wait for all executors to finish and check errors
     auto errors = ex.wait();
     MRDOCS_CHECK_OR(errors.empty(), Unexpected(errors));
+
+    if (! corpus.config->tagfile.empty())
+    {
+        MRDOCS_TRY(auto tagFileWriter, TagfileWriter::create(
+                domCorpus,
+                corpus.config->tagfile,
+                outputPath));
+        tagFileWriter.build();
+    }
+
     return {};
 }
 
--- a/src/lib/Lib/TagfileWriter.cpp
+++ b/src/lib/Lib/TagfileWriter.cpp
@@ -1,0 +1,281 @@
+//
+// This is a derivative work. originally part of the LLVM Project.
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (c) 2024 Fernando Pelliccioni (fpelliccioni@gmail.com)
+//
+// Official repository: https://github.com/cppalliance/mrdocs
+//
+
+#include "../Gen/xml/CXXTags.hpp"
+#include "TagfileWriter.hpp"
+#include "lib/Lib/ConfigImpl.hpp"
+#include "lib/Support/RawOstream.hpp"
+
+#include <llvm/Support/FileSystem.h>
+
+#include <mrdocs/Support/Path.hpp>
+
+namespace clang {
+namespace mrdocs {
+
+//------------------------------------------------
+//
+// TagfileWriter
+//
+//------------------------------------------------
+
+TagfileWriter::
+TagfileWriter(
+    hbs::HandlebarsCorpus const& corpus,
+    os_ptr os,
+    std::string_view const defaultFilename
+    ) noexcept
+    : corpus_(corpus)
+    , os_(std::move(os))
+    , tags_(*os_)
+    , defaultFilename_(defaultFilename)
+{
+    tags_.nesting(false);
+}
+
+Expected<TagfileWriter>
+TagfileWriter::
+create(
+    hbs::HandlebarsCorpus const& corpus,
+        std::string_view tagfile,
+        std::string_view defaultFilename)
+{
+    std::error_code ec;
+
+    auto os = std::make_unique<llvm::raw_fd_ostream>(
+        tagfile.data(),
+        ec,
+        llvm::sys::fs::OF_None);
+
+    if (ec)
+    {
+        return Unexpected(formatError("llvm::raw_fd_ostream(\"{}\") failed with error: {}", tagfile, ec.message()));
+    }
+    return TagfileWriter(corpus, std::move(os), defaultFilename);
+}
+
+void
+TagfileWriter::
+build()
+{
+    initialize();
+    operator()(corpus_->globalNamespace());
+    finalize();
+}
+
+void
+TagfileWriter::
+initialize()
+{
+    (*os_) << "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?>\n";
+    (*os_) << "<tagfile>\n";
+}
+
+void
+TagfileWriter::
+finalize()
+{
+    (*os_) << "</tagfile>\n";
+}
+
+
+template<class T>
+void
+TagfileWriter::
+operator()(T const& I)
+{
+    if constexpr (T::isNamespace())
+    {
+        // Namespaces are compound elements with members
+        writeNamespace(I);
+    }
+    else if (!T::isFunction())
+    {
+        // Functions are described as namespace members in the
+        // scoped they belong to.
+        // Everything else is described as a compound element of
+        // type "class" because it's the type doxygen supports.
+        writeClassLike(I);
+    }
+}
+
+#define INFO(Type) template void TagfileWriter::operator()<Type##Info>(Type##Info const&);
+#include <mrdocs/Metadata/InfoNodesPascal.inc>
+
+void
+TagfileWriter::
+writeNamespace(
+    NamespaceInfo const& I)
+{
+    // Check if this namespace contains only other namespaces
+    bool onlyNamespaces = true;
+    corpus_->traverse(I, [&](Info const& I)
+    {
+        if (I.Kind != InfoKind::Namespace)
+        {
+            onlyNamespaces = false;
+            return false;
+        }
+        return true;
+    });
+
+    // Write the compound element for this namespace
+    if (!onlyNamespaces)
+    {
+        tags_.open("compound", {
+            { "kind", "namespace" }
+        });
+
+        tags_.write("name", corpus_->getFullyQualifiedName(I));
+        tags_.write("filename", generateFilename(I));
+
+        // Write the class-like members of this namespace
+        corpus_->traverse(I, [this]<typename U>(U const& J)
+        {
+            if (!U::isNamespace() && !U::isFunction())
+            {
+                tags_.write(
+                    "class",
+                    corpus_->getFullyQualifiedName(J),
+                    {{"kind", "class"}});
+            }
+        });
+
+        // Write the function-like members of this namespace
+        corpus_->traverse(I, [this]<typename U>(U const& J)
+        {
+            if constexpr (U::isFunction())
+            {
+                writeFunctionMember(J);
+            }
+        });
+
+        tags_.close("compound");
+    }
+
+    // Write compound elements for the members of this namespace
+    corpus_->traverse(I, [this]<typename U>(U const& J)
+        {
+            this->operator()(J);
+        });
+}
+
+template<class T>
+void
+TagfileWriter::
+writeClassLike(
+    T const& I
+)
+{
+    tags_.open("compound", {
+        { "kind", "class" }
+    });
+    tags_.write("name", corpus_->getFullyQualifiedName(I));
+    tags_.write("filename", generateFilename(I));
+    if constexpr (T::isRecord())
+    {
+        // Write the function-like members of this record
+        corpus_->traverse(I, [this]<typename U>(U const& J)
+        {
+            if constexpr (U::isFunction())
+            {
+                writeFunctionMember(J);
+            }
+        });
+    }
+    tags_.close("compound");
+}
+
+void
+TagfileWriter::
+writeFunctionMember(FunctionInfo const& I)
+{
+    tags_.open("member", {{"kind", "function"}});
+    tags_.write("type", toString(*I.ReturnType));
+    tags_.write("name", I.Name);
+    auto [anchorFile, anchor] = generateFileAndAnchor(I);
+    tags_.write("anchorfile", anchorFile);
+    tags_.write("anchor", anchor);
+    std::string arglist = "(";
+    for(auto const& J : I.Params)
+    {
+        arglist += toString(*J.Type);
+        arglist += " ";
+        arglist += J.Name;
+        arglist += ", ";
+    }
+    if (arglist.size() > 2) {
+        arglist.resize(arglist.size() - 2);
+    }
+    arglist += ")";
+    tags_.write("arglist", arglist);
+    tags_.close("member");
+}
+
+
+template<class T>
+std::string
+TagfileWriter::
+generateFilename(T const& I)
+{
+    std::string url = corpus_.getURL(I);
+    // In single page mode, getURL always returns an
+    // anchor.
+    if (!corpus_->config->multipage) {
+        // filename is #<anchor>
+        if (!url.starts_with('#'))
+        {
+            return defaultFilename_ + '#' + url;
+        }
+        return defaultFilename_ + url;
+    }
+    // In multipage mode, getURL returns a file path
+    // starting with a slash, and it may contain an anchor
+    if (url.starts_with('/'))
+    {
+        url.erase(0, 1);
+    }
+    return url;
+}
+
+template<class T>
+std::pair<std::string, std::string>
+TagfileWriter::
+generateFileAndAnchor(T const& I)
+{
+    std::string url = corpus_.getURL(I);
+    // Make relative to output dir
+    if (url.starts_with('/'))
+    {
+        url.erase(0, 1);
+    }
+    // In single page mode, getURL always returns an
+    // anchor. So the filename is the default and the
+    // anchor is the URL.
+    if (!corpus_->config->multipage) {
+        if (url.starts_with('#'))
+        {
+            url.erase(0, 1);
+        }
+        return {defaultFilename_, url};
+    }
+    // In multipage mode, getURL returns a file path
+    // starting with a slash, and it may contain an anchor
+    auto const pos = url.find('#');
+    if (pos == std::string::npos)
+    {
+        return {url, ""};
+    }
+    return {url.substr(0, pos), url.substr(pos + 1)};
+}
+
+} // mrdocs
+} // clang
--- a/src/lib/Lib/TagfileWriter.hpp
+++ b/src/lib/Lib/TagfileWriter.hpp
@@ -1,0 +1,118 @@
+//
+// This is a derivative work. originally part of the LLVM Project.
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (c) 2024 Fernando Pelliccioni (fpelliccioni@gmail.com)
+//
+// Official repository: https://github.com/cppalliance/mrdocs
+//
+
+#ifndef MRDOCS_LIB_TAGFILEWRITER_HPP
+#define MRDOCS_LIB_TAGFILEWRITER_HPP
+
+#include "../Gen/xml/XMLTags.hpp"
+#include <mrdocs/Metadata.hpp>
+#include "lib/Gen/hbs/HandlebarsCorpus.hpp"
+#include <mrdocs/Support/Error.hpp>
+
+#include <fstream>
+#include <string>
+
+namespace clang {
+namespace mrdocs {
+
+struct SimpleWriterTag {}; //Tag dispatch for simple writers
+
+class jit_indenter;
+
+/** A writer which outputs Tagfiles.
+*/
+class TagfileWriter
+{
+    using os_ptr = std::unique_ptr<llvm::raw_fd_ostream>;
+
+    hbs::HandlebarsCorpus const& corpus_;
+    os_ptr os_;
+    xml::XMLTags tags_;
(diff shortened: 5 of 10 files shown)


include/mrdocs/Corpus.hpp
```diff
@@ -160,6 +160,39 @@ class MRDOCS_VISIBLE
...
+        @param args The arguments to pass to the function.
+    */
+    template <InfoParent T, class Pred, class F, class... Args>
+    void
+    traverseIf(
+        T const& I, Pred&& pred, F&& f, Args&&... args) const
+    {
+        for (auto const& id : I.Members)
+        {
+            if (pred(get(id)))
```
alandefreitas, line 188: `std::forward<Pred>(pred)(get(id))` ? I'm not sure.

include/mrdocs/Generator.hpp
```diff
@@ -64,6 +64,7 @@ class MRDOCS_VISIBLE
         @li "adoc" Asciidoctor
         @li "xml" XML
         @li "html" HTML
+        @li "tagfile" Tagfile
```
alandefreitas, line 67: Shouldn't the tagfile generator be associated with a generator? Different generators with different options would generate different tagfiles and there's always a 1-to-1 relationship. 

For instance, at the very least, https://github.com/boostorg/url/blob/develop/doc/tagfiles/boost-url-doxygen.tag.xml uses .adoc for the file extensions. And any option that influences what files are generated should be reflected in the tagfile. So some generators, tagfiles don't make much sense.

It also seems easier to implement.

alandefreitas (commented):
The logic for the Writer looks great. I think the only issue is where it would be called from.

alandefreitas:
The problem CI has been fixed. Could you rebase this?

alandefreitas:
However, we will still have a problem with the USR from Krystian's commit blocking CI. 

I want to start by comparing the XML in the artifacts. Do you have the result for Boost.URL multipage? That's the one we can't risk breaking because it's already in production. It has to be compatible with the existing one.

alandefreitas:
It fails with

```
assertion failed: n <= indent_.size() on line 203 in src/lib/Gen/xml/XMLTags.cpp
```

alandefreitas (commented):
It fails with

```
assertion failed: n <= indent_.size() on line 203 in src/lib/Gen/xml/XMLTags.cpp
```

alandefreitas:
The implementation of the tagfile write is excellent.

CI failed after I rebased. That was a good thing because it helped me catch a bug. 

So, the reason CI is failing is trivial. The `TestRunner` calls `if(auto err = gen_->buildOneString(generatedDocs, **corpus))`, which calls `Generator::buildOneString`, where you changed the call to `buildOne` to `buildOne(ss, corpus, "");`. We _could_ fix CI by just adapting `AdocGenerator::buildOne` to not generate the tagfile when this string is empty. However, this doesn't address the deeper issue here.

The real issue causing all of that is the API design. Your new `fileName` parameter in `Generator::buildOne` is breaking the contract of this abstract class. 

1) First of all, the new parameter is documented as `@param fileName The file name to use for the output.`, but that's not what the implementation is doing. The filename is the filename for the tagfile, which is unrelated to the filename of the output. The parameter isn't even used in the `XMLGenerator`. It's even more confusing because `fileName` is not the filename of the tagfile but the basename to which "tag.xml" will be appended to form the tagfile filename.
2) Most importantly, the contract of `Generator::buildOne` is that it renders the documentation to any output stream regardless of the filename because (i) there's not even a filename for most output types and (ii) if there were a file for each stream, the stream would become redundant. For instance, tests use `buildOneString`, which outputs the result to a string. 

This bug was already there, but CI only made it evident. It worked before rebasing because `Generator::buildOneString` wasn't being covered with the `AdocGenerator`.

Because of how `Generator` expects the output to be decoupled from any filename and `buildOne` is based on streaming only that result to the output, the `Generator` should have received new virtual functions to `buildTagfile(std::ostream& os, Corpus const& corpus)` or something instead of `buildOne` trying to fill this role. 

If this is hard to achieve because we can't decouple the logic of `buildOne` and `buildTagfile`, a second-best would be `buildOne(std::ostream& os, Corpus const& corpus, std::ostream& tagFileOs)` instead of `buildOne(std::ostream& os, Corpus const& corpus, std::string_view fileName)` and a new overload in `Generator` without `std::ostream& tagFileOs` which would call `buildOne` with a noop `std::ostream` for the `tagFileOs`.

This fixes the underlying problem and the original issue with `buildOneString` that's breaking CI.

fpelliccioni:
> The implementation of the tagfile write is excellent.
> 
> CI failed after I rebased. That was a good thing because it helped me catch a bug.
> 
> So, the reason CI is failing is trivial. The `TestRunner` calls `if(auto err = gen_->buildOneString(generatedDocs, **corpus))`, which calls `Generator::buildOneString`, where you changed the call to `buildOne` to `buildOne(ss, corpus, "");`. We _could_ fix CI by just adapting `AdocGenerator::buildOne` to not generate the tagfile when this string is empty. However, this doesn't address the deeper issue here.
> 
> The real issue causing all of that is the API design. Your new `fileName` parameter in `Generator::buildOne` is breaking the contract of this abstract class.
> 
> 1. First of all, the new parameter is documented as `@param fileName The file name to use for the output.`, but that's not what the implementation is doing. The filename is the filename for the tagfile, which is unrelated to the filename of the output. The parameter isn't even used in the `XMLGenerator`. It's even more confusing because `fileName` is not the filename of the tagfile but the basename to which "tag.xml" will be appended to form the tagfile filename.
> 2. Most importantly, the contract of `Generator::buildOne` is that it renders the documentation to any output stream regardless of the filename because (i) there's not even a filename for most output types and (ii) if there were a file for each stream, the stream would become redundant. For instance, tests use `buildOneString`, which outputs the result to a string.
> 
> This bug was already there, but CI only made it evident. It worked before rebasing because `Generator::buildOneString` wasn't being covered with the `AdocGenerator`.
> 
> Because of how `Generator` expects the output to be decoupled from any filename and `buildOne` is based on streaming only that result to the output, the `Generator` should have received new virtual functions to `buildTagfile(std::ostream& os, Corpus const& corpus)` or something instead of `buildOne` trying to fill this role.
> 
> If this is hard to achieve because we can't decouple the logic of `buildOne` and `buildTagfile`, a second-best would be `buildOne(std::ostream& os, Corpus const& corpus, std::ostream& tagFileOs)` instead of `buildOne(std::ostream& os, Corpus const& corpus, std::string_view fileName)` and a new overload in `Generator` without `std::ostream& tagFileOs` which would call `buildOne` with a noop `std::ostream` for the `tagFileOs`.
> 
> This fixes the underlying problem and the original issue with `buildOneString` that's breaking CI.

This issue is addressed. I’ll let you know once CI passes so you can review, and if it’s good to go, feel free to merge.

fpelliccioni:
@alandefreitas done

include/mrdocs/Generator.hpp
```diff
@@ -100,6 +108,33 @@ class MRDOCS_VISIBLE
...
+
+        @param context The context to use for the output.
+    */
+    MRDOCS_DECL
+    virtual
+    Error
+    buildOne(
+        std::ostream& os,
+        Corpus const& corpus,
+        GenerationContext const& context) const = 0;
```
alandefreitas, line 135: My impression is that this solution expands the problem rather than solves it. Now, not only we expect a `std::string tagFileName;` (which breaks the contract of the class) but we expect a redundant `std::ostream& tagFileOs;` that needs to be handled and maintained.

I don't know if you're familiar with SOLID, but maybe have a look at https://medium.com/@oleksandra_shershen/solid-principles-implementation-and-examples-in-c-99f0d7e3e868. 

The original problem is that expecting `std::string tagFileName` as a parameter breaks the contract of the parent class because the single responsibility of `buildOne` should be streaming the documentation results to any output. In other words, ideally, the parent `Generator` class shouldn't mention anything related to tagfiles because being able to generate tagfiles is not a property of all generators. If only some generators have this functionality, ideally, there should be an intermediary class (say `GeneratorThatAlsoGeneratesTagfiles` to avoid bikeshedding for now) between `Generator` and the generators that so have tagfiles functionality (`AdocGenerator` and `HTMLGenerator`). "`GeneratorThatAlsoGeneratesTagfiles`" would inherit from `Generator` and define the common API for generating tagfiles (which might have many similarities for all generators). `AdocGenerator` and `HTMLGenerator` would then inherit from "`GeneratorThatAlsoGeneratesTagfiles`" and only implement whatever is different for Adoc and HTML, which I don't expect to be much because only the file extension of the filenames should be different.

The solution I proposed is a `buildTagfile(std::ostream& os, Corpus const& corpus)`, ideally in this new intermediary base class with the common logic for tagfiles. If we really don't want an intermediary class for "convenience", then this `buildTagfile(std::ostream& os, Corpus const& corpus)` function could go directly in the generator and do nothing when the Generator cannot generate tagfiles.

If the logic for tag files can't be decoupled from the logic for `buildOne`, maybe we could consider more complex solutions because it's impossible to implement its single responsibility. But it's clear that tagfiles can be implemented separately because they're basically an XML generator with a different schema.

src/lib/Gen/adoc/AdocGenerator.cpp
```diff
@@ -69,10 +76,35 @@ build(
     if(! ex)
         return ex.error();
 
-    MultiPageVisitor visitor(*ex, outputPath, corpus);
+    std::string path = files::appendPath(outputPath, "reference.tag.xml");
```
alandefreitas, line 79: We need an option here, such as:

https://www.doxygen.nl/manual/config.html#cfg_generate_tagfile

Its default value can be `"reference.tag.xml"` or whatever and tagfiles won't be generated when it's empty.

src/lib/Gen/adoc/AdocGenerator.cpp
```diff
@@ -69,10 +76,35 @@ build(
...
+                std::ios_base::trunc // | std::ios_base::noreplace
+            );
+    }
+    catch(std::exception const& ex)
+    {
+        return formatError("std::ofstream(\"{}\") threw \"{}\"", path, ex.what());
+    }    
+
+    RawOstream raw_os(os);
+    auto tagfileWriter = TagfileWriter(raw_os, corpus);
```
alandefreitas, line 100: Completely independent behavior. It's not even using the `domCorpus` that includes the reference file names. It shouldn't be made dependent on `build` for "convenience".

include/mrdocs/Corpus.hpp
```diff
@@ -160,6 +160,39 @@ class MRDOCS_VISIBLE
...
+        argument, followed by `args...`.    
+
+        @param I The Info to traverse.
+        @param pred The predicate to use to determine if the member should be visited.
+        @param f The function to invoke with the member as the first argument, followed by `args...`.
+        @param args The arguments to pass to the function.
+    */
+    template <InfoParent T, class Pred, class F, class... Args>
+    void
+    traverseIf(
```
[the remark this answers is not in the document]
fpelliccioni: I gave it a try, but the resulting code actually ended up more complex than keeping this new function. Believe me! 😅

src/lib/Support/Generator.cpp
```diff
@@ -114,5 +101,37 @@ buildOneString(
     }
 }
 
+Error
+Generator::
+buildOne(
+    std::ostream& os,
+    Corpus const& corpus) const
+{
+    return Error("buildOne() not implemented for this generator type");
```
alandefreitas, line 110: All generators implement this function. It's the most basic function they have. There's no point to make not abstract just to return a runtime error saying it's not implemented. That's the exact problem the abstract version solves in compile time.

include/mrdocs/Generator.hpp
```diff
@@ -157,6 +157,23 @@ class MRDOCS_VISIBLE
...
+        the filename and adds the generator's specific extension.
+
+        @param outputPath The original path, which can be a directory or a filename.
+
+        @return A string representing the complete path to the output file with 
+        the appropriate extension.
+    */
+    MRDOCS_DECL
+    std::string
+    prepareOutputPath(
```
alandefreitas, line 175: There's no reason for this to be a member function. It needs no members of the `Generator`, and not all generators will need it. 

If we need to reuse this logic, we can put that in Path.hpp with a name that makes sense (`prepare` is not informative), but I suspect we already have functions there to implement this logic with very few lines.

include/mrdocs/GeneratorWithTagfile.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include <string_view>
+
+namespace clang {
+namespace mrdocs {
+
+/** Base class for generators that also create tagfiles.
+    This class provides an additional API to handle
+    the generation of tagfiles alongside the main documentation output.
+*/
+class GeneratorWithTagfile
```
alandefreitas, line 33: Better than before, which was breaking the API. Recently, we merged all the code from `adoc` and `html` because the only difference between them was the file extensions. Now that we have simplified the generators into HandlebarsGenerator, this can be even simpler because these latest changes made this class redundant. 

```cpp
build(std::string_view outputPath,Corpus const& corpus)
```

is the function we use from `GenerateAction`. And it's already part of the contract that it might generate multiple files (because multi-page). 

`buildOne` is what we use for tests, and it can't generate multiple outputs.

So now the `generateTagfile(...)` logic can go in an anonymous namespace in `HandlebarsGenerator.cpp` (so no need to have another Generator class and not breaking the contract of `build` for the handlebars generator), and the logic of `HandlebarsGenerator::build` only needs to become:

```
     if (!corpus.config->multipage)
    {
        Error e = Generator::build(outputPath, corpus);
        if (!e.failed() && !corpus.config->tagfilePath.empty())
        {
            return generateTagfile(...);
        }
    }
// ...
    auto errors = ex->wait();
    if (!errors.empty())
    {
        return Error(errors);
    }
    if (!corpus.config->tagfilePath.empty())
    {
          return generateTagfile(...);
    }
    return Error::success();
```

This anon function `generateTagfile` can take whatever you need, which is probably just the dom corpus and the file extension.

src/lib/Gen/adoc/AdocGenerator.cpp
```diff
@@ -105,9 +156,22 @@ buildOne(
...
+        if (auto err = files::createDirectory(files::getParentDir(tagFilePath))) 
+            return err;
+    }
+
+    auto tagFileWriterEx = createTagfileWriter(corpus, tagFilePath);
+    if(! tagFileWriterEx)
+        return tagFileWriterEx.error();
+
+    auto const justFileName = path::filename(anchorFileName);
+    SinglePageVisitor visitor(*ex, corpus, os, justFileName, std::move(*tagFileWriterEx));
```
alandefreitas, line 171: This unnecessary coupling between these visitors and the tagfile writer seems to be what's causing much of the problem because now you want to propagate these properties up to the API. But there's not needed there. `build` already has all the information it needs about the tagfile path from the corpus. 

Then it's easier to check if the tagfilePath is empty or not and run that step or not. If we even need a visitor for that (the visitor operates in the corpus but the domCorpus has the refs), the tagfile visitor could be much simpler because it would have the exact same logic for all nodes. 

In fact, if anything is wrong about `build` is that it should be simpler: not more complex. Because it also already has the information about the output path from the config options. In fact, I'm going to implement this right now in my branch.

include/mrdocs/Generator.hpp
fpelliccioni: I'll rename the function to `getOutputFullPath` to better reflect its purpose. Regarding making it a free function or moving it to Path, I believe it doesn't fit there. While not all generators use it, it's already used in `AdocGenerator`, `HtmlGenerator`, and the base `Generator`, making it closely tied to generator-specific logic.

As for the dependency on `fileExtension()`, I could make it a parameter to decouple it further, but since it's tied to the generator's behavior, I think keeping it as a member function maintains clarity.

src/lib/Gen/adoc/AdocGenerator.cpp
fpelliccioni: I understand your point about simplifying the `build` function and reducing the coupling between the visitors and the `TagfileWriter`. However, there's a key detail I’d like to clarify: the visitors manage specific information that the `TagfileWriter` relies on for proper tagfile generation.

- **MultiPageVisitor** determines the file names of the generated documentation files and passes them to the `TagfileWriter`.
- **SinglePageVisitor** provides unique data, such as the total number of pages, which the `TagfileWriter` uses for its logic.

This creates a necessary dependency between the visitors and the `TagfileWriter`.

Do you have any suggestions for how the `TagfileWriter` could receive this data without being directly coupled to the visitors?

alandefreitas: > MultiPageVisitor determines the file names of the generated documentation

That comes from the Dom. Not from the visitor.

A visitor has access to the build, which has access to the Dom, which has access to the xrefs (The code literally says `builder.domCorpus.getXref(I)`). 

The visitor has no information the generator doesn't. That's the information it has:

```
ExecutorGroup<Builder>& ex_;
std::string_view outputPath_;
Corpus const& corpus_;
```

The generator has direct access to the Dom without the Visitor or the Builder. A separate TagfileVisitor can take everything from the Dom directly without duplicated logic for single-page and multi-page visitors.

> SinglePageVisitor provides unique data, such as the total number of pages

If it's a single page, the number of pages is not very special information. The number of pages is 1. 

For multipage, it's the number of symbols + overloads (which will be simplified to the number of symbols in the future).

In any case, an XML writer should be able to write an array without knowing the number of elements in advance. Arrays are not prefixed by the number of elements in XML.

fpelliccioni: If I recall correctly, we included page numbers in the anchor based on a discussion (possibly on Slack or in a previous PR review). This design made sense because SinglePageVisitor calculates this information, and the TagfileWriter uses it.

If your intention is to discard page number information entirely, the design of TagfileWriter and related code could indeed be simplified. 

Could you clarify if that’s the direction you’re suggesting?

alandefreitas: What is a page number in the anchor? I'm not following. Do you have an example in a tagfile where the page number is included?

I don't see page numbers in any of these files: https://github.com/boostorg/url/tree/develop/doc/tagfiles

I don't know where the page numbers would go or what that would mean because there is no order relationship between the pages in MrDocs.

All I see in the SinglePageVisitor is `page = numPages_`, which is the index of the current page so that pages are in order in the final output regardless of the process being asynchronous. But I don't see it being used in the output.

src/lib/Gen/hbs/HandlebarsGenerator.cpp
```diff
@@ -15,13 +15,73 @@
...
+#include <fstream>
 #include <sstream>
 
 namespace clang {
 namespace mrdocs {
 namespace hbs {
 
+namespace {
+
+/** Generate a Tagfile for cross-referencing. */
```
alandefreitas, line 37: Minor grammar issue: "cross-referencing" needs a subject.

alandefreitas: What should `generatedFilesPath` represent in the single and multi-page cases? (Just the output path, if I understand correctly).

And why does the generator need this information? It doesn't need to know where these absolute files are because tag files only include relative paths available from the hbsCorpus.

src/lib/Gen/hbs/HandlebarsGenerator.cpp
```diff
@@ -15,13 +15,73 @@
...
+Error
+generateTagfile(
+    HandlebarsCorpus const& hbsCorpus,
+    std::string_view tagFilePath,
+    std::string_view generatedFilesPath,
+    bool isSinglePage)
```
alandefreitas, line 43: One common trick in SOLID is to avoid `bool` such as `isWhatever` because they don't represent what they do via types. 

When there are many parameters with categorical options, the usual solution is to create an enum for that. 

But when there's a single parameter, the usual solution is to rename `generateTagFile` to `generateTagFileImpl` and define `generateSinglePageTagFile` and `generateMultiPageTagFile` as short inline functions that call `generateTagFileImpl` with true and false to make the calling code more legible.

src/lib/Gen/hbs/HandlebarsGenerator.cpp
```diff
@@ -15,13 +15,73 @@
...
+{
+    if (tagFilePath.empty())
+        return Error::success();
```
alandefreitas, line 46: That's not a success :)

The caller should ensure the tagfile is not empty, so it's a precondition: `MRDOCS_ASSERT(!tagFilePath.empty());`

src/lib/Gen/hbs/HandlebarsGenerator.cpp
```diff
@@ -15,13 +15,73 @@
...
+        return tagFileWriterEx.error();
+
+    auto tagfileWriter = std::move(*tagFileWriterEx);
+    tagfileWriter.initialize();
+    visit(hbsCorpus->globalNamespace(), tagfileWriter);
+    tagfileWriter.finalize();
+    return Error::success();
+}
+
+std::string
```
alandefreitas, line 59: It would be good to have a comment explaining what `getSinglePageFileName` is intended to do. At least I don't get it anyway.

In part I don't get it because I still don't understand the necessity of `generatedFilesPath` above (why would it need that if all symbols only have anchors).

But putting that aside, if the tagfile writer needs the directory where the files are, then we just need `getParentDir(outputPath);`.

src/lib/Gen/hbs/HandlebarsGenerator.cpp
```diff
@@ -15,13 +15,73 @@
...
+
+    using SmallString = llvm::SmallString<0>;
+
+    SmallString ext(".");
+    ext += extension;
+
+    SmallString fileName(outputPath);
+    if(path::extension(outputPath).compare_insensitive(ext) != 0)
+    {
+        // directory specified
```
alandefreitas, line 74: So, if the outputPath doesn't end in `.html` or `.adoc` (let's say "x.jpg"), do we consider this a directory?

And then we append "reference.html" or "reference.adoc" to it? What for?

Are you duplicating the code from BuildOne here? More duplicated code is everything we don't need. In the future, we'll never remember to replicate the changes in multiple places.

We can create a free function in Generator and call it from both places. We can also improve the logic of the free function (Vinnie wrote that code in Feb 2023 in a hurry) to check if the outputPath is not a directory and adjust it accordingly.

As for the new logic, we have kind of a conflict here. If output path is already a directory (`isDirectory(outputPath)`) then things are clear. If the output path doesn't exist yet, then we don't know if the user intended a directory where "reference.adoc/html" should be appended or just a file without an extension. 

If the directory doesn't exist yet, we could assume anything with a dot is a file and anything without a dot should be a directory. Or we could return an error. All I know is the current heuristic (the filename matching the extension of the generator) is not good because extensions change and if the output path is "index.htm", we know the user doesn't want "index.htm/reference.html".

src/lib/Gen/hbs/HandlebarsGenerator.cpp
```diff
@@ -70,7 +130,16 @@ build(
 {
     if (!corpus.config->multipage)
     {
-        return Generator::build(outputPath, corpus);
+        Error e = Generator::build(outputPath, corpus);
+        if (!e.failed() && !corpus.config->tagfilePath.empty())
```
alandefreitas, line 134: Great :)

src/lib/Gen/hbs/HandlebarsGenerator.cpp
```diff
@@ -84,7 +153,15 @@ build(
 
     // Wait for all executors to finish and check errors
     auto errors = ex->wait();
-    MRDOCS_CHECK_OR(errors.empty(), errors);
+    if (!errors.empty())
+    {
+        return Error(errors);
+    }
```
alandefreitas, line 159: MRDOCS_CHECK_OR resolves to the exact same thing.

src/lib/Lib/TagfileWriter.hpp
```diff
@@ -0,0 +1,91 @@
...
+namespace clang {
+namespace mrdocs {
+
+struct SimpleWriterTag {}; //Tag dispatch for simple writers
+
+class jit_indenter;
+
+/** A writer which outputs Tagfiles.
+*/
+class TagfileWriter
```
alandefreitas, line 32: Many functions could be private.

src/lib/Lib/TagfileWriter.cpp
```diff
@@ -0,0 +1,475 @@
...
+{
+    // Do nothing
+}
+
+void
+TagfileWriter::
+writeVariable(
+    VariableInfo const& I)
+{
+    // Do nothing
```
alandefreitas, line 431: writeClassLike?

alandefreitas:
I downloaded the demo artifacts and saw no tag files in there.

src/lib/Gen/hbs/HandlebarsGenerator.cpp
fpelliccioni: The parameter `generatedFilesPath` was renamed to `outputFilesBasePath` to clarify its purpose.

In "multipage" mode, the anchors for symbols come directly from the Corpus, which includes the generated filenames. However, in "singlepage" mode, the anchor is the resulting document (e.g., `reference.adoc`). Since this information is not present in the Corpus, we need to generate it explicitly.

The function `getSinglePageFileName` exists precisely for this reason: to determine the output file name (e.g., .adoc) for use as the anchor in "singlepage" mode.

fpelliccioni:
> I downloaded the demo artifacts and saw no tag files in there.

@alandefreitas 

You need to specify the `tagfile-path` option. By default, it's empty, and no tag file will be generated unless explicitly set. 

Currently, `tagfile-path` is defined as `"relativeto": "<config-dir>"` in the configuration. 
I'm wondering if this is the best approach or if it should require an absolute path instead.

alandefreitas:
> By default, it's empty

We can change that default. Or at least enable it in CI. This can't go untested. We need these file artifacts in CI.

> if it should require an absolute path instead

We never require an absolute path. The tagfile is relative to the output path. Not the config path.

fpelliccioni:
> > By default, it's empty
> 
> We can change that default. Or at least enable it in CI. This can't go untested. We need these file artifacts in CI.
> 
> > if it should require an absolute path instead
> 
> We never require an absolute path. The tagfile is relative to the output path. Not the config path.

I didn’t see a way to make the tagfile path relative to the output path directly. I believe I’ll need to update the code handling the configuration options to support this behavior (though I’m not entirely sure yet).

I’ll go ahead and make the necessary changes, but I’ll address this in a separate PR.

alandefreitas:
Whatever you put in "relativeto" will go to an `OptionProperties::relativeto` string in the public settings. This function is called for all arguments so you can normalize them:

https://github.com/cppalliance/mrdocs/blob/develop/src/lib/Lib/Config.cpp#L75

When the argument is a relative path, it gets the base path for normalization here:

https://github.com/cppalliance/mrdocs/blob/develop/src/lib/Lib/Config.cpp#L95

The base path can come from "relativeto" or from the default path prefix. There's no default prefix because the string is empty by default. When it comes from "relative-to", we get it with this function call:

https://github.com/cppalliance/mrdocs/blob/6156756ac462ebfdaec2d04f17c356eb4c987894/src/lib/Lib/Config.cpp#L229

The function only considers the Reference directories for now:

https://github.com/cppalliance/mrdocs/blob/6156756ac462ebfdaec2d04f17c356eb4c987894/src/lib/Lib/Config.cpp#L192

It doesn't currently consider other options as valid prefixes. It would probably need to change to consider the `PublicSettings& self`. 

One thing to consider is that the output path will need to be normalized before the tagfiles path. In practice, it means the output needs to come before the tagfile option in the json file. In principle, the python script could identify dependencies between options to rearrange them but I don't think it's something we need for now.

fpelliccioni:
> Whatever you put in "relativeto" will go to an `OptionProperties::relativeto` string in the public settings. This function is called for all arguments so you can normalize them:
> 
> https://github.com/cppalliance/mrdocs/blob/develop/src/lib/Lib/Config.cpp#L75
> 
> When the argument is a relative path, it gets the base path for normalization here:
> 
> https://github.com/cppalliance/mrdocs/blob/develop/src/lib/Lib/Config.cpp#L95
> 
> The base path can come from "relativeto" or from the default path prefix. There's no default prefix because the string is empty by default. When it comes from "relative-to", we get it with this function call:
> 
> https://github.com/cppalliance/mrdocs/blob/6156756ac462ebfdaec2d04f17c356eb4c987894/src/lib/Lib/Config.cpp#L229
> 
> The function only considers the Reference directories for now:
> 
> https://github.com/cppalliance/mrdocs/blob/6156756ac462ebfdaec2d04f17c356eb4c987894/src/lib/Lib/Config.cpp#L192
> 
> It doesn't currently consider other options as valid prefixes. It would probably need to change to consider the `PublicSettings& self`.
> 
> One thing to consider is that the output path will need to be normalized before the tagfiles path. In practice, it means the output needs to come before the tagfile option in the json file. In principle, the python script could identify dependencies between options to rearrange them but I don't think it's something we need for now.


PR created https://github.com/cppalliance/mrdocs/pull/743

PR #682, opened 2024-10-01
fix: infinite recursion when identifying SFINAE types

### LINKED ISSUE #665: Infinite recursion while detecting SFINAE
I was trying to compile MrDocs and ran into a stack-overflow while building the documentation for MrDocs itself (which is build by default).

The error occurs when trying to detect whether [`fmt::formatter`](https://github.com/fmtlib/fmt/blob/10.2.1/include/fmt/core.h#L1086-L1090) is such a template. One of its specializations is in [`mrdocs/Dom/String.hpp`](https://github.com/cppalliance/mrdocs/blob/bd54dc7c2562ec751e42fb161116468a4838cb6d/include/mrdocs/Dom/String.hpp#L344-L355). In this specialization, the `fmt::formatter` extends its own template. When checking for SFINAE, the base template is checked as well, leading to infinite recursion.

The following change "fixes" the issue, but I don't know if it's correct to just "ignore" that base class.
```diff
diff --git a/src/lib/AST/ASTVisitor.cpp b/src/lib/AST/ASTVisitor.cpp
index 1d48999..893997e 100644
--- a/src/lib/AST/ASTVisitor.cpp
+++ b/src/lib/AST/ASTVisitor.cpp
@@ -1723,6 +1723,8 @@ public:
                             return true;
                         continue;
                     }
+                    if (sfinae_info->Template == TD)
+                        continue;
                     auto sfinae_result = isSFINAETemplate(
                         sfinae_info->Template, Member);
                     if(! sfinae_result)
```

Minimal example:

```cpp
template <typename T> struct Foo {};
template<> struct Foo<char> : public Foo<int> {};
```

This isn't caught in CI, as the docs generation is disabled there.

----

_Aside:_

- The documentation generation from CMake doesn't work right now, because it tries to find a [`local-antora-playbook.yml`](https://github.com/cppalliance/mrdocs/blob/bd54dc7c2562ec751e42fb161116468a4838cb6d/CMakeLists.txt#L439) - this should probably be `antora-playbook.yml`.
- MSVC fails to compile `lib/AST/ASTVisitor.cpp`, because [`CompleteExternalDeclaration`](https://github.com/cppalliance/mrdocs/blob/bd54dc7c2562ec751e42fb161116468a4838cb6d/src/lib/AST/ASTVisitor.cpp#L4681) is overridden with a `DeclaratorDecl*`, but [declared with a `VarDecl*`](https://github.com/llvm/llvm-project/blob/7a28a5b3fee6c78ad59af79a3d03c00db153c49f/clang/include/clang/AST/ASTConsumer.h#L108) in the version referenced on https://mrdocs.com/docs/mrdocs/install.html (7a28a5b3fee6c78ad59af79a3d03c00db153c49f; I suppose that version is outdated)

sdkrystian:
Fixes #665. Note that this fix isn't perfect as it does not handle cases where the recursion period is greater than one.

### DIFF
--- a/src/lib/AST/ASTVisitor.cpp
+++ b/src/lib/AST/ASTVisitor.cpp
@@ -1716,6 +1716,11 @@ public:
                             return true;
                         continue;
                     }
+                    // if the class inherits from itself, we can't determine whether
+                    // it's a SFINAE type
+                    if(declaresSameEntity(TD, sfinae_info->Template))
+                        return true;
+
                     auto sfinae_result = isSFINAETemplate(
                         sfinae_info->Template, Member);
                     if(! sfinae_result)


src/lib/AST/ASTVisitor.cpp
```diff
@@ -1723,6 +1723,11 @@ class ASTVisitor
                             return true;
                         continue;
                     }
+                    // if the class inherits from itself, we can't determine whether
+                    // it's a SFINAE type
+                    if(declaresSameEntity(TD, sfinae_info->Template))
+                      return true;
```
vinniefalco, line 1729: is this the cause of the reported issue?

sdkrystian: @vinniefalco yes

PR #683, opened 2024-10-01
refactor: drop bitcode support

sdkrystian:
In its current form, the bitcode serialization feature of MrDocs significantly increases complexity for no tangible benefit. The size of a serialized `Info` object is _larger_ than its deserialized representation, and a significant amount of serialization/deserialization must be performed in order to merge `Info` across TUs. Moreover, we ultimately discard the serialized representation once all `Info` is extracted and the final corpus is built. 

Since the cost of maintaining the bitcode serialization feature is so high & the benefits non-existent (so far as to negatively impact performance), it would be best to drop the feature for now & perhaps revisit it if the need arises.


sdkrystian:
Merged in [e6205d5](https://github.com/cppalliance/mrdocs/commit/e6205d51f72b21d12732cb141c574543da249b8d)

alandefreitas:
That's amazing.

PR #688, opened 2024-10-03
Docs: Fix Cmake example in usage.adoc

akrzemi1:
Replace `OUTPUT_FILE` with `TMP_CPP_FILE` in the Cmake example for header-only libraries, so that it compiles.


alandefreitas:
> Replace OUTPUT_FILE with TMP_CPP_FILE in the Cmake example for header-only libraries, so that it compiles.

Oh... that came from the template. I thought you had come up with that for your project. :)

PR #701, opened 2024-10-11
feat: generate YAML schema for config

### LINKED ISSUE #700: Provide Config Schema
RedHat's [YAML language server](https://github.com/redhat-developer/yaml-language-server) provides the option to associate a file with some schema by adding a [special top level comment](https://github.com/redhat-developer/yaml-language-server?tab=readme-ov-file#using-inlined-schema) (or through the settings). It would be really helpful is MrDocs would provide a schema for the config. This makes editing a much nicer experience, in my opinion. [Mkdocs-material provides such a schema for example](https://squidfunk.github.io/mkdocs-material/creating-your-site/#configuration).

Given that MrDocs has the configuration in a [machine-readable format already](https://github.com/cppalliance/mrdocs/blob/5e0f628e5b1cdb06a3dd260e0e42069f12733353/src/lib/Lib/ConfigOptions.json), this seems easy to add. I modified `generate-config-info.py` a bit to generate a JSON (and thus YAML) schema:

If this something you'd like to add? If so, I'd open a PR.

Nerixyz:
This PR extends `generate-config-info.py` to generate a YAML/JSON schema for the config file. As explained in the issue, this can be used to provide autocompletion and linting. The "YAML schema" is just a [JSON schema](https://json-schema.org).

Currently, the schema is committed to the repo. I'm not sure if that's desired. The main upside of this is, that users can link to the schema using (raw) GitHub links. This allows linking to tags, but most importantly it allows linking to specific commits. However, there's now duplicate information in the repo. CI ensures the schema is always up-to-date, though.

Alternatively, this could be done as part of the Antora build (probably, I'm not that familiar with it).

Closes #700.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -383,6 +383,9 @@ jobs:
           package-generators: ${{ matrix.mrdocs-package-generators }}
           package-artifact: false
 
+      - name: Check YAML schema
+        run: npx -y -p ajv-cli -- ajv compile -s docs/mrdocs.schema.json
+
       - name: Upload GitHub Release Artifacts
         if: ${{ matrix.is-main && matrix.compiler != 'clang' }}
         uses: actions/upload-artifact@v4
@@ -845,4 +848,3 @@ jobs:
           llvm_dir="/var/www/mrdox.com/llvm+clang"
           chmod 755 ${{ matrix.llvm-archive-filename }}
           scp -o StrictHostKeyChecking=no $(pwd)/${{ matrix.llvm-archive-filename }} ubuntu@dev-websites.cpp.al:$llvm_dir/
-
--- a/docs/modules/ROOT/pages/config-file.adoc
+++ b/docs/modules/ROOT/pages/config-file.adoc
@@ -13,6 +13,19 @@ include::partial$mrdocs-example.yml[]
 The xref:usage.adoc[Usage] page provides a detailed explanation of what to combine options from the configuration file and the command line.
 The <<config-options-reference>> section provides a detailed explanation of the options available.
 
+== YAML Schema
+
+To get linting and autocompletion in the config file, a schema for the config can be specified.
+In JetBrains IDEs, `# $schema: <url>` can be used to bind a schema to a file.
+In editors with plugins based on the https://github.com/redhat-developer/yaml-language-server[YAML language server], `# yaml-language-server: $schema=<url>` can be used.
+The schema for `mrdocs.yml` is provided xref:attachment$mrdocs.schema.json[here].
+The following shows an example of a file specifying an inline-schema that's compatible with JetBrains IDEs and editors using the YAML language server.
+
+[source,yaml]
+--------
+include::partial$mrdocs-schema-example.yml[]
+--------
+
 == Build Options
 
 A number of options can be used to specify with which compile options MrDocs should be run.
--- a/docs/mrdocs.schema.json
+++ b/docs/mrdocs.schema.json
@@ -1,0 +1,252 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "properties": {
+    "addons": {
+      "default": "<mrdocs-root>/share/mrdocs/addons",
+      "description": "Path to the Addons directory. The Addons directory contains the template files used by generators to create the documentation. When a custom Addons directory is not specified, the default templates are used. The default templates are located at the `share/mrdocs/addons` directory of the MrDocs installation. Users can create custom templates by copying the default templates to a custom directory and specifying the custom directory using this option.",
+      "title": "Path to the Addons directory",
+      "type": "string"
+    },
+    "anonymous-namespaces": {
+      "default": "always",
+      "description": "Determine whether symbols in anonymous namespaces should be extracted. When set to `always`, symbols in anonymous namespaces are always extracted. When set to `dependency`, symbols in anonymous namespaces are extracted only if they are referenced by the source code. When set to `never`, symbols in anonymous namespaces are never extracted.",
+      "enum": [
+        "always",
+        "dependency",
+        "never"
+      ],
+      "title": "Extraction policy for anonymous namespaces"
+    },
+    "base-url": {
+      "default": "",
+      "description": "Base URL for links to source code. The base URL is used to create links to the source code in the documentation. The base URL is combined with the path to the source file to create the link.",
+      "title": "Base URL for links to source code",
+      "type": "string"
+    },
+    "cmake": {
+      "default": "",
+      "description": "When the compilation-database option is a CMakeLists.txt file, these arguments are passed to the cmake command to generate the compilation_database.json.",
+      "title": "CMake arguments when generating the compilation database from CMakeLists.txt",
+      "type": "string"
+    },
+    "compilation-database": {
+      "default": "",
+      "description": "Path to the compilation database or a build script to generate it. The compilation database is a JSON file that contains the compiler commands used to build the source code. The compilation database is used to extract the compiler flags and the source files used to build the source code and extract symbols. This option also accepts the path to a build script such as CMakeLists.txt to be used to generate the compilation database. In this case, MrDocs will look for CMake in `PATH` or in `CMAKE_ROOT` and run the script to generate the compilation database file.",
+      "title": "Path to the compilation database",
+      "type": "string"
+    },
+    "defines": {
+      "default": [],
+      "description": "Additional defines passed to the compiler when building the source code. These defines are added to the compilation database regardless of the strategy to generate it.",
+      "items": {
+        "type": "string"
+      },
+      "title": "Additional defines passed to the compiler",
+      "type": "array"
+    },
+    "detect-sfinae": {
+      "default": true,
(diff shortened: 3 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -383,6 +383,10 @@ jobs:
           package-generators: ${{ matrix.mrdocs-package-generators }}
           package-artifact: false
 
+      # If this fails, the config was updated without updating the schema
+      - name: Check YAML schema
+        run: git --no-pager diff --exit-code -- docs/mrdocs.schema.json
```
alandefreitas, line 388: I'm sorry for my ignorance. How will this command know if the config JSON file changed? 

As far as I understand, this command checks for changes only in the `docs/mrdocs.schema.json` file and compares it to the last committed version. But this file not changing is not an error by itself.

Also, if I understand correctly, `docs/mrdocs.schema.json` only contains a subset of the config file so not all changes in the config file require changes in `docs/mrdocs.schema.json`. For instance, if some field changes from `string` to `path`, `docs/mrdocs.schema.json` will not change.

docs/modules/ROOT/pages/config-file.adoc
```diff
@@ -12,6 +12,8 @@ include::partial$mrdocs-example.yml[]
 
 The xref:usage.adoc[Usage] page provides a detailed explanation of what to combine options from the configuration file and the command line.
 The <<config-options-reference>> section provides a detailed explanation of the options available.
+To get linting and autocompletion in the config file, the https://github.com/redhat-developer/yaml-language-server[YAML language server] can be used.
```
alandefreitas, line 15: This deserves a section including where the file can be found, the specification, etc... 

If I understand correctly, as the file follows that json spec, many IDEs support this format even without the user having to know about anything.

The most important/actionable information is that the user needs to put that comment at the top of the file.

alandefreitas (commented):
That's very interesting. I'll research how IDEs integrate that and include it in the documentation. Is this a standard format for validating the fields, or are there competing formats? It's the same format as the schema for vcpkg manifest files, right?

> Currently, the schema is committed to the repo. I'm not sure if that's desired. The main upside of this is, that users can link to the schema using (raw) GitHub links. This allows linking to tags, but most importantly it allows linking to specific commits. However, there's now duplicate information in the repo. CI ensures the schema is always up-to-date, though.

Yes, that's the first thing that came to mind. And I'm still trying to figure it out. If we follow the pattern of other artifacts from `generate-config-info.py`, then it wouldn't be checked in. It could be added as an extra file in the releases.

However, we ensure these other artifacts from generate-config-info.py are always correct because the code that uses them still has to compile and pass the tests. We can't guarantee the same for `mrdocs.schema.json`. Is there a way to check that the resulting file is at least valid? Some tool or something?

Or should the original config.json file follow the specification directly at some point in the future? Is there a way to associate categories with options in the specification?

Unless there's evidence a specific design is much better, we can keep the file directly in the repository unless there's a lot of evidence this is a bad idea.

> Alternatively, this could be done as part of the Antora build (probably, I'm not that familiar with it).

Yes. I would rule that out.

.github/workflows/ci.yml
Nerixyz: The idea is that `generate-config-info.py` runs during the build and updates `docs/mrdocs.schema.json`. Then, after the build, CI checks if the file was updated (i.e. the schema changed, but wasn't committed).

Nerixyz:
> Is this a standard format for validating the fields, or are there competing formats?

JSON-schema is the de-facto standard for many config-languages (i.e. not just JSON), since it's so versatile. In addition to YAML, it's also used for TOML (e.g. in [taplo](https://taplo.tamasfe.dev/) - [schema for Cargo.toml](https://json.schemastore.org/cargo.json)). 

> It's the same format as the schema for vcpkg manifest files, right?

Yes, in JSON the standard is to specify a `"$schema"`. At least on [Microsoft's documentation](https://learn.microsoft.com/en-us/vcpkg/reference/vcpkg-json), they guide users to use the schema from GitHub.

> We can't guarantee the same for `mrdocs.schema.json`. Is there a way to check that the resulting file is at least valid? Some tool or something?

That's mostly what I'm trying to do with the check in CI. An alternative could be to have a separate script, run that one in CI, and check if the output changed (with Git).

> Or should the original config.json file follow the specification directly at some point in the future? Is there a way to associate categories with options in the specification?

Unfortunately, there's no way of having categories for options in a JSON schema. Furthermore, the current config.json has a few options that specify "command-line-only".

.github/workflows/ci.yml
alandefreitas: Oh. OK. So, it should return 0 even if the contents are the same.

In CI, we're also having an issue with checkout not constantly downloading the source files from the repository. This is fixable by setting up git first. 

Or we could remove this test since it does not test much—only that the file has been touched.

docs/mrdocs.schema.json
```diff
@@ -0,0 +1,255 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema",
```
alandefreitas, line 2: https://json-schema.org/draft/2020-12/schema ?

alandefreitas:
> That's mostly what I'm trying to do with the check-in CI. An alternative could be to have a separate script, run that one in CI, and check if the output changed (with Git).

I see. Unless there's a tool to check if the file is valid, checking if the file has been touched or changed does not provide much additional value. We could remove the test for now.

I considered using the meta schema to check if the resulting file matches it. But we can go with what we have for now. Whoever is pushing a new commit should check if the file changed. I still believe it's possible to use the [metaschema](https://json-schema.org/draft/2020-12/schema) of the [specification](https://json-schema.org/draft/2020-12/json-schema-core). 

I'll at least create an issue. I saw [this list of tools](https://json-schema.org/tools?query=&sortBy=name&sortOrder=ascending&groupBy=toolingTypes&licenses=&languages=&drafts=&toolingTypes=&environments=#validator), but I haven't investigated it yet.

> Unfortunately, there's no way of having categories for options in a JSON schema. Furthermore, the current config.json has a few options that specify "command-line-only".

I see. Some fields could contain extra information, so using the same file for both things could reduce redundancy and implicitly ensure we're testing the file.

.github/workflows/ci.yml
Nerixyz: > In CI, we're also having an issue with checkout not constantly downloading the source files from the repository. This is fixable by setting up git first.
> 
> Or we could remove this test since it does not test much—only that the file has been touched.

Then it's easier to have a separate script, and remove the test from CI, I think.

docs/mrdocs.schema.json
Nerixyz: Yea, we could update to the new draft. Most schemas I saw were using older drafts, but I don't think that should be a problem.

.github/workflows/ci.yml
alandefreitas: Mmm... Yes. We don't strictly need a separate script, but that's the best idea (Unless the logic for the schema reuses a lot of the logic for generating the C++ files but I doubt that.) because the CMake scripts would be generating a file we don't need to compile the library. 

I'm looking at https://www.schemastore.org/json/, and they suggest a [MegaLinter](https://megalinter.github.io/) action to check files. It could replace the whole step. The only problem is this MegaLinter is way more than we need.

Nerixyz: > I'm looking at [schemastore.org/json](https://www.schemastore.org/json/), and they suggest a [MegaLinter](https://megalinter.github.io/) action to check files. It could replace the whole step. The only problem is this MegaLinter is way more than we need.

[ajv](https://ajv.js.org/guide/environments.html#command-line-interface) can validate YAML files. We use it [at Chatterino to validate JSON files in CI](https://github.com/Chatterino/chatterino2/blob/c0a5a3e8058255376494408730c9948aae8f66af/.github/workflows/lint.yml#L30-L33). Here, the command-line would be

```sh
npx ajv validate -s mrdocs.schema.json -d mrdocs.yml
```

Through this, it also validates the schema. Seems it doesn't like `https://` in the schema, but expects `http://`.

alandefreitas: Yes. This could make a better GHA step here. Using npx makes it all much more manageable.

However, we want to validate mrdocs.schema.json against the meta schema https://json-schema.org/draft/2020-12/schema.

Nerixyz:
I've updated the branch to make sure we're talking about the same thing.

- Only the draft-07 is supported by RedHat's LSP (https://redirect.github.com/redhat-developer/yaml-language-server/issues/780)
- I found that JetBrains IDEs [support YAML schemas as well](https://www.jetbrains.com/help/idea/yaml.html#use-schema-keyword). They use a different syntax that has also been proposed for RedHat's LSP (https://redirect.github.com/redhat-developer/yaml-language-server/issues/959). I opened a PR to support this (https://redirect.github.com/redhat-developer/yaml-language-server/pull/992) but I don't think it will be merged in the near future.
- The schema is generated from `util/generate-yaml-schema.py` (can run in any directory).

alandefreitas:
> I've updated the branch to make sure we're talking about the same thing.

Nice. Thanks. All issues are solved then. Let's see if it passes CI, and we'll be good to go. We can open issues for whatever smaller issues we find or want to investigate.

Issue #528, opened 2024-01-03
Include source file information for warnings or errors


anarthal:
I'm currently getting this:

```
Failed to resolve reference to 'as_uint64' from 'as_uint64'
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'async_start_statement_execution(const' from 'async_start_statement_execution'
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'rows::operator==(const' from 'operator=='
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'get_datetime' from 'get_datetime'
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'async_execute_statement' from 'async_execute_statement'
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'get_uint64' from 'get_uint64'
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'as_blob' from 'as_blob'
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'field::operator' from 'operator='
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'field::operator' from 'operator='
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'operator=(const' from 'operator='
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'operator=(unsigned' from 'operator='
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'operator=(signed' from 'operator='
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'operator=(unsigned' from 'operator='
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'row::operator!=(const' from 'operator!='
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'async_execute' from 'async_execute'
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'async_ping' from 'async_ping'
    Reported at src/lib/Metadata/Finalize.cpp(173)
Failed to resolve reference to 'rows::operator==(const' from 'operator=='
[...]

PR #704, opened 2024-10-15
feat: source location included in warnings for unresolved references

fpelliccioni:
fixes https://github.com/cppalliance/mrdocs/issues/528

### DIFF
--- a/src/lib/Metadata/Finalize.cpp
+++ b/src/lib/Metadata/Finalize.cpp
@@ -190,17 +190,13 @@ class Finalizer
 
             if constexpr(std::derived_from<NodeTy, doc::Reference>)
             {
-#if 0
                 // This warning shouldn't be triggered if the symbol has
                 // been explicitly marked excluded in mrdocs.yml
                 if(! resolveReference(N))
                 {
-                    report::warn("Failed to resolve reference to '{}' from '{}'",
-                        N.string, current_->Name);
+                    report::warn("Failed to resolve reference to '{}' from '{}' at file '{}' line {} column {}",
+                        N.string, current_->Name, N.fileName, N.line, N.column);
                 }
-#else
-                resolveReference(N);
-#endif
             }
         });
     }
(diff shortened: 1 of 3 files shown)


src/lib/Metadata/Finalize.cpp
```diff
@@ -190,17 +190,13 @@ class Finalizer
...
             if constexpr(std::derived_from<NodeTy, doc::Reference>)
             {
-#if 0
                 // This warning shouldn't be triggered if the symbol has
                 // been explicitly marked excluded in mrdocs.yml
                 if(! resolveReference(N))
                 {
-                    report::warn("Failed to resolve reference to '{}' from '{}'",
-                        N.string, current_->Name);
+                    report::warn("Failed to resolve reference to '{}' from '{}' at file '{}' line {} column {}",
```
alandefreitas, line 197: #528 describes one problem B: the need for source file information in warnings and errors. This is important.

However, the original problem A, which motivated the `#if 0`, is independent: this message is a false positive in the caller's context. 

For instance, if I have these functions:

```cpp
detail::my_impl func_a();

std::optional<int> func_b();
```

Because `detail::my_impl` is an implementation-defined type and `std::optional` is not part of my project, I would get messages like `"Failed to resolve reference to 'detail::my_impl' from 'func_a()' at file 'X' line Y column Z"` or `"Failed to resolve reference to 'std::optional<int>' from 'func_b()' at file 'X' line Y column Z"`. 

Having the line numbers improves the message (problem B). But that doesn't change the fact that the message is wrong (problem A) because these return types are not supposed to be documented, and the user is not obliged not to use undocumented types. So, the user is correct but still receiving a warning.

So, if we solve problem B without solving problem A, it doesn't justify removing `#if 0` from the caller here. The possible solutions would be:

1) Fixing these false negatives so the caller only prints messages when they are correct. The message would not be a false negative if, instead of the return type in code, the reference were an explicit `@ref` in the javadoc that can't be resolved.
2) Leaving the `#if 0` there, leaving the potential improved error messages there, and create another issue for these false negatives.

Also, regarding #528, he got this warning without the source location. However, the issue is about warnings and errors in general, and that's not the only warning MrDocs emits.

Issue #721, opened 2024-10-30
Typedef duplication with virtual bases


anarthal:
[This mrdocs generated page](https://develop.http-proto.cpp.al/http_proto/reference/boost/http_proto/message_base.html) looks like having duplicated entries for the typedefs. This is apparently due to virtual inheritance (as shown by [this diagram](https://develop.http-proto.cpp.al/http_proto/Message.html)).

alandefreitas:
@sdkrystian We still need to ensure we take the right one ("dominance"), and we need tests. So, let's keep this one open.

https://godbolt.org/z/E6GnvzE1a

PR #722, opened 2024-10-30
fix: deduplicate inherited members

sdkrystian:


sdkrystian:
closes #721

src/lib/Metadata/Interface.cpp
```diff
@@ -75,7 +75,8 @@ class TrancheBuilder
         std::vector<SymbolID>& M,
         const Info& I)
     {
-        M.emplace_back(I.id);
+        if(std::ranges::find(M, I.id) == M.end())
```
alandefreitas, line 78: Will this ensure we take the right inherited member, though?

sdkrystian: @alandefreitas All this does is ensure that elements in the tranche are unique.

alandefreitas: So let's at least open another issue

PR #743, opened 2024-11-28
feat: <output> base directory support for relative paths

fpelliccioni:


src/lib/Lib/Config.cpp
```diff
@@ -198,6 +199,12 @@ struct PublicSettingsVisitor {
         else if (referenceDirKey == "mrdocs-root") {
             return dirs.mrdocsRoot;
         }
+        else if (referenceDirKey == "output") {
+            if (settings.output.empty()) {
```
alandefreitas, line 203: `MRDOCS_ASSERT` is a pre-condition the developer needs to fix if it's not met. The user has nothing to do with it.

alandefreitas:
CI failed. `MRDOCS_ASSERT` doesn't support two arguments. Did you run this locally?

fpelliccioni:
> CI failed. `MRDOCS_ASSERT` doesn't support two arguments. Did you run this locally?

the first commit was compiled and tested, including support for this parameter in `ConfigOptions.json` (local change, not pushed):

```
{
  "name": "tagfile-path",
  "type": "file-path",
  "default": "<output>/reference.tag.xml",
  "relativeto": "<output>",
  "must-exist": false
}
```

I sent the second commit without checking it—my apologies for that. I'll fix and re-test it.

PR #771, opened 2024-12-19
feat: navigation button links to enclosing namespace

### LINKED ISSUE #469: Need a button to navigate to the enclosing namespace
It is often useful to navigate directly to the enclosing namespace of a symbol. Right now we link each component name individually in qualified names, but this could be replaced with a button to navigate to the enclosing namespace.

fpelliccioni:
Fixes #469

It took me quite a while to investigate where the SVG "house" icon is being generated.

image

I explored various parts of the codebase and toolchain but couldn't locate it. As of now, I still don’t know if this behavior can be modified. Based on my research so far, it doesn't seem to be controlled by MrDocs but rather by subsequent tools in the toolchain, such as Antora.

I came up with this change—not ideal, but it’s a starting point to evaluate if it’s acceptable. If it’s not, any pointers on how to control the house icon generation would be much appreciated. As I mentioned, it doesn’t seem to be within the MrDocs repository.

### DIFF
--- a/share/mrdocs/addons/generator/adoc/layouts/wrapper.adoc.hbs
+++ b/share/mrdocs/addons/generator/adoc/layouts/wrapper.adoc.hbs
@@ -9,7 +9,7 @@
 
     Breadcrumbs are included in the symbol template as an alternative.
 }}
-= {{> symbol/qualified-name symbol }}
+= {{> symbol/enclosing-namespace symbol }}{{> symbol/qualified-name symbol }}
 {{!
     Antora does not support relative links to pages in
     parent directories. Therefore, unlike in HTML templates,
--- a/share/mrdocs/addons/generator/common/partials/symbol/enclosing-namespace.hbs
+++ b/share/mrdocs/addons/generator/common/partials/symbol/enclosing-namespace.hbs
@@ -1,0 +1,14 @@
+{{!--
+    Renders a link to the enclosing namespace with a custom symbol (⬆).
+    If no enclosing namespace exists, nothing is rendered.
+
+    Expected Context: {Symbol Object}
+
+    Example:
+        {{> symbol/enclosing-namespace symbol }}
+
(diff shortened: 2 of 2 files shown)


share/mrdocs/addons/generator/common/partials/symbol/enclosing-namespace.hbs
```diff
@@ -0,0 +1,14 @@
...
+
+    Expected Context: {Symbol Object}
+
+    Example:
+        {{> symbol/enclosing-namespace symbol }}
+
+    See: https://mrdocs.com/docs/mrdocs/develop/generators.html#dom_reference
+--}}
+{{~#if (and parent parent.name)~}}
+    xref:{{{remove_prefix parent.url "/"}}}[⬆]
```
alandefreitas, line 13: xref is not valid HTML. These common partials are used by both generators. See the other partials in this folder.

share/mrdocs/addons/generator/adoc/layouts/wrapper.adoc.hbs
```diff
@@ -9,7 +9,7 @@
 
     Breadcrumbs are included in the symbol template as an alternative.
 }}
-= {{> symbol/qualified-name symbol }}
+= {{> symbol/enclosing-namespace symbol }}{{> symbol/qualified-name symbol }}
```
alandefreitas, line 12: This worsens things because we have removed the links to other enclosing namespaces, which the `qualified-name` partial does. It's removing a feature rather than including one.

alandefreitas: I was surprised the test files didn't change. I think it's because tests bypass the wrapper. But losing this information about the enclosing namespaces in real documentation like Boost.URL sure looks bad.

alandefreitas:
See https://github.com/cppalliance/mrdocs/issues/469#issuecomment-2554516137

Issue #765, opened 2024-12-06
Descriptions with paragraphs, code, and other elements rendered incorrectly


anarthal:
```cpp
/// This is another function.
void other_function();

/**
 * \brief A function
 * \details Relies on \ref other_function.
 *
 * Example:
 * \code
 *   my_function();
 * \endcode
 *
 * \par Exception safety
 * No-throw guarantee.
 */
void my_function();
```

Renders

![Image](https://github.com/user-attachments/assets/3b4c685c-e0bf-4932-92f3-e6f837c55c60)

Nerixyz:
Which version of mrdocs are you using? This looks fine in the [current release](https://github.com/cppalliance/mrdocs/releases/tag/develop-release):

![Image](https://github.com/user-attachments/assets/92d0d2d0-856a-4e8f-a846-889fb17754f6)

anarthal:
No, this image is not fine. The title should be Exception safety, which is missing, and not No-throw guarantee.

PR #775, opened 2024-12-23
fix: use arguments to `\par` if available

Nerixyz:
`\par` accepts the heading as the first argument.

Fixes https://github.com/cppalliance/mrdocs/issues/765.

Aside: I noticed that the tests for the adoc and HTML generators are taking much longer than the XML tests (also visible in CI).

### DIFF
--- a/test-files/golden-tests/para-3.adoc
+++ b/test-files/golden-tests/para-3.adoc
@@ -1,0 +1,51 @@
+= Reference
+:mrdocs:
+
+[#index]
+== Global namespace
+
+=== Functions
+[cols=2]
+|===
+| Name | Description 
+
+| <<#my_function,`my&lowbar;function`>> 
+| A function
+
+|===
+
+[#my_function]
+== my&lowbar;function
+
+A function
+
+=== Synopsis
+
+Declared in `&lt;para&hyphen;3&period;cpp&gt;`
+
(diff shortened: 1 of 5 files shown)


test-files/golden-tests/para-3.adoc
```diff
@@ -0,0 +1,51 @@
...
+[source,cpp,subs="verbatim,replacements,macros,-callouts"]
+----
+void
+my&lowbar;function();
+----
+
+=== Description
+
+
+=== A heading&period;
```
Nerixyz, line 35: This seems wrong - the "A heading." should probably be one level "lower" than "Description" (but that's not really part of this PR, I suppose).

test-files/golden-tests/para-3.html
```diff
@@ -0,0 +1,67 @@
...
+<pre>
+<code class="source-code cpp">
+void
+my_function();
+</code>
+</pre>
+</div>
+<div>
+<h3>Description</h3>
+<h3>A heading.</h3>
```
Nerixyz, line 47: Same here.

test-files/golden-tests/para-3.adoc
alandefreitas: Yes. This seems reasonable enough. There's a lot to think about regarding how this overview is structured.

PR #784, opened 2025-01-04
fix: use correct config path

sdkrystian:


alandefreitas:
So `cmdLineInputs` has priority over `--config`?

sdkrystian:
@alandefreitas yes... there is a way to do this in llvm itself (I think), but that requires adding features to the configuration generator

alandefreitas:
That looks weird to me. 

Why would anyone attempt to implicitly set the `--config` option as part of the positional inputs and explicitly via `--config`?

In the weird case that the user does that, why would the implicit option have precedence over the explicit option? If someone is set _explicitly_, it needs to have precedence by definition.

alandefreitas:
Another point is this behavior would be in conflict with what `PublicSettingsVisitor` does. Do different parts of the code could assume different config files. Or the effective config file could randomly change.

sdkrystian:
> That looks weird to me.
> 
> Why would anyone attempt to implicitly set the `--config` option as part of the positional inputs and explicitly via `--config`?

I don't know. I don't think we should accept config files as positional arguments --  only via `--config`. Nonetheless, we support passing both ways, and that is a problem because we parse the arguments prior to knowing whether  a config file was passed as a positional argument (unless we use `c_callback` or something), resulting in the default config path being used.

alandefreitas:
> I don't know. I don't think we should accept config files as positional arguments -- only via --config. 

Yes. This is for backward compatibility. We don't really need backward compatibility at this point (0.0.3) but I just didn't want people yelling at me at the time.

> Nonetheless, we support passing both ways, and that is a problem because we parse the arguments prior to knowing whether a config file was passed as a positional argument (unless we use c_callback or something), 

I don't agree with that. The function this PR changes has access to both the inputs (toolArgs.cmdLineInputs) and the positional argument (toolArgs.config).

Maybe you're talking about the normalization logic in Config.cpp. We could see if we have some bug there. But this PR doesn't touch that.

> resulting in the default config path being used.

Can you provide an example command line where the current implementation would lead to an incorrect result?

alandefreitas:
Obsoleted by c73a03b0c527d7de2b41a14766a6349ee778b7d5 and f3e0b4a8d9ea1c181c518b9c69a9b5e4542c76b6

Issue #676, opened 2024-09-28
Provide a list of all supported `@` annotations


akrzemi1:
I know of at least one `@` annotation -- `@tparam` -- that is not part of "standard" Javadoc annotations.
Please have the documentation provide a comprehensive list.

Also, the docs mention the annotations like `@em`, but I am missing an example. In HTML there is a starting tag for `<em>` and an an ending tag `</em>`. This pair clearly delimits the code that needs to be emphasized. How does this work for `@em? How do I tell it that only the three next words need to be emphasized?

akrzemi1:
For instance, I would like to know, if it has things like `@precondition`, `@postcondition`...

alandefreitas:
In the mrdocs documentation, each @ command needs its own page with:

- the name of the command
- the syntax of the command
- a verbal description of what it does
- an example javadoc and declaration which uses the command
- the result of rendering the example above

alandefreitas:
Update:

We support everything clang does plus a few other markdown features clang doesn’t support yet. It’s really hard to maintain a list because there’s no single source of truth for what’s supported (support is scattered throughout the parsing code). There’s also this idea floating around where we’re going to operate directly in the clang parser via some kind of hook that’s going to make things still a little more complex. 

One idea I had was that for each kind of javadoc element possible, we could force the code to include a short description of how to create that kind of element. That would give us some sort of source of truth in terms of what elements could be created, although there’s no way to ensure people will keep the list of ways to create an element up to date. Also, the current direction for the project is to work on other things.

Someone actually worked on this issue a little bit but the result wasn’t that great. It ended up like a list of manual pages replicating the content from doxygen.nl in asciidoc. We didn't merge it.

PR #821, opened 2025-01-24
doc: Comprehensive list of supported commands

fpelliccioni:
Fixes https://github.com/cppalliance/mrdocs/issues/676

In the file [`src/lib/AST/ParseJavadoc.cpp`](https://github.com/cppalliance/mrdocs/blob/develop/src/lib/AST/ParseJavadoc.cpp), there are mainly two functions in charge of handling documentation commands:

1. `JavadocVisitor::visitInlineCommandComment(...)`  
   This one mentions inline commands such as: `@n`, `@a`, `@e`, `@em`, `@copybrief`, `@copydetails`, `@copydoc`, and `@ref`.

2. `JavadocVisitor::visitBlockCommandComment(...)`  
   - **It treats certain block commands in a special manner**, for example:  
     `@brief`, `@short`, `@return` / `@returns` / `@result`, `@throw` / `@throws` / `@exception`, `@note`, `@warning`, `@par`, `@li`, `@details`, `@see`, `@pre`, `@post`.  
     These commands trigger specific behavior in MrDocs.
   - **It does not produce errors but also does nothing** with many other commands (e.g., `@addindex`, `@addtogroup`, `@anchor`, `@arg`, …, `@xrefitem`).  
     Although they are present in the `CommandTraits` enum and do not cause errors, MrDocs currently does nothing with them. 

I have decided to **include all these "passively supported" commands** in the documentation because:
- Even though MrDocs does nothing concrete with them, they are recognized in the `CommandTraits` enumeration.
- Some of the "passively supported" commands reference others in the list, creating a chain of mentions.

Additionally, I have made the following documentation adjustments:

- **Removed references to other programming languages** (Objective-C, Java, etc.). For instance, commands like `@category` and `@protocol` (which apply only to Objective-C) were removed. We kept The C programming languages for compatibility and relevance alongside C++. Code, examples and mentions to other programming languages were removed.

- **Added the `@endlink` command**, which was commented out in the `visitBlockCommandComment(...)` switch but referenced by `@link`. I decided to keep it to cover the `@link ... @endlink` pair.
- There are **open questions** about how to handle other points mentioned in the docs:
  - Typical Doxygen configuration options (e.g. `CALLER_GRAPH`, `ENABLED_SECTIONS`, etc.).
  - Support for HTML-like syntax (e.g. `<em>multiple words</em>`) which is referenced in multiple places.
  - The `@startuml` command, which suggests needing Java and PlantUML. We need to decide if we want to support it. 

With this PR, all commands included in the enumeration are now documented, so they can be referenced, even though MrDocs does not yet perform additional processing for some of them.


alandefreitas:
> Even though MrDocs does nothing concrete with them, they are recognized in the CommandTraits enumeration.

So they're unsupported?

> Some of the "passively supported" commands reference others in the list, creating a chain of mentions.

So they're supported?

docs/modules/ROOT/pages/commands-list.adoc
```diff
@@ -0,0 +1,176 @@
+= Comprehensive List of Commands
```
alandefreitas, line 1: There's already a section for this on the commands page.

The term "commands" for documentation commands might be ambiguous with command line commands.

This could be something like "Documentation Command Reference" but the navtitle would have to be shorter like "Documentation Commands" or something,

docs/modules/ROOT/pages/commands-list.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+Below is the complete list of supported commands:
+
+* xref:commands/a.adoc[@a]
```
alandefreitas, line 5: This page is kind of boring and ugly.

Commands could be on a table like the ones mrdocs generates where the first column has the command and the second column has a brief description.

To make it less repetitive, commands can be put into subsections categories. The basic categories are block and inline commands. Then there's subcategories based on what the command does.

Each section can have a small paragraph explaining the table to come.

The command text should be as wrapped in "`"s.

Unsupported commands can be in red without a page so that users and developers can keep track of them.

Unimplemented commands are unsupported commands.

docs/modules/ROOT/pages/commands.adoc
```diff
@@ -56,6 +56,11 @@ The following commands can be used to format the text in the doc comments:
 |
 |===
 
+== Full List of Commands
```
alandefreitas, line 59: We don't need that (the nav is more visible already). The one that already exists also needs to be removed.

docs/modules/ROOT/pages/commands/a.adoc
```diff
@@ -0,0 +1,20 @@
...
+== Syntax
+@a &lt;word&gt;
+
+== Description
+Displays the argument <word> in italics. Use this command to emphasize words. Use this command to refer to member arguments in the running text.
+
+Equivalent to xref:commands/e.adoc[@e] and xref:commands/em.adoc[@em]. To emphasize multiple words use <em>multiple words</em>.
+
+== Example
+```
```
alandefreitas, line 12: That's markdown. Not asciidoctor.

docs/modules/ROOT/pages/commands/anchor.adoc
```diff
@@ -0,0 +1,10 @@
+= anchor
+
+== Syntax
+@anchor &lt;word&gt;
+
+== Description
+This command places an invisible, named anchor into the documentation to which you can refer with the xref:commands/ref.adoc[@ref] command.
```
alandefreitas, line 7: All supported commands need examples.

docs/modules/ROOT/pages/commands/callergraph.adoc
```diff
@@ -0,0 +1,14 @@
+= callergraph
+
+== Syntax
+@callergraph
+
+== Description
+When this command is put in a comment block of a function or method and HAVE_DOT is set to `YES`, then MrDocs will generate a caller graph for that function (provided the implementation of the function or method is called by other documented functions). The caller graph will be generated regardless of the value of CALLER_GRAPH.
```
alandefreitas, line 7: No point in describing unsupported commands.

docs/modules/ROOT/pages/commands/callergraph.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+== Syntax
+@callergraph
+
+== Description
+When this command is put in a comment block of a function or method and HAVE_DOT is set to `YES`, then MrDocs will generate a caller graph for that function (provided the implementation of the function or method is called by other documented functions). The caller graph will be generated regardless of the value of CALLER_GRAPH. 
+
+
+
+*Note:* The completeness (and correctness) of the caller graph depends on the MrDocs code parser which is not perfect.
```
alandefreitas, line 11: That's not imperfect. It's 100% unsupported.

docs/modules/ROOT/pages/commands/dontinclude.adoc
```diff
@@ -0,0 +1,47 @@
...
+The class and member declarations and definitions inside the code fragment are 'remembered' during the parsing of the comment block that contained the `@dontinclude` command.
+
+For line by line descriptions of source files, one or more lines of the example can be displayed using the xref:commands/line.adoc[@line], xref:commands/skip.adoc[@skip], xref:commands/skipline.adoc[@skipline], and xref:commands/until.adoc[@until] commands. An internal pointer is used for these commands. The `@dontinclude` command sets the pointer to the first line of the example.
+
+== Example
+```
+ 
+/*! A test class. */
+ 
+class Include_Test
```
alandefreitas, line 22: Please don't tell me you're copy/pasting these from doxygen's documentation.

PR #861, opened 2025-02-14
correctly handle files included through symlinks

### LINKED ISSUE #819: Support symlinks as source files
When we add symlinks to `test-files/golden-tests/snippets/<test-file>.cpp` with the target `docs/website/snippets/<test-file>.cpp`, `FileEntry::tryGetRealPathName();` will return the `docs/website/snippets/<test-file>.cpp` target instead of the file we used in the compilation database and that's how we populate the FileInfo.

The problem is no symbol defined in `docs/website/snippets/<test-file>.cpp` passes the `input` filter `test-files/golden-tests/snippets` and MrDocs extracts nothing. FileInfo should be populated with the filename the user provided to clang instead.

sdkrystian:
fix #819


alandefreitas:
By the way, the current test case is replacing the snippets in golden-tests/snippets/*.cpp with symlinks to docs/website/snippets/*.cpp

These are the cases we have for now but there will be many more in the future because we will include examples of almost everything in the documentation (commands, configuration options, etc).

PR #895, opened 2025-04-05
fix: group overloads per access level

Nerixyz:
The `OverloadFinalizer` groups functions for one access level and "static" level. However, the ID assigned to the overloads only took the parent and function name into account. This meant that functions of the same name in a different access level would map to the same ID and be discarded if that ID already existed (`emplace` behavior).

The ID for overloads now takes the access level and "static" level into account.


alandefreitas:
Oh... That's nice. Is it good to go?

Nerixyz:
Yes, this is good to go.

alandefreitas:
I just noticed the test in [test-files/golden-tests/config/overloads/visibility.cpp](https://github.com/cppalliance/mrdocs/pull/895/files#diff-53f5c74cdfe42f16268108baf5361902323605081308095144e16f040cbdd32a) won't form any overload sets. It's good to check if it's not forming when it shouldn't form but also forming these separate sets when they are expected by the PR.

alandefreitas:
OK. Merging it after #896.

PR #896, opened 2025-04-05
fix: merge namespaces when collecting results

Nerixyz:
I'm not 100% sure if this is the correct approach. I would guess that something like this already happens somewhere, but I couldn't find anything.

When using multiple sources (→ multiple extraction jobs), then only the first one will be "visible", because its global namespace/global symbol will be the only one that gets merged with `info_.merge(info)`. All other global namespaces will be discarded, because there's already one present. Their members will still be in the corpus, but there's no path from the global symbol to them.

Because of ODR, only namespaces have to be merged.

### DIFF
--- a/src/lib/Lib/ExecutionContext.cpp
+++ b/src/lib/Lib/ExecutionContext.cpp
@@ -13,7 +13,9 @@
 
 #include "ExecutionContext.hpp"
 #include "lib/Metadata/Reduce.hpp"
+#include "mrdocs/Support/Assert.hpp"
 #include <mrdocs/Metadata.hpp>
+#include <mrdocs/Metadata/Info/Namespace.hpp>
 #include <ranges>
 
 namespace clang {
@@ -69,7 +71,11 @@ report(
     {
         auto it = info_.find(other->id);
         MRDOCS_ASSERT(it != info_.end());
-        merge(**it, std::move(*other));
+        visit(**it, [&]<typename T>(T& target) {
+            auto *source = dynamic_cast<T*>(other.get());
+            MRDOCS_ASSERT(source);
+            merge(target, std::move(*source));
+        });
     }
 
     // Merge diagnostics and report any new messages.


alandefreitas:
This looks very weird to me.

> When using multiple sources (→ multiple extraction jobs), then only the first one will be "visible"

If by "visible" you mean copied into `info_`, then no. `info_.merge(info);` is called on all `info`s for all TUs. Any new symbols will be added there. The loop is only for symbols that couldn't be included in the `info_` because they're duplicates. Then we have the `merge` algorithm for this.

> its global namespace/global symbol will be the only one that gets merged with info_.merge(info)

Not really. The second, third, ... is supposed to be merged with `merge(**it, std::move(*other));`.

> Their members will still be in the corpus, but there's no path from the global symbol to them.

`merge(**it, std::move(*other));` is supposed to create this "path" (include the members in the well... members of the namespace). Even if `merge(**it, std::move(*other));` didn't do that, I'm not sure how this casting would change things unless the implementation of `merge` is wrong.

Nerixyz:
> Even if `merge(**it, std::move(*other));` didn't do that, I'm not sure how this casting would change things unless the implementation of `merge` is wrong.

Then the implementation of `merge` might be wrong. In the execution context, [`merge(Info& I, Info&& Other)`](https://github.com/cppalliance/mrdocs/blob/5ee5c49d477d86e8ae35fe072b63519533773050/src/lib/Metadata/Info.cpp#L40-L68) is called, which doesn't perform any downcasting, it only upcasts to `SourceInfo`. `merge(NamespaceInfo& I, NamespaceInfo&& Other)` is never called from there. In fact, it would [merge the infos](https://github.com/cppalliance/mrdocs/blob/5ee5c49d477d86e8ae35fe072b63519533773050/src/lib/Metadata/Info/Namespace.cpp#L109) (if called).

I think the intention was to call the most specific `merge` function. One could either turn `merge` into a virtual function or use `visit`:

```cpp
visit(**it, [&]<typename T>(T& target) {
    auto *source = dynamic_cast<T*>(other.get());
    MRDOCS_ASSERT(source);
    merge(target, std::move(*source));
});
```

^ this might be the intention here.

alandefreitas:
Exactly!

alandefreitas:
Now, this is good to go, right? I'll merge this one first because whatever I merge, the other needs to be rebased, and this one is more important.

Nerixyz:
Thank you!

PR #900, opened 2025-04-23
fix(clang-cl): reduce warnings

Nerixyz:
When compiling with clang-cl, there are warnings not present with MSVC

- `/MP` isn't a thing in clang-cl.
- `/W4` enables `-Wunused-parameter` in clang-cl (or this might be enabled by default) - either way, it's not enabled in MSVC and generates some warnings here.
- `getRawUnqualified` returned a `std::string_view` to a temporary `std::string` and clang warned about it.
- `-Wself-move` was ignored when `__GNUC__` was defined. This isn't defined with clang-cl, so ignore it in the clang block.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -276,11 +276,12 @@ if (WIN32)
         target_compile_options(
             mrdocs-core
             PUBLIC
-            /permissive-    # strict C++
-            /W4             # enable all warnings
-            /MP             # multi-processor compilation
-            /EHs            # C++ Exception handling
-            $<$<CONFIG:Debug>:/Oy-> # Disable frame pointer omission
+            /permissive-                    # strict C++
+            /W4                             # enable all warnings
+            $<$<CXX_COMPILER_ID:MSVC>:/MP>  # multi-processor compilation
+            /EHs                            # C++ Exception handling
+            $<$<CONFIG:Debug>:/Oy->         # Disable frame pointer omission
+            $<$<CXX_COMPILER_ID:Clang>:-Wno-unused-parameter>  # ClangCL warning enabled with /W4
         )
     endif()
 endif ()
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -276,11 +276,12 @@ if (WIN32)
...
-            /W4             # enable all warnings
-            /MP             # multi-processor compilation
-            /EHs            # C++ Exception handling
-            $<$<CONFIG:Debug>:/Oy-> # Disable frame pointer omission
+            /permissive-                    # strict C++
+            /W4                             # enable all warnings
+            $<$<CXX_COMPILER_ID:MSVC>:/MP>  # multi-processor compilation
+            /EHs                            # C++ Exception handling
+            $<$<CONFIG:Debug>:/Oy->         # Disable frame pointer omission
+            $<$<CXX_COMPILER_ID:Clang>:-Wno-unused-parameter>  # ClangCL warning enabled with /W4
```
alandefreitas, line 284: Shouldn't we fix the code triggering the warning instead of disabling the warning?

src/lib/Support/LegibleNames.cpp
```diff
@@ -192,7 +192,7 @@ class LegibleNames::Impl
         If the symbolhas no name, then a reserved name based
         on the type is returned instead.
      */
-    std::string_view
+    std::string
```
alandefreitas, line 195: :)

CMakeLists.txt
Nerixyz: I made a separate commit now. I initially added this because the warnings weren't present in the output of GCC and Clang on macOS/Linux.

alandefreitas: Yes. At some point we have to enable all these warnings there too. It's just that the project started without the these warnings and we always have higher priority stuff to work on.

Issue #914, opened 2025-07-02
MrDocs should print the commands it is running


gennaroprota:
When running an external command, MrDocs should print the full command for the user, so that they can more easily replicate it in case of errors, and know that the error is coming from an external process.

PR #918, opened 2025-07-03
Print all external commands

gennaroprota:
This closes issue #914.


alandefreitas:
> Maybe the function name should be "executeAndWaitWithLogging" (lowerCamelCase)?

We can do that in passing later because otherwise we'd have to wait for CI all over again. I've been restarting CI since yerterday lol

PR #922, opened 2025-07-08
fix: crash on unimplemented when populating overloads destructor entry

mizvekov:
I am not sure why this is not seen in CI presently, but I can reproduce it when building mrdocs from the llvm version bundled in homebrew (currently 20.1.7).

### DIFF
--- a/src/lib/Metadata/Finalizers/OverloadsFinalizer.cpp
+++ b/src/lib/Metadata/Finalizers/OverloadsFinalizer.cpp
@@ -105,6 +105,7 @@ foldOverloads(SymbolID const& contextId, std::vector<SymbolID>& functionIds, boo
         MRDOCS_CHECK_OR_CONTINUE(recordInfoPtr);
         auto* function = recordInfoPtr->asFunctionPtr();
         MRDOCS_CHECK_OR_CONTINUE(function);
+        MRDOCS_CHECK_OR_CONTINUE(function->Class != FunctionClass::Destructor);
 
         // Check if the FunctionInfo is unique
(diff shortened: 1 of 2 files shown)


src/lib/Metadata/Finalizers/Javadoc/Overloads.hpp
```diff
@@ -76,6 +76,10 @@ populateOverloadsFromClass(OverloadsInfo& I)
         return true;
     }
     case FunctionClass::Destructor:
+    {
```
alandefreitas: Destructor overloads?

alandefreitas: Isn't the problem in the calling function?

mizvekov: If you accept the logical consistency about having a destructor overload set which can only be a single element.

The resulting adoc looks perfectly reasonable. Maybe we could just rename the brief to 'Destructor' instead of 'Destructors'?

alandefreitas: Yes. You could be correct in theoretical terms. What I mean is that, in practical terms, as an implementation detail of MrDocs, where an overload set must have at least two members, that code is unreachable. If it's being reached, there's a bug somewhere else we need to fix. Or maybe it should be reachable for some reason I haven't understood yet.

Even if the resulting adoc looks good in terms of what we were able to investigate so far, we still have a problem here, and we have to understand why it's happening and what else it's affecting in the corpus and the output.

PR #923, opened 2025-07-08
fix: handle constructor from reference to array

mizvekov:
isCopyOrMoveConstructorOrAssignment currently doesn't handle a constructor from a reference to array. I am not sure why this is not seen on CI, but it is reproducible from llvm 20.1.7, as available in macos homebrew.

### DIFF
--- a/src/lib/Metadata/Finalizers/Javadoc/Function.hpp
+++ b/src/lib/Metadata/Finalizers/Javadoc/Function.hpp
@@ -60,9 +60,11 @@ isCopyOrMoveConstructorOrAssignment(FunctionInfo const& I)
     auto const& paramRefType = get<RefType const&>(paramType);
     auto const& paramRefPointee = paramRefType.PointeeType;
     MRDOCS_CHECK_OR(paramRefPointee, false);
-    auto const& paramRefPointeeNamed = get<NamedTypeInfo const&>(paramRefPointee);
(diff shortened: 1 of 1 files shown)


src/lib/Metadata/Finalizers/Javadoc/Function.hpp
```diff
@@ -60,9 +60,11 @@ isCopyOrMoveConstructorOrAssignment(FunctionInfo const& I)
     auto const& paramRefType = get<RefType const&>(paramType);
     auto const& paramRefPointee = paramRefType.PointeeType;
     MRDOCS_CHECK_OR(paramRefPointee, false);
-    auto const& paramRefPointeeNamed = get<NamedTypeInfo const&>(paramRefPointee);
-    MRDOCS_CHECK_OR(paramRefPointeeNamed.Name, false);
-    auto const& paramName = paramRefPointeeNamed.Name;
+    auto const* paramRefPointeeNamed = get<NamedTypeInfo const*>(paramRefPointee);
+    if (!paramRefPointeeNamed)
```
alandefreitas, line 64: For reference, this could be `MRDOCS_CHECK_OR(paramRefPointeeNamed, false)`.

MRDOCS_CHECK is that pattern people use with `std::expected` (like `?` in Rust), so that code based on error types doesn't become impossible to read.

PR #927, opened 2025-07-11
fix: replace uses of deprecated fmt::has_formatter

mizvekov:
use is_formattable instead.

has_formatter was never a good API, since it only asked if the type has a `formatter` specialization, but a type can also be formattable if it provides an implicit conversion to a type which does.

The newer API takes care of the that problem.


alandefreitas:
I rebased your branch, but still not working. Good thing the compilers that failed are the ones you have easy access to in your OS. :)

PR #935, opened 2025-07-18
ci: enable ubsan for clang and gcc

mizvekov:
And fixes a member function call through a null pointer and an incorrect use of static_cast.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,8 @@ jobs:
           standards: '20'
           latest-factors: |
             msvc Optimized-Debug
-            gcc Coverage
+            gcc UBSan Coverage
+            clang UBSan
           factors: ''
           runs-on: |
             apple-clang: macos-15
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -215,7 +217,7 @@ jobs:
 
       - name: Download LLVM Binaries
         id: llvm-download
-        if: steps.llvm-cache.outputs.cache-hit != 'true'
+        if: steps.llvm-cache.outputs.cache-hit != 'true' && !(matrix.ubsan || matrix.asan || matrix.msan)
```
alandefreitas: Ideally, we should also upload these binaries. The binaries on the website are not only useful for developers. They're also helpful when testing CI, and the GHA cache is not yet available.

alandefreitas: I think it would be really, really good to upload these binaries to the site as well. 

I even thought this would already happen because of that automatic matrix, but I guess the naming conflict (with and without sanitizer having the same filename) is probably blocking it. But that’s a problem, because in a run where the cache is cold, there’s a chance the job with sanitizer finishes first and uploads the sanitizer version to the server as the default one. 

You could fix this by filtering out those sanitizer jobs in the script that filters the binaries, but honestly, at that point it’s less work to just upload the files there.

PR #946, opened 2025-07-28
cmake: require libc++ and clang headers when building with tests or install

mizvekov:
libc++ and clang header includes are required when running tests, or when installing mrdocs.

Make the CMakeLists.txt enforce that requirement.

Otherwise, the tests fail in non-obvious ways.


mizvekov:
> > Make the CMakeLists.txt enforce that requirement.
> 
> The code might be flawed but it already enforces this requirement. This PR only changes the condition (it should be unconditional anyway).

You can make both calls to find_package (LLVM, Clang) succeed, without specifying `LLVM_ROOT` at all, by specifying `LLVM_DIR` and `Clang_DIR` instead.

This makes configuration succeed, but it results in a non-functional installation and tests.

If you mean we should just unconditionally require those things instead, that also works for me.

mizvekov:
> In other words, this is more of a fix than a feature.

Yes, that was the intention here, this is strictly a fix.

PR #950, opened 2025-07-30
ci: add asan and msan Linux jobs

mizvekov:
ASan is not viral, so it is added to both GCC and Clang.

MSan is viral, so it is added to Clang only.

Do not add MSan to MacOS job, since it is not supported there.


.github/workflows/ci.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Generate Test Matrix
-        uses: alandefreitas/cpp-actions/cpp-matrix@v1.8.10
+        uses: alandefreitas/cpp-actions/cpp-matrix@80162e6100a99e06322e3a7ef2b6fe9d5b6819cd
```
alandefreitas: Let me create this release.

alandefreitas: v1.8.11

.github/workflows/ci.yml
```diff
@@ -66,21 +66,29 @@ jobs:
             clang: git build-essential pkg-config python3 curl openjdk-11-jdk pkg-config libncurses-dev libxml2-utils libxml2-dev g++-14=14.2.0-4ubuntu2~24.04
             msvc: ''
           extra-values: |
-            llvm-hash: a1b6e7ff393533a5c4f3bdfd4efe5da106e2de2b
-            llvm-build-preset-prefix: {{#if optimized-debug}}debwithopt{{else}}{{{lowercase build-type}}}{{/if}}
+            use-libcxx: {{#if (and (ieq compiler 'clang') (or (eq major 19) (eq major 20) (eq major 21))) }}true{{else}}false{{/if}}
```
alandefreitas: clang will be always tested with libcxx now? Even on linux? Even when unrelated to sanitizers.

alandefreitas: Isn't there a ge (an "operator>=") handlebars helper?

.github/workflows/ci.yml
```diff
@@ -66,21 +66,29 @@ jobs:
...
+            libcxx-runtimes: libcxx{{#if (ne compiler 'msvc')}};libcxxabi{{/if}}
+            llvm-runtimes: {{#if (ine use-libcxx 'true') }}{{{ libcxx-runtimes }}}{{/if}}
+            llvm-path: {{#if (ieq compiler 'msvc') }}D:/a{{else if (ieq compiler 'apple-clang')}}/Users/runner/work{{else}}/__w{{/if}}/llvm
```
alandefreitas: Hard-coding this is very unstable. Any new container will keep breaking this.

I know it's so we can `-isystem{{{ llvm-path }}}/include/c++/v1{{/if}}` later on, but the expression can still be evaluated at the job step because we're allowed to use more than one variable in the cmake workflow step or wherever you might want these flags.

In GHA actions, expressions are represented with ${{ ... }}, such as:

```
https://github.com/doxygen/doxygen/releases/download/Release_1_9_7/doxygen-1.9.7.{{ os }}${{ ( runner.os == 'Windows' && '.x64' ) || '' }}.bin.${{ ( runner.os == 'Windows' && 'zip' ) || 'tar.gz' }}
```

.github/workflows/ci.yml
```diff
@@ -198,18 +206,15 @@ jobs:
         with:
           apt-get: ${{ matrix.install }}
 
-      - name: Resolve LLVM Root
```
alandefreitas, line 201: Yes. This was the step resolving it.

.github/workflows/ci.yml
```diff
@@ -261,9 +266,44 @@ jobs:
               echo "found=$found" >> $GITHUB_OUTPUT
           fi
 
+      - name: Install libc++
```
alandefreitas, line 306: It's good to include some comments here because this is very confusing. I undertand why this is being done, but it's very confusing to anyone new at this.

alandefreitas: Wait... Now I'm not sure even I understand it. The "Install LLVM" step is building libc++ again, right? Like, every single time for all workflows. Is that correct?  Isn't there a better way?

.github/workflows/ci.yml
```diff
@@ -882,25 +930,12 @@ jobs:
           fi
           echo "exists=$exists" >> $GITHUB_OUTPUT
 
-      - name: Resolve LLVM Root
```
alandefreitas, line 963: And here...

alandefreitas (commented):
We'll need to replicate this logic in bootstrap because I don't think anyone can replicate this themselves. So if there's an error in MSan everyone will be blocked from solving it locally.

.github/workflows/ci.yml
mizvekov: I didn't find an operator equivalent to `>=`, it should be spelled 'ge' I think for consistency with naming of existing operators.

.github/workflows/ci.yml
mizvekov: No it's not building it again. Either this step runs because we have 'use-libcxx' true, and then we don't build libc++ later using the LLVM bootstrap build, or this step is skipped because 'use-libcxx' is not true, but then we build libc++ on the bootstrap build later.

The libc++ on this step is built using the host compiler, and this can only be done with the compilers which are supported by it: [apple-]clang >= 19 and gcc >= 15.

.github/workflows/ci.yml
mizvekov: It would be good if cpp-actions supported expanding these '${{ ' expressions later, but that doesn't work ATM,
this string would end up used as-is.

mizvekov: I know hard-coding these paths is not ideal, but GitHub is not very helpful here, and the options to avoid this end up being much more complicated.

.github/workflows/ci.yml
mizvekov: Yeah, I felt that there was some value in using these on clang even without sanitizers, because at least the libcxx built here will be controlled by the same flags, so it will have debug info, will have runtime assertions if we choose to enable it, and so on. It would give some extra variation on top of what we already do on the GCC jobs.

alandefreitas: The operators are here:

https://github.com/alandefreitas/cpp-actions/blob/80162e6100a99e06322e3a7ef2b6fe9d5b6819cd/cpp-matrix/index.js#L1189

alandefreitas: We don't have it but we need to define it.

.github/workflows/ci.yml
alandefreitas: > It would be good if cpp-actions supported expanding these '${{ ' expressions later, but that doesn't work ATM,

Yes. This is impossible.

> but GitHub is not very helpful here

The example above is how you expand that later. We use this solution all the time. Even mrdocs ci.yml uses it all the time, for instance:

```
${{ github.event_name == 'push' && (contains(fromJSON('["master", "develop"]'), github.ref_name) || startsWith(github.ref, 'refs/tags/')) }}
```

There's nothing wrong with this pattern to expand it later.

.github/workflows/ci.yml
alandefreitas: By again, I mean later. This step is building it for the first time when `use-libcxx` is true, and that's clear. The problem is the "Install LLVM" step will always build it for a second time (if `use-libcxx` is true).

mizvekov: I am saying the "Install LLVM" won't build it for a second time, because if 'use-libcxx' is true, then `llvm-runtimes` will be empty and we won't build any runtime component in that step.

mizvekov: This is actually crucial by the way, if we didn't build these things separately, we just wouldn't have disk space for the sanitizer jobs.

alandefreitas: > I am saying the "Install LLVM" won't build it for a second time, because if 'use-libcxx' is true, then llvm-runtimes will be empty and we won't build any runtime component in that step.

OK. Please add some comments in the ci.yml file because the logic is quite hard to follow and remember. 

We also need that logic replicated in the bootstrap. Maybe the libcxx step should always be separate there. I don't know.

mizvekov: libcxx as a separate step will use the host compiler, but the support here is quite limited, just GCC 15 and Clang 19 onwards are supported. It can be compiled on windows, but that requires clang-cl as the host compiler.

.github/workflows/ci.yml
mizvekov: I tried that before submitting this patch, but I felt like it makes things so much harder to read because it needs to be expanded in quite a few places.

alandefreitas: > I tried that before submitting this patch

I don't understand why you can't just expand it once. 

What pattern did you use?

.github/workflows/ci.yml
alandefreitas: We're probably talking past each other because I don't understand how that relates to the previous comment.

Yes. We have libcxx as a separate step. That's OK. My only concern is 1) this is going to build libcxx twice (unless that's how it should be anyway), 2) whatever happens, we have comments explaining what's happening (especially because this runtimes var seems to be overridding LLVM_ENABLE_RUNTIMES in the presets with a different value / we have many sources of truth for the same thing and that's not great). 

We also build libcxx separately sometimes, and sometimes we don't. That's another source of truth that's not great. But well, I guess this is just the case for msan. That's why we need the bootstrap procedure. You're the only one who understands this thing.

.github/workflows/ci.yml
mizvekov: So the idea I had in mind was to define llvm-path as `{0}/../../llvm`.

Then from there, whenever a variable which depends on llvm-path needs to be used, we can format it like so:
`cxxflags: ${{ format(matrix.common-cxxflags, github.workspace) }}`
`-D CMAKE_EXE_LINKER_FLAGS="${{ format(matrix.common-ldflags, github.workspace) }}"` 
etc

There are quite a few places which depend on a variable which depends on llvm-path.

Now given GitHub's scale and Hyrum's law,  I seriously doubt GitHub can realistically change this, at least for the current runner images which exist.

alandefreitas: Do you know how to form step outputs, like the one that was being created in the "Resolve LLVM Root" you removed? Why not just also concatenate the flags there and call it a day? Whenever you need the flags, you get it from the step output.

.github/workflows/ci.yml
mizvekov: If you build libcxx as a separate step pior to building LLVM, as we are doing here some times, then that necessarily means you are going to be using the host compiler to compile libcxx, unlike what happens in the unified build, where it builds clang first then uses that clang to build libcxx.

Now libcxx supports a limited set of compilers, it each newer version drops support for old compilers pretty fast.
Currently only clang-19 onwards and GCC 15 are supported.

alandefreitas: We're still talking past each other. I understand that only these compilers are supported and necessary, and I'm 100% OK with it. However, I expressed concern about multiple sources of truth, the lack of comments in this file, and how to explain the process to others. I don't see how these two things are related.

mizvekov: Ah okay, I had added the comment already, I thought you had seen that.

Yes, I agree the sanitizer builds add a bunch of complexity.

This is not just for MSan though, now we use the just built libcxx for the other sanitizers, which increases the effectiveness of the whole thing.

alandefreitas: > Ah okay, I had added the comment already, I thought you had seen that.

Oh... OK. Great. The issue is resolved then. :)

> Yes, I agree the sanitizer builds add a bunch of complexity.

Yes, they do.

> This is not just for MSan though, now we use the just built libcxx for the other sanitizers, which increases the effectiveness of the whole thing.

Oh... I had no idea other sanitizers needed (or were improved by) that, too. It's really hard to communicate all the details in this process.

alandefreitas: Please make sure the comments explain what happens in each case. When there are sanitizers, no sanitizers, etc, and how that happens in regards to the values being overridden in the presets.

.github/workflows/ci.yml
```diff
@@ -66,21 +66,24 @@ jobs:
             clang: git build-essential pkg-config python3 curl openjdk-11-jdk pkg-config libncurses-dev libxml2-utils libxml2-dev g++-14=14.2.0-4ubuntu2~24.04
             msvc: ''
           extra-values: |
-            llvm-hash: a1b6e7ff393533a5c4f3bdfd4efe5da106e2de2b
-            llvm-build-preset-prefix: {{#if optimized-debug}}debwithopt{{else}}{{{lowercase build-type}}}{{/if}}
+            use-libcxx: {{#if (and (ieq compiler 'clang') (or (eq major 19) (eq major 20) (eq major 21))) }}true{{else}}false{{/if}}
```
alandefreitas: I'm adding this gt feature now. That logic is going to break soon. Especially because it says nothing about sanitizers. It's only relying on the version.

.github/workflows/ci.yml
```diff
@@ -198,37 +201,57 @@ jobs:
...
-          llvm_root=$(pwd)/third-party/llvm-project/install
-          if [[ ${{ runner.os }} == 'Windows' ]]; then
-              llvm_root=$(echo "$llvm_root" | sed 's/\\/\//g')
-              llvm_root=$(echo $llvm_root | sed 's|^/d/|D:/|')
-              echo "$llvm_root"
-          fi
-          echo -E "llvm-root=$llvm_root" >> $GITHUB_OUTPUT
+          echo "deb http://apt.llvm.org/noble/ llvm-toolchain-noble-21 main" >> /etc/apt/sources.list
+          apt-get update
+          apt-get install -y libclang-18-dev libclang-19-dev libclang-20-dev libclang-21-dev
```
alandefreitas: We need all of these versions?

.github/workflows/ci.yml
```diff
@@ -198,37 +201,57 @@ jobs:
...
 
-      - name: Resolve Third-Party Directory
-        id: resolve-third-party-dir
+      - name: Resolved Matrix
```
alandefreitas, line 216: Please comments :) 

(not about "what"/ about "why")

.github/workflows/ci.yml
alandefreitas: We already have a step for installing packages and the matrix already sets the packages we need. Do we need to implement any feature in the actions to make this work there?

.github/workflows/ci.yml
```diff
@@ -66,21 +66,24 @@ jobs:
...
             llvm-archive-basename: llvm-{{{ lowercase os }}}-{{{ compiler }}}{{{ llvm-compiler-version }}}-{{{ llvm-build-preset-prefix }}}{{{ llvm-sanitizer }}}-{{{ substr llvm-hash 0 7 }}}
-            llvm-root: ../third-party/llvm-project/install
             llvm-archive-extension: {{#if (ieq os 'windows') }}7z{{else}}tar.bz2{{/if}}
             llvm-archive-filename: {{{ llvm-archive-basename }}}.{{{ llvm-archive-extension }}}
             llvm-sanitizer-config: {{#if (and (ne compiler 'clang') (ne compiler 'apple-clang'))}}{{else if ubsan}}Undefined{{else if asan}}Address{{else if msan}}MemoryWithOrigins{{/if}}
-            mrdocs-flags: {{#if (and (eq compiler 'gcc') (not asan)) }}-static{{/if}}{{#if (and (eq compiler 'clang') msan) }}-fsanitize-memory-track-origins{{/if}}
-            mrdocs-ccflags: {{{ ccflags }}} {{{ mrdocs-flags }}}
-            mrdocs-cxxflags: {{{ cxxflags }}} {{{ mrdocs-flags }}}
-            mrdocs-linkflags: {{#if asan }}-fsanitize=address{{/if}}
+            common-flags-base: {{#if (ieq compiler 'clang')}}-gz=zstd {{/if}}
```
alandefreitas, line 82: I can't find the comments we discussed.

.github/workflows/ci.yml
```diff
@@ -374,26 +456,23 @@ jobs:
...
-        uses: alandefreitas/cpp-actions/cmake-workflow@v1.8.10
+        uses: alandefreitas/cpp-actions/cmake-workflow@v1.8.11
         with:
           cmake-version: '>=3.26'
           cxxstd: ${{ matrix.cxxstd }}
           cc: ${{ steps.setup-cpp.outputs.cc || matrix.cc }}
           ccflags: ${{ matrix.mrdocs-ccflags }}
           cxx: ${{ steps.setup-cpp.outputs.cxx || matrix.cxx }}
-          cxxflags: ${{ matrix.mrdocs-cxxflags }}
+          cxxflags: ${{ steps.rmatrix.outputs.common-cxxflags }} ${{ matrix.mrdocs-flags }}
```
alandefreitas: Wasn't the new matrix step supposed to unify these things?

.github/workflows/ci.yml
```diff
@@ -932,8 +1011,22 @@ jobs:
       contents: write
 
     steps:
+      - name: Resolved Matrix
```
alandefreitas, line 1031: So maybe we could have a step to resolve paths. I think this already exists for a single path at a time, though. Or it  could be an action to take another dictionary and allow lots of operations on the values, like resolving paths, concatenating strings or whatever.

.github/workflows/ci.yml
mizvekov: Sure but the flag was only used once here, and it was a simple composition, it didn't feel worth creating a variable just for that.

src/lib/AST/TerminalTypeVisitor.hpp
```diff
@@ -239,7 +239,7 @@ class TerminalTypeVisitor
      */
     void
     buildTerminal(
-        NestedNameSpecifier const*,
+        NestedNameSpecifier,
```
alandefreitas, line 242: These changes are fixes MSan trigger? The calling code just worked unchanged? I smell something.

.github/workflows/ci.yml
alandefreitas: It just feels like `mrdocs-flags` lost its meaning. It's more about that than the effort or the duplication.

.github/workflows/ci.yml
mizvekov: Yes, there is a bug in how the llvm-dev package is split from the clang-dev only one, where some of the cmake module code errors out.

This happens when configuring the standalone libc++ build, where it tries to find the llvm cmake module, calls something into it, but that errors out because it can't find some cmake files which are only shipped in the clang-dev package.

.github/workflows/ci.yml
mizvekov: I think it makes sense to also use the split libcxx for the clang non-sanitizer version, because that would be a different test checking we build correctly with libc++ instead of libstdc++, and if there is any problems there we will both get the answer faster, and the isolated error without any sanitizer involvement.

.github/workflows/ci.yml
mizvekov: I think the main problem is that we need to insert the llvm repo here.

If the action adds support for that, then we can eliminate this.

.github/workflows/ci.yml
mizvekov: It's the big block of comments on top of the `Install libc++` action.

src/lib/AST/TerminalTypeVisitor.hpp
mizvekov: I missed those changes from the LLVM version bump I did a while ago.

I don't understand how the code compiled without those changes, but with the new jobs we started seeing linker errors because of this.

So yes, I think there is some deeper issue here, I just didn't look into it yet.

mizvekov:
> Well... bootstrap. :)

Yeah, as we talked about this, I assume there will be no problem if I post this on a separate PR, as these are orthogonal things we don't need to mix up in one commit.

.github/workflows/ci.yml
alandefreitas: If that's the case, then you don't even need to check `major`. You just have to check if it's `clang`.

.github/workflows/ci.yml
alandefreitas: Maybe it's because I don't understand the implications of this bug. However, I have no idea what the connection is between this bug and the answer to the question I asked about why we needed all these versions. Is it a bug in the CMake module that randomly gets fixed when you install multiple equally buggy versions simultaneously?

alandefreitas: Have you been able to check the documentation of the package-install action, or are you assuming it doesn't have this feature? IIRC, it does have this feature because we needed it before. We can also include it if necessary, as it would simply be a loop over some keys. But I believe the feature is already there.

.github/workflows/ci.yml
mizvekov: But that can't work with clang-18, as libc++ has already removed support for compiling with it.

.github/workflows/ci.yml
mizvekov: Ah, it's all the clang versions we compile libc++ with. Regarding versions, I'll remove clang-18 since it's unsupported anyway.

alandefreitas:
> An automated preview of the documentation is available at https://950.mrdocs.prtest2.cppalliance.org/index.html

@sdarwin, is it possible to tell the bot to only make this comment once per PR?

sdarwin:
@alandefreitas reviewing the available settings in Jenkins, there is a possibility:  "Ignore Pull Requests marked as Drafts".   Coincidentally, a pull request that's in very active development, getting 100 updates, is probably a "draft", right?  It's not ready to merge.  
Set this PR 950 to be a "draft" and lets see if previews stop.

sdarwin:
> It's not possible to turn PRs back into drafts 

It's now a draft.

Issue #963, opened 2025-08-10
Support the final keyword on class definitions.


jll63:
MrDocs drops the `final` keyword, when applied to a class. Example:

```c++
struct Foo final {};
```

Generated:

```
struct Foo;
```

`final` appears to be preserved when applied to virtual functions.

PR #965, opened 2025-08-12
feat: support the final specifier for classes and class templates, in AsciiDoc and HTML

gennaroprota:
Fixes issue #963.

### DIFF
--- a/share/mrdocs/addons/generator/common/partials/symbol/signature/record.hbs
+++ b/share/mrdocs/addons/generator/common/partials/symbol/signature/record.hbs
@@ -6,6 +6,7 @@
 {{~else~}}
     {{>symbol/name-text symbol ~}}
 {{~/if}}
+{{~#if isFinal}} final{{/if~}}
 {{#unless bases~}}
 {{else if isFriend~}}
 {{else}}
@@ -17,4 +18,11 @@
     {{#unless (eq access ../defaultAccess)}} {{access}}{{/unless~}}
     {{#if isVirtual}} virtual{{/if}} {{>type/declarator type~}}
 {{/each~}}
-{{/unless}}{{#isSeeBelow}} { /* see-below */ }{{/isSeeBelow}};
+{{/unless}}
+{{!-- Unless we are going to show a class body ("{ ... }"), don't add a
+      semicolon if the record is final, because something like `class A
+      final;` is ill-formed (`final` can't be used on a non-defining
+      declaration). Similarly, don't show a semicolon if there are base
+      classes. --}}
+{{~#isSeeBelow}} { /* see-below */ }{{/isSeeBelow~}}
+{{#if (or isSeeBelow (and (not isFinal) (eq (len bases) 0)))}};{{/if}}
--- a/docs/modules/ROOT/pages/generators.adoc
+++ b/docs/modules/ROOT/pages/generators.adoc
@@ -258,6 +258,10 @@ When the symbol kind is `record` (e.g., `class`, `struct`, `union`), the symbol 
 | `string`
 | The default access level of the record members (e.g., `public`, `private`).
 
+| `isFinal`
+| `bool`
+| Whether the record is final.
+
 | `isTypedef`
 | `bool`
 | Whether the record is a typedef.
--- a/include/mrdocs/Metadata/Info/Record.hpp
+++ b/include/mrdocs/Metadata/Info/Record.hpp
@@ -353,6 +353,7 @@ tag_invoke(
     tag_invoke(t, io, dynamic_cast<Info const&>(I), domCorpus);
     io.map("tag", I.KeyKind);
     io.map("defaultAccess", getDefaultAccessString(I.KeyKind));
+    io.map("isFinal", I.IsFinal);
     io.map("isTypedef", I.IsTypeDef);
     io.map("bases", dom::LazyArray(I.Bases, domCorpus));
     io.map("derived", dom::LazyArray(I.Derived, domCorpus));
--- a/test-files/golden-tests/config/auto-relates/derived.adoc
+++ b/test-files/golden-tests/config/auto-relates/derived.adoc
@@ -52,7 +52,7 @@ Declared in `&lt;derived&period;cpp&gt;`
 [source,cpp,subs="verbatim,replacements,macros,-callouts"]
 ----
 struct A
-    : link:#ABase[ABase];
+    : link:#ABase[ABase]
 ----
 
 === Base Classes
@@ -141,7 +141,7 @@ Declared in `&lt;derived&period;cpp&gt;`
 [source,cpp,subs="verbatim,replacements,macros,-callouts"]
 ----
 struct AView
-    : link:#ABase[ABase];
+    : link:#ABase[ABase]
 ----
 
 === Base Classes
@@ -194,7 +194,7 @@ Declared in `&lt;derived&period;cpp&gt;`
 [source,cpp,subs="verbatim,replacements,macros,-callouts"]
 ----
 struct AView2
-    : link:#AView[AView];
+    : link:#AView[AView]
 ----
 
 === Base Classes
--- a/test-files/golden-tests/config/auto-relates/derived.html
+++ b/test-files/golden-tests/config/auto-relates/derived.html
@@ -60,7 +60,7 @@
 Declared in <code>&lt;derived.cpp&gt;</code></div>
 <pre>
 <code class="source-code cpp">struct A
-    : <a href="#ABase">ABase</a>;
+    : <a href="#ABase">ABase</a>
 
 </code>
 </pre>
@@ -171,7 +171,7 @@ Declared in <code>&lt;derived.cpp&gt;</code></div>
 Declared in <code>&lt;derived.cpp&gt;</code></div>
 <pre>
 <code class="source-code cpp">struct AView
-    : <a href="#ABase">ABase</a>;
+    : <a href="#ABase">ABase</a>
 
 </code>
 </pre>
@@ -240,7 +240,7 @@ Declared in <code>&lt;derived.cpp&gt;</code></div>
 Declared in <code>&lt;derived.cpp&gt;</code></div>
 <pre>
 <code class="source-code cpp">struct AView2
-    : <a href="#AView">AView</a>;
+    : <a href="#AView">AView</a>
 
 </code>
 </pre>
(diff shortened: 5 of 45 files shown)


share/mrdocs/addons/generator/common/partials/symbol/signature/record.hbs
```diff
@@ -17,4 +17,7 @@
     {{#unless (eq access ../defaultAccess)}} {{access}}{{/unless~}}
     {{#if isVirtual}} virtual{{/if}} {{>type/declarator type~}}
 {{/each~}}
-{{/unless}}{{#isSeeBelow}} { /* see-below */ }{{/isSeeBelow}};
\ No newline at end of file
+{{!-- Don't show a semicolon if the record is final, because something
+      like `class A final;` is ill-formed (`final` can't be used on a
+      non-defining declaration). --}}
+{{/unless}}{{#isSeeBelow}} { /* see-below */ }{{/isSeeBelow}}{{#if isFinal}} final{{else}};{{/if}}
```
alandefreitas: That still leaves the case where there are base classes.

share/mrdocs/addons/generator/common/partials/symbol/signature/record.hbs
```diff
@@ -17,4 +18,11 @@
...
-{{/unless}}{{#isSeeBelow}} { /* see-below */ }{{/isSeeBelow}};
\ No newline at end of file
+{{/unless}}
+{{!-- Unless we are going to show a class body ("{ ... }"), don't add a
+      semicolon if the record is final, because something like `class A
+      final;` is ill-formed (`final` can't be used on a non-defining
+      declaration). Similarly, don't show a semicolon if there are base
+      classes. --}}
+{{~#unless (or isSeeBelow isFinal bases)}};{{/unless~}}
+{{~#isSeeBelow}} { /* see-below */ };{{/isSeeBelow~}}
```
alandefreitas:
```
{{/unless}}
{{!-- Unless we are going to show a class body ("{ ... }"), don't add a
      semicolon if the record is final, because something like `class A
      final;` is ill-formed (`final` can't be used on a non-defining
      declaration). Similarly, don't show a semicolon if there are base
      classes. --}}
{{~#unless (or isSeeBelow isFinal bases)}};{{/unless~}}
{{~#isSeeBelow}} { /* see-below */ };{{/isSeeBelow~}}
```

This is redundant because you're removing the `;` when it's see-below then reincluding it when it is see-below (the reinclusion being part of this PR).

The non-redundant equivalent code would be something like the same logic, but not removing the first line and without the unnecessary condition (because see-below does include the `;`):

```
{{/unless}}{{#isSeeBelow}} { /* see-below */ }{{/isSeeBelow}}; 
{{!-- Unless we are going to show a class body ("{ ... }"), don't add a
      semicolon if the record is final, because something like `class A
      final;` is ill-formed (`final` can't be used on a non-defining
      declaration). Similarly, don't show a semicolon if there are base
      classes. --}}
{{~#unless (or isFinal bases)}};{{/unless~}}
```

gennaroprota: Your code would add two semicolons for a non-final, see-below record without bases; e.g.:

```
class A { /* see-below */ };;
```

and if you remove the semicolon after "{ /* see-below */ }" in the Handlebars code, that produces no semicolon (e.g.) for this:

```
class A final { /* see-below */ }
```
.

alandefreitas: Yes. I incorrectly pasted it. My code was supposed to be:

```
{{/unless}}{{#isSeeBelow}} { /* see-below */ }{{/isSeeBelow}}
{{!-- Unless we are going to show a class body ("{ ... }"), don't add a
      semicolon if the record is final, because something like `class A
      final;` is ill-formed (`final` can't be used on a non-defining
      declaration). Similarly, don't show a semicolon if there are base
      classes. --}}
{{~#unless (or isFinal bases)}};{{/unless~}}
```

(it's to be *something like that* but it might obviously need some small indentation or change because I haven't tested it with all cases)

But this change is important because then things are in order, and the code reads as:

First condition) Render the code block if the code block is due
Second condition) Render the semicolon if the semicolon is due

instead of:

First condition) Render a semicolon that is due, but don't render a semicolon even if it's due if it's also a see-below symbol, because we're going to render this subcase of semicolon with the whole block later on
Second condition) Render the block now, even though we were already at the semicolon rendering stage one line above, but that's OK, but we put an exception there to ignore see-below, which is the only case where we need to render blocks for now.

src/lib/Support/Handlebars.cpp
```diff
@@ -3793,6 +3793,16 @@ block_helper_missing_fn(
     return {};
 }
 
+Expected<dom::Value>
+is_empty_fn(dom::Array const& arguments)
```
alandefreitas: Wait a minute... So this whole time, we were assuming this helper existed out of nowhere?

src/lib/Support/Handlebars.cpp
```diff
@@ -3804,6 +3814,7 @@ registerBuiltinHelpers(Handlebars& hbs)
     hbs.registerHelper("log", dom::makeVariadicInvocable(log_fn));
     hbs.registerHelper("helperMissing", dom::makeVariadicInvocable(helper_missing_fn));
     hbs.registerHelper("blockHelperMissing", dom::makeInvocable(block_helper_missing_fn));
+    hbs.registerHelper("isEmpty", dom::makeVariadicInvocable(is_empty_fn));
```
alandefreitas: `isEmpty` is not a handlebars builtin helper:

https://handlebarsjs.com/guide/builtin-helpers.html

It could be a custom container helper or something. (see `registerContainerHelpers`)
I'd be surprised if there's no container helper that does something similar, though.

Also, I'm not sure `isEmpty` does the same thing as https://handlebarsjs.com/api-reference/utilities.html#handlebars-utils-isempty-value. You'd have to check.

share/mrdocs/addons/generator/common/partials/symbol/signature/record.hbs
```diff
@@ -17,4 +18,11 @@
...
-{{/unless}}{{#isSeeBelow}} { /* see-below */ }{{/isSeeBelow}};
\ No newline at end of file
+{{/unless}}
+{{!-- Unless we are going to show a class body ("{ ... }"), don't add a
+      semicolon if the record is final, because something like `class A
+      final;` is ill-formed (`final` can't be used on a non-defining
+      declaration). Similarly, don't show a semicolon if there are base
+      classes. --}}
+{{~#isSeeBelow}} { /* see-below */ }{{/isSeeBelow~}}
+{{#if (or isSeeBelow (and (not isFinal) (isEmpty bases)))}};{{/if}}
```
alandefreitas: Additionally, it's essential to have a formal explanation for why `not bases` did not work. Because in general, the inverse conditional with `#unless` shouldn't work either. That's an important question.

src/lib/Support/Handlebars.cpp
gennaroprota: There was an `isEmpty()` function but no corresponding registered helper.

share/mrdocs/addons/generator/common/partials/symbol/signature/record.hbs
gennaroprota: `not bases` always gives `false` because of [this](https://github.com/cppalliance/mrdocs/blob/e35441053063d055b465c609b34b0ef34b627f64/src/lib/Dom/Value.cpp#L280). It doesn't matter whether the array is empty or not. Now, are you saying `{{#unless <condition>}}` gets internally translated to `{{#if (not (<condition>))}}`?

alandefreitas: > Now, are you saying {{#unless <condition>}} gets internally translated to {{#if (not (<condition>))}}?

That's not how it works internally, but yes, in the sense that they're logically equivalent.

src/lib/Support/Handlebars.cpp
alandefreitas: Great. So all we need to fix is the category of the helper. It can't be built-in because this is a specific list determined by the specification, and isEmpty is not in it. It should be included in the container helpers. Please review the casing used for helper names in that section.

share/mrdocs/addons/generator/common/partials/symbol/signature/record.hbs
alandefreitas: But we have a problem with this point. If you look at `if_fn` in the same file, you'll see `#if` uses `isEmpty` and not `isTruthy`. Other built-in helpers should follow suit.

alandefreitas: My hunch is the problem has nothing to do with `if` and `unless`. The problem is the `and`, `or`, and all other logical helpers. 

They're not built-in helpers. They came from the Antora default UI bundle. The comments should have references to their implementation: https://gitlab.com/antora/antora-ui-default/-/tree/master/src/helpers

Their implementation relies on `isTruthy`, and we replicated that when porting it. But they should really depend on `isEmpty` because a condition that always returns `true` or `false` is useless. Not only that, but it's confusing because it doesn't match the behavior of `#if` and `#unless`. This should be fixed in all Antora helpers.

gennaroprota: You want to change other built-in helpers to use `isEmpty()`?

src/lib/Support/Handlebars.cpp
gennaroprota: I have verified that `eq (len bases) 0` also works. Please tell me which solution you prefer.

share/mrdocs/addons/generator/common/partials/symbol/signature/record.hbs
alandefreitas: No. I said that:

1) All built-in helpers should already use isEmpty (not isTruthy)
2) All logical and antora helpers should use isEmpty instead of isTruthy

alandefreitas: I'm not happy that it returns true for `0` (that is, for all integers) though. Let's not change that for now. 

You could just use the helper to get the array size (`0` would decay to `false` with `isTruthy`), and everything else about handlebars in the PR becomes unnecessary.

This way, the problem is solved with minimal effort, and we can consider the isEmpty issue later.

PR #976, opened 2025-08-25
fix: bootstrap uses the directory of the latest includes available under `llvm_install_dir` for `--stdlib-includes`

gennaroprota:
I left the code as is, but we shouldn't hard-code directory names like that.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1678,6 +1678,7 @@ class MrDocsInstaller:
         else:
             print(f"\nMrDocs has been successfully installed in {self.options.mrdocs_install_dir}.\n")
 
+    @lru_cache(maxsize=1)
     def find_latest_clang_include_dir(self):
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -1678,6 +1678,16 @@ def install_mrdocs(self):
         else:
             print(f"\nMrDocs has been successfully installed in {self.options.mrdocs_install_dir}.\n")
 
+    def find_latest_clang_include_dir(self):
```
alandefreitas, line 1682: I had suggested the logic before the call so that we could reuse the work. If this is being passed into a function, then at least prefix it with `@lru_cache(maxsize=1)` so we can reuse the work. We don't want to be recalculating this for no reason. Filesystem calls are expensive, all things considered here (in other words, considering it costs nothing to do this only once).

PR #977, opened 2025-08-25
ci: enable ASan on GCC

mizvekov:


alandefreitas:
Should we close this PR?

mizvekov:
Yeah, supplanted by the Clang based sanitizers, I don't think we need this anymore.

PR #978, opened 2025-08-25
Accessibility improvements

2LoS:
This is just a little part of the overall work, but I'd like to receive a feedback before continuing. I'd also like to know which level of Accessibility the website should at least support, since, in some situations, this may also have an impact on the design. My idea is to keep the WCAG 2.2, Level AA. This should be a good trade-off between simplicity for development and a better UX for people with disabilities. Moreover, this should make the website conformant with the EAA (EN 301 549).

The `header-cta` links had "aria-label" attributes that essentially repeated what the information that can be retrieved by the text node. Also, the first link had a mismatch between what there was in the `aria-label` attribute ("Documentation") and what  there's in the text node ("Get Started").

The `menu-bar` unordered list heavily uses the `role` attribute. However, this doesn't seem necessary, since the semantics is already provided by the DOM. If the intention was to remove the semantics of `<li>` to replace it with `<a>`, I think the issue is more related to this [article](https://ivogomes.medium.com/navega%C3%A7%C3%A3o-sem-listas-um-teste-de-acessibilidade-b5d5876ac528).

The `role="navigation"` attribute was used in a `<nav>` tag, even if this is already provided by the tag itself. Indeed, the `<nav>` tag has implicitly the navigation role. If the intention was to keep it for compatibility with browsers that doesn't support HTML5, it should be at least applied consistently.

About the `star` section, there's already a text node, which says "Give us a Star on GitHub", and therefore the use of an `aria-label` attribute that essentially reports the same isn't useful.

Finally, the SVG code in the `logo.hbs` file wasn't hidden with the `aria-hidden` attribute.

### DIFF
--- a/docs/ui/src/partials/logo.hbs
+++ b/docs/ui/src/partials/logo.hbs
@@ -1,4 +1,4 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1080 1080">
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1080 1080" aria-hidden="true">
     <defs>
         <style>
             .st0{fill:#fff}
--- a/docs/website/index.html.hbs
+++ b/docs/website/index.html.hbs
@@ -36,7 +36,7 @@
 </head>
 
 <body>
-    <nav class="nav container-fluid" role="navigation" aria-label="Main navigation">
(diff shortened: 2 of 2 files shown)


rbbeeston (commented):
I don't see any issues here, all good stuff!

julioest:
LGTM!

julioest (approved):

PR #979, opened 2025-08-27
fix: copy website assets properly in CI

julioest:
- Inline logo styles to resolve landing page build dependencies
- Update CI workflow for correct asset copying
- Generate landing page website with all required styles


alandefreitas:
This is undoing some of the work from #978. Is that on purpose?

alandefreitas:
Actually, index.html shouldn't even be in the repo. This is an artifact. When did that happen? That has to go.

julioest:
I'll add that to the gitignore

alandefreitas:
The links and code blocks are very hard to read.

image

image

alandefreitas:
> I'll add that to the gitignore

Already did in some other PR just in case.

alandefreitas:
Let me know when this is good to go. :)

PR #1000, opened 2025-09-09
fix:input filter matches too much

### LINKED ISSUE #999: Input filter matches too much
An [input filter](https://mrdocs.com/docs/mrdocs/config-file.html#input_option) `"include/a"` will match any files in "include"'s subdirectory "a", but also on "abc" and any other subdirectory starting with "a".

K-ballo:
fixes #999


test-files/golden-tests/filters/file/subdirectories.yml
```diff
@@ -0,0 +1,3 @@
+# Include a (but not abc)
+input:
+  - './a/'
```
alandefreitas, line 3: Isn't `./a/` the case where it was working anyway? We need to test `./a/` and `./a`.

K-ballo: no, the trailing slash is removed when the path is resolved/normalized

i'll add another test case

K-ballo:
> In the commit message, I would describe what the new behavior is rather than what it is not or what it used to be.

you have a chance to modify the commit message when hitting "squash and merge"

src/test/lib/Support/Path.cpp
```diff
@@ -29,9 +29,45 @@ struct Path_test
     */
     }
 
+    void
+    testStartsWith()
+    {
+        using namespace files;
+
+        // empty
```
alandefreitas, line 37: Great work herre with the unit tests

PR #1004, opened 2025-09-10
refactor: cl arguments always map to regular arguments

K-ballo:
as discussed in #997

### DIFF
--- a/src/lib/MrDocsCompilationDatabase.cpp
+++ b/src/lib/MrDocsCompilationDatabase.cpp
@@ -300,18 +300,6 @@ adjustCommandLine(
         cmdLineCStrs.data() + cmdLineCStrs.size());
 
     // ------------------------------------------------------
-    // Get driver mode
-    // ------------------------------------------------------
-    // The driver mode distinguishes between clang/gcc and msvc
-    // command line option formats. The value is deduced from
-    // the `-drive-mode` option or from `progName`.
-    // Common values are "gcc", "g++", "cpp", "cl" and "flang".
-    StringRef const driver_mode = driver::getDriverMode(progName, cmdLineCStrs);
-    // Identify if we should use "msvc/clang-cl" or "clang/gcc" format
-    // for options.
-    bool const is_clang_cl = driver::IsClangCL(driver_mode);
-
-    // ------------------------------------------------------
     // Supress all warnings
     // ------------------------------------------------------
     // Add flags to ignore all warnings. Any options that
@@ -490,6 +478,18 @@ adjustCommandLine(
     {
       new_cmdline.emplace_back(std::format("-I{}", inc));
     }
+
+    // ------------------------------------------------------
+    // Get driver mode
+    // ------------------------------------------------------
+    // The driver mode distinguishes between clang/gcc and msvc
+    // command line option formats. The value is deduced from
+    // the `-drive-mode` option or from `progName`.
+    // Common values are "gcc", "g++", "cpp", "cl" and "flang".
+    StringRef const driver_mode = driver::getDriverMode(progName, cmdLineCStrs);
+    // Identify if we should use "msvc/clang-cl" or "clang/gcc" format
+    // for options.
+    bool const is_clang_cl = driver::IsClangCL(driver_mode);
 
     // ------------------------------------------------------
     // Adjust each argument in the command line
--- a/src/lib/MrDocsCompilationDatabase.cpp
+++ b/src/lib/MrDocsCompilationDatabase.cpp
@@ -224,6 +224,7 @@ isValidMrDocsOption(
              driver::options::OPT__SLASH_EH,
              driver::options::OPT__SLASH_GR,
              driver::options::OPT__SLASH_GR_,
+             driver::options::OPT__SLASH_M_Group,
 
             // input file options
             //driver::options::OPT_INPUT,
--- a/src/lib/MrDocsCompilationDatabase.cpp
+++ b/src/lib/MrDocsCompilationDatabase.cpp
@@ -402,9 +402,10 @@ adjustCommandLine(
         isExplicitCCompileCommand || (!isExplicitCppCompileCommand && isImplicitCSourceFile);
 
     constexpr auto is_std_option = [](std::string_view const opt) {
-        return opt.starts_with("-std=") || opt.starts_with("--std=") || opt.starts_with("/std:") || opt.starts_with("-std:");
+        return opt.starts_with("-std=") || opt.starts_with("--std=") ||
+            opt.starts_with("/std:") || opt.starts_with("-std:");
     };
-    if (std::ranges::find_if(cmdline, is_std_option) == cmdline.end())
+    if (std::ranges::none_of(cmdline, is_std_option))
     {
         if (!isCCompileCommand)
         {
@@ -440,7 +441,7 @@ adjustCommandLine(
             for (auto const& inc : it->second)
             {
               new_cmdline.emplace_back(std::format("-isystem{}", inc));
-            }          
+            }
         }
     }
 
--- a/src/lib/MrDocsCompilationDatabase.cpp
+++ b/src/lib/MrDocsCompilationDatabase.cpp
@@ -510,8 +510,6 @@ adjustCommandLine(
     while (idx < cmdline.size())
     {
         // Parse one argument as a Clang option
-        // ParseOneArg updates Index to the next argument to be parsed.
-        unsigned const idx0 = idx;
         std::unique_ptr<llvm::opt::Arg> arg =
             opts_table.ParseOneArg(args, idx, visibility);
         if (!isValidMrDocsOption(workingDir, arg))
--- a/src/lib/MrDocsCompilationDatabase.cpp
+++ b/src/lib/MrDocsCompilationDatabase.cpp
@@ -222,6 +222,8 @@ isValidMrDocsOption(
              driver::options::OPT_flang_ignored_w_Group,
              driver::options::OPT__SLASH_O,
              driver::options::OPT__SLASH_EH,
+             driver::options::OPT__SLASH_GR,
+             driver::options::OPT__SLASH_GR_,
 
             // input file options
             //driver::options::OPT_INPUT,
--- a/src/lib/MrDocsCompilationDatabase.cpp
+++ b/src/lib/MrDocsCompilationDatabase.cpp
@@ -525,7 +525,7 @@ adjustCommandLine(
             MRDOCS_ASSERT(arg->getNumValues() == 1);
 
             std::string_view v = arg->getValue();
-            if (v == "c++latest" || v == "c++preview")
+            if (v == "c++latest" || v == "c++23preview")
             {
                 new_cmdline.emplace_back("-std=c++23");
             }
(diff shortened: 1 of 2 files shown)


vinniefalco:
This will be squashed I hope?

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -288,7 +292,7 @@ adjustCommandLine(
     // Copy the compiler path
     // ------------------------------------------------------
     std::string const& progName = cmdline.front();
-    std::vector new_cmdline = {progName};
+    std::vector new_cmdline = {std::string{"clang"}};
```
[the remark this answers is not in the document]
K-ballo: nope, that would result in a vector of pointers

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -411,9 +403,10 @@ adjustCommandLine(
         isExplicitCCompileCommand || (!isExplicitCppCompileCommand && isImplicitCSourceFile);
 
     constexpr auto is_std_option = [](std::string_view const opt) {
-        return opt.starts_with("-std=") || opt.starts_with("--std=") || opt.starts_with("/std:");
+        return opt.starts_with("-std=") || opt.starts_with("--std=") ||
+            opt.starts_with("/std:") || opt.starts_with("-std:");
     };
-    if (std::ranges::find_if(cmdline, is_std_option) == cmdline.end())
+    if (std::ranges::none_of(cmdline, is_std_option))
```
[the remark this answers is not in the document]
K-ballo: ok?

alandefreitas:
This is great. These ternary operators with `is_clang_cl` were error-prone. 

> feat: drop msvc emulation mode #1004

This title is misleading to someone who doesn't understand the problem being fixed here because it sounds like MrDocs won't recognise compilation commands in the MSVC format. What's happening is just that all commands will be converted to the same format so that we can ensure the same behaviour and so that options included by MrDocs are less prone to errors. But that's internal and transparent to the user.

Also, this it's is more of a "refactor" than a "feat" because nothing is changing in terms of MrDocs expected behavior. It could also be a "fix" in the sense that (I think) it's fixing an issue with nostdinc in passing.

> @alandefreitas please advice on how to test this (so that we are informed when the tools stops working on Windows)

What I would do is upsert a test suite for MrDocsCompilationDatabase (I don't know if it already exists) and then compare the input and output. Things should remain the same when it's not MSVC mode, and things should change when it's in MSVC mode. I wouldn't touch the golden tests at all.

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -288,7 +294,7 @@ adjustCommandLine(
     // Copy the compiler path
     // ------------------------------------------------------
     std::string const& progName = cmdline.front();
-    std::vector new_cmdline = {progName};
+    std::vector new_cmdline = {std::string{"clang"}};
```
alandefreitas, line 297: Nice. :)

That information was so annoying. The only thing it affected was the driver::getDriverMode function.

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -300,24 +306,12 @@ adjustCommandLine(
...
-    // for options.
-    bool const is_clang_cl = driver::IsClangCL(driver_mode);
-
     // ------------------------------------------------------
     // Supress all warnings
     // ------------------------------------------------------
     // Add flags to ignore all warnings. Any options that
     // affect warnings will be discarded later.
-    new_cmdline.emplace_back(is_clang_cl ? "/w" : "-w");
+    new_cmdline.emplace_back("-w");
```
alandefreitas, line 314: That's also great. This `is_clang_cl ?` is prone to errors (especially in the sense that we might get it wrong and get different behavior in the output).

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -411,9 +405,10 @@ adjustCommandLine(
         isExplicitCCompileCommand || (!isExplicitCppCompileCommand && isImplicitCSourceFile);
 
     constexpr auto is_std_option = [](std::string_view const opt) {
-        return opt.starts_with("-std=") || opt.starts_with("--std=") || opt.starts_with("/std:");
+        return opt.starts_with("-std=") || opt.starts_with("--std=") || // clang options
+            opt.starts_with("/std:") || opt.starts_with("-std:"); // clang-cl options
     };
-    if (std::ranges::find_if(cmdline, is_std_option) == cmdline.end())
+    if (std::ranges::none_of(cmdline, is_std_option))
```
alandefreitas, line 411: :)

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -492,6 +487,18 @@ adjustCommandLine(
...
+    // ------------------------------------------------------
+    // The driver mode distinguishes between clang/gcc and msvc
+    // command line option formats. The value is deduced from
+    // the `-drive-mode` option or from `progName`.
+    // Common values are "gcc", "g++", "cpp", "cl" and "flang".
+    StringRef const driver_mode = driver::getDriverMode(progName, cmdLineCStrs);
+    // Identify if we should use "msvc/clang-cl" or "clang/gcc" format
+    // for options.
+    bool const is_clang_cl = driver::IsClangCL(driver_mode);
+
```
alandefreitas, line 501: This logic being moved down here is also great. It limits errors that could go up there.

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -507,18 +514,40 @@ adjustCommandLine(
...
         if (!isValidMrDocsOption(workingDir, arg))
         {
             continue;
         }
-        new_cmdline.insert(
-            new_cmdline.end(),
-            cmdline.begin() + idx0,
-            cmdline.begin() + idx);
+
+        // Translate clang-cl /std: into clang -std=
```
alandefreitas, line 524: So `arg->render()` works for all kinds of arguments except `-std` (not that bad, but it would be good to include a small comment)? Or does that mean `arg->render()` requires custom logic for lots of kinds of arguments, but the only one for which we got custom logic is `-std` (much worse)?

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -507,18 +514,40 @@ adjustCommandLine(
...
+                new_cmdline.emplace_back("-std=c++23");
+            }
+            else // c++14,c++17,c++20,c11,c17
+            {
+                new_cmdline.emplace_back(std::format("-std={}", v));
+            }
+            continue;
+        }
+
+        // Append the translated arguments to the new command line
```
alandefreitas, line 543: This is really nice. Clang just translates the whole thing. :)

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -507,18 +514,40 @@ adjustCommandLine(
...
+
+        // Translate clang-cl /std: into clang -std=
+        const llvm::opt::Option opt =
+            arg->getOption().getUnaliasedOption();
+        if (opt.matches(clang::driver::options::OPT__SLASH_std))
+        {
+            MRDOCS_ASSERT(arg->getNumValues() == 1);
+
+            std::string_view v = arg->getValue();
+            if (v == "c++latest" || v == "c++23preview")
```
alandefreitas, line 532: This is cumbersome. Since we have access to the whole Clang library, I'm sure there should be some way to know what the highest std is. Even if it's some enum of standards somewhere, we can evaluate an inline constexpr requires to check if STD_NUMBER exists in that enum.

A few ways to get the highest stardard could be:

```
#include "clang/Frontend/LangStandards.def"
```
 
or

```
clang::getLangStandardForName("c++23")
```

or

```
bool ok = clang::CompilerInvocation::CreateFromArgs(
    CI,
    {"-std=c++23"},
    *Diags
);
```

there must be any other ways.

These functions only give you if one specific standard is there, but you can just iterate 23 + X * 3 until you find the highest valid number, potentially at compile time.

src/test/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -0,0 +1,184 @@
...
+#include <mrdocs/Support/ThreadPool.hpp>
+#include <test_suite/test_suite.hpp>
+
+namespace clang {
+namespace mrdocs {
+
+namespace {
+/** Compilation database for a single .cpp file.
+*/
+class SingleFileTestDB
```
alandefreitas, line 24: Can't you can just use the existing one?

src/test/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -0,0 +1,184 @@
...
+        testClangCLIncludeFlag();
+    }
+
+    void run()
+    {
+        testClangCL();
+    }
+};
+
+TEST_SUITE(
```
alandefreitas, line 179: Your test suite is great. You had nothing to worry about.

src/test/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -0,0 +1,184 @@
...
+        cc.Heuristic = "unit test";
+
+        // Create an adjusted MrDocsDatabase
+        std::unordered_map<std::string, std::vector<std::string>> defaultIncludePaths;
+        MrDocsCompilationDatabase compilations(
+            llvm::StringRef(cc.Directory), SingleFileTestDB(cc), config_, defaultIncludePaths);
+        return compilations.getCompileCommands(cc.Filename).front().CommandLine;
+    }
+    
+    static bool has(std::vector<std::string> const& commandLine, std::string_view flag)
```
alandefreitas, line 84: We already have a `contains` function in `Algorithm.hpp` IIRC.

src/lib/MrDocsCompilationDatabase.cpp
K-ballo: The later, not all arguments can be automatically translated.

src/test/lib/MrDocsCompilationDatabase.cpp
K-ballo: No, the existing one has hard coded arguments.

src/lib/MrDocsCompilationDatabase.cpp
alandefreitas: Oh.... OK.

> requires custom logic for lots of kinds of arguments

So what are we going to do about this? Do we know the kinds of arguments that can't be translated? Do we know what happens when there's one of them? Do we know how to handle them?

K-ballo: Whenever an argument can't be translated clang will emit an error. The argument then has to either be added to the ignore list defined inside `isValidMrDocsOption`, or custom handling logic has to be added here.

src/lib/MrDocsCompilationDatabase.cpp
K-ballo: No, this affects the driver clang uses, which in turn affects things like command line options and option styles, defaulted flags, system includes, etc.

src/lib/MrDocsCompilationDatabase.cpp
K-ballo: This particular case was unnecessary, both clang and cl accept arguments in `-` and `/` forms, despite having different preferences.

PR #1013, opened 2025-09-12
Fix 1012

K-ballo:


alandefreitas (commented):
This is interesting. There are no links or descriptions but I reversed engineered the code and I believe it's a refactor to just pass the whole object and the desired tranche instead of a reference to the trench? This probably conflicts with your other PR where we fix the inheritances where the derived classes overrides it so we could probably wait on this one so we can evaluate the rebased version.

PR #1026, opened 2025-09-15
Fix 1025

### LINKED ISSUE #1025: Base members and shadowing
EDIT: using declarations are not necessary

Given

```
struct X {
void foo();
void bar();
};

struct Y : X {
    void foo();
protected:
    void bar();
};
```

The reference for `Y` includes `X::bar` as a (public) member function, despite it being hidden by `Y::bar`. It correctly hides `X::foo`.

The code that handles shadowing (name hiding) only considers derived members within one trenche.

https://github.com/cppalliance/mrdocs/blob/4a24503396284e05b727b1648e6e6bc07efd9dd6/src/lib/Metadata/Finalizers/BaseMembersFinalizer.cpp#L117

K-ballo:
fixes #1025


K-ballo:
Note: `extract-private` is necessary or else the corpus doesn't have the required information to properly resolve name hiding

src/lib/Metadata/Finalizers/BaseMembersFinalizer.cpp
```diff
@@ -30,53 +30,57 @@ inheritBaseMembers(
     RecordInterface const& base,
     AccessKind const A)
 {
+    auto&& members = allMembers(derived);
```
alandefreitas: Isn't `members` a value type with const-reference semantics? Is this a reference to a temporary we're relying on the compiler to keep alive?

If so, wouldn't 

```
auto members = allMembers(derived);
```

mean the same thing here since this is just a view?

src/lib/Metadata/Finalizers/BaseMembersFinalizer.cpp
```diff
@@ -30,53 +30,57 @@ inheritBaseMembers(
...
     if (A == AccessKind::Public)
     {
         // When a class uses public member access specifier to derive from a
         // base, all public members of the base class are accessible as public
         // members of the derived class and all protected members of the base
         // class are accessible as protected members of the derived class.
         // Private members of the base are never accessible unless friended.
-        inheritBaseMembers(derivedId, derived.Public, base.Public);
-        inheritBaseMembers(derivedId, derived.Protected, base.Protected);
+        inheritBaseMembers(derivedId, derived.Public, base.Public, members);
```
alandefreitas, line 42: Is this safe? `members` is a const-view of the derived members, but the function is changing the set of members. It's OK to ask for all members there again, but this is supposed to be a lazy operation (unless there's some bug somewhere else).

src/lib/Metadata/Finalizers/BaseMembersFinalizer.cpp
```diff
@@ -116,7 +122,7 @@ inheritBaseMembers(
 
         // Check if derived class has a member that shadows the base member
         auto shadowIt = std::ranges::find_if(
```
alandefreitas, line 124: Oh... Here's the relevant point. Yes. We can just call "allMembers" here to get all members of derived. This should be a lazy operation so there's no extra non-constant cost in just calling it again.

The only unnecessary extra-cost we have here is the cost of iterating all kinds of members while we want the functions in different access modes. We can fix that by just creating a lazy range that only contains the functions, or by iterating the functions for each access mode.

alandefreitas:
This is great. 

I just checked how the example in #1025 behaves:

https://godbolt.org/z/PbfEsTq5b

`Y` makes `bar()` protected, and the inherited version is not there. The code we have right now wouldn't check this interaction. And `foo()` just works as expected.

Something to think about is the use of the term' shadowing.' I don't know if there's a good solution to that, but there's another issue: the relationship between these inherited members and using decl shadows. We've been using the same term for two different things. When we work on this other issue, we would be doing it in the code and it could be confusing.

You probably already did this, but please also check if there's any interaction between this solution and the option regarding how these members should be inherited (`config->inheritBaseMembers`).

src/lib/Metadata/Finalizers/BaseMembersFinalizer.cpp
K-ballo: All members can hide the name of a base member, not just function members.

PR #1028, opened 2025-09-17
Use a distinct include guard in MrDocsSettingsDB.hpp

gennaroprota:
That file used the same include guard macro as SingleFileDB.hpp.


alandefreitas:
Ooops...

PR #1041, opened 2025-09-30
fix(ci): setup C++ when building releases

K-ballo:
Setup C++ in releases, same as used for building. Note this causes windows releases to use msvc rather than mingw.

If supporting mingw is explicitly desired then I suggest giving it its own distinct build where it is used throughout the process, rather than building with msvc but testing against mingw.


K-ballo:
> I'll wait for CI :)

The windows failure reflects the fact that the tool does not work in conjunction with MSVC.

This is now blocked by #997

PR #1042, opened 2025-09-30
feat: use system libs by default

K-ballo:


test-files/golden-tests/symbols/function/sfinae.yml
```diff
@@ -1,2 +1,4 @@
 warn-if-doc-error: false
 warn-no-paramdoc: false
+exclude-symbols:
+  - 'std::**'
```
mizvekov: Nit: missing LF

test-files/golden-tests/snippets/sqrt.yml
```diff
@@ -0,0 +1,4 @@
+warn-if-doc-error: false
+warn-no-paramdoc: false
+exclude-symbols:
+  - 'std::**'
```
mizvekov: Nit: missing LF

test-files/golden-tests/core/libcxx.yml
```diff
@@ -0,0 +1,2 @@
+use-system-stdlib: false
+use-system-libc: false
```
mizvekov: Nit: missing LF

test-files/golden-tests/config/sfinae/return-based.yml
```diff
@@ -0,0 +1,4 @@
+warn-if-doc-error: false
+warn-no-paramdoc: false
+exclude-symbols:
+  - 'std::**'
```
mizvekov: Nit: missing LF

test-files/golden-tests/config/sfinae/redeclare.yml
```diff
@@ -0,0 +1,4 @@
+warn-if-doc-error: false
+warn-no-paramdoc: false
+exclude-symbols:
+  - 'std::**'
```
mizvekov: Nit: missing LF

test-files/golden-tests/config/sfinae/redeclare.cpp
```diff
@@ -1,6 +1,28 @@
...
+
+template <typename T>
+struct is_class
+{
+    static constexpr bool value = true;
+};
+
+template <typename T>
+bool is_class_v = is_class<T>::value;
+}
```
mizvekov: Can we add these to some headers instead, so that we can just reuse them across tests? Ditto for other instances.

K-ballo: added common header in an implicit include directory, is this what you had in mind?

mizvekov: Yes, thanks!

mizvekov (approved):
LGTM

src/test/TestRunner.cpp
```diff
@@ -114,7 +114,9 @@ handleFile(
 
     // Create an adjusted MrDocsDatabase
     auto parentDir = files::getParentDir(filePath);
-    std::unordered_map<std::string, std::vector<std::string>> defaultIncludePaths;
+    std::unordered_map<std::string, std::vector<std::string>> defaultIncludePaths = {
+      {"clang", {std::format("{}/include", dirs_.cwd)}}
```
K-ballo: this is not the right path when executing a single test

PR #1045, opened 2025-10-01
fix: support clang-cl

K-ballo:


src/lib/CorpusImpl.cpp
```diff
@@ -711,6 +713,41 @@ lookupCacheSet(
 
 //------------------------------------------------
 
+namespace {
+// FIXME(k-ballo): MrDocsCompilationDatabase already figured this out, but we lost that information.
+// do we want to have to recompute it here? if so, dedup this query
```
K-ballo: should `CorpusImpl::build` take a `MrDocsCompilationDatabase` instead of any `CompilationDatabase`?

src/test/TestRunner.cpp
```diff
@@ -114,9 +148,15 @@ handleFile(
 
     // Create an adjusted MrDocsDatabase
     auto parentDir = files::getParentDir(filePath);
+    SingleFileDB const db =
+#if defined(WIN32)
+        makeSingleFileDBForClang(filePath);
+#else
+        makeSingleFileDBForClangCL(filePath);
+#endif
```
K-ballo: this is still arbitrary.. there's no reason why both drivers couldn't be tested on every platform

K-ballo: furthermore, this triggers unused function warnings

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -448,8 +449,9 @@ adjustCommandLine(
             it != implicitIncludeDirectories.end()) {
             for (auto const& inc : it->second)
             {
-              new_cmdline.emplace_back(std::format("-isystem{}", inc));
-            }          
+              new_cmdline.emplace_back(is_clang_cl ? "-external:I" : "-isystem");
```
mizvekov: We should avoid this duplication here, just decide what's the spelling of the flag once and put it in a variable.

src/lib/CorpusImpl.cpp
mizvekov: Yeah, I'd prefer to avoid duplicating this in this PR.

src/test/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    void testClang()
+    {
+        std::string const programName = "clang";
+
+        {
+            auto adjusted = adjustCompileCommand({ programName });
+            BOOST_TEST(has(adjusted, "-std=c++23"));
+        }
```
mizvekov, line 85: I think this would look and work better if we just separated them into each their own test cases.

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -411,17 +411,18 @@ adjustCommandLine(
...
+        return opt.starts_with("-std=") || opt.starts_with("--std=") || // clang options
+            opt.starts_with("-std:") || opt.starts_with("/std:"); // clang-cl options
     };
-    if (std::ranges::find_if(cmdline, is_std_option) == cmdline.end())
+    if (std::ranges::none_of(cmdline, is_std_option))
     {
         if (!isCCompileCommand)
         {
-            new_cmdline.emplace_back("-std=c++23");
+            new_cmdline.emplace_back(is_clang_cl ? "-std:c++latest" : "-std=c++23");
```
alandefreitas: Don't we have a stable equivalent to `latest` in the non-CL version?

src/lib/CorpusImpl.cpp
alandefreitas: > should CorpusImpl::build take a MrDocsCompilationDatabase instead of any CompilationDatabase?

That sounds good, because the public API (Corpus.hpp) doesn't expose the build function anyway. Another option, if you want to be safe, is to move this functionality to ClangHelpers.hpp.

src/test/TestRunner.cpp
```diff
@@ -65,6 +66,39 @@ replaceCRLFWithLF(std::string &str)
...
+    tooling::CompileCommand cc(
+        parentDir,
+        fileName,
+        std::move(cmds),
+        parentDir);
+    cc.Heuristic = "unit test";
+    return SingleFileDB(std::move(cc));
+}
+
+SingleFileDB makeSingleFileDBForClangCL(llvm::StringRef pathName)
```
alandefreitas: Oh... This looks great. There's a warning because it's not being used, but the logic is great for tests.

src/test/TestRunner.cpp
alandefreitas: > this is still arbitrary

:)

src/lib/MrDocsCompilationDatabase.cpp
K-ballo: no https://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-std

alandefreitas: Ah... OK. I see. You can calculate it with something like:

```cpp
#include <clang/Basic/LangStandard.h>
#include <llvm/ADT/StringRef.h>
#include <string>

/// ...

static
std::string 
latest_cxx_std_supported() 
{
  auto supports = [](llvm::StringRef name) {
    const clang::LangStandard* LS = clang::LangStandard::getLangStandardForName(name);
    return LS && LS->isCPlusPlus();
  };

  std::string cur = "c++23";
  std::string last_ok = "c++23";

  auto bump_by_3 = [](std::string& s) {
    // s is "c++NN" with two decimal digits.
    int n = (s[3] - '0') * 10 + (s[4] - '0');
    n += 3;
    s[3] = char('0' + (n / 10) % 10);
    s[4] = char('0' + (n % 10));
  };

  for (;;) {
    if (!supports(cur)) break;
    last_ok = cur;
    bump_by_3(cur);
  }
  return last_ok;
}
```

It also has the benefit of giving us the same version for CL and non-CL arguments in case the latest happens to have any special meaning.

src/test/TestRunner.cpp
K-ballo: would you suggest we run every test twice?
or should I rather conditionally define the one we're actually testing?

src/lib/MrDocsCompilationDatabase.cpp
K-ballo: We have complete control over clang, we can adjust this value manually when bumping llvm. If the target commit hash is exposed (`-DLLVM_COMMIT_HASH=...`) we can even static assert on this, so it doesn't go undetected.

src/test/TestRunner.cpp
alandefreitas: > would you suggest we run every test twice?

Sure. All platforms must be able to read both file formats, so it makes sense to test both formats on all platforms.

src/lib/MrDocsCompilationDatabase.cpp
alandefreitas: > we can adjust this value manually when bumping llvm

Yes. We could. As you noticed, the problem is remembering. MrDocs has dozens, if not hundreds, of moving parts reflecting information back and forth. We always forget to update all these parts because no one remembers to do so. That's why things like the XML generator are broken.

> we can even static assert on this

Exactly. For each type of information, we need a strategy that either forces us to review it because it's broken or that does it automatically. 

Your solution is in the first category, and I do like it. There's no LLVM_COMMIT_HASH in their headers, but I'm guessing you would end up with something like:

```
#include <llvm/Config/llvm-config.h>

// ...

template <size_t N, size_t M>
consteval
bool
str_equal(const char (&a)[N], const char (&b)[M]) {
    if (N != M) return false;
    for (size_t i = 0; i < N; ++i)
        if (a[i] != b[i]) return false;
    return true;
}

static_assert(str_equal(LLVM_VERSION_STRING, "22.0.0git"),
              "This code was designed with 22.0.0git in mind."
              "Since you're moving to " LLVM_VERSION_STRING ", please: "
              "1) review the code below and fix the c++ version if necessary"
              "2) update the string in this assertion");
```

or

```
#include <llvm/Config/llvm-config.h>

// ...

static_assert(std::string_view(LLVM_VERSION_STRING) == "22.0.0git",
              "This code was designed with 22.0.0git in mind."
              "Since you're moving to " LLVM_VERSION_STRING ", please: "
              "1) review the code below and fix the c++ version if necessary"
              "2) update the string in this assertion");
```

I'm OK with some variant of that. All we have to be sure is it's always working. Not something the user can bypass by mistake.

By default, I usually prefer things in the second category whenever possible, but this hasn't been the case for every problem.  For instance, if you look at the code I wrote, you'll see that I use `"c++23"` as the baseline. The idea is that there's no cost if we forget to update it, but we wouldn't lose anything in terms of correctness if we did. There's probably some smart solution based on some enum or something somewhere.

Anyway, your macro based solution is also fine to me. I'm just explaining the rationale here.

K-ballo: this is now #1047

src/test/TestRunner.cpp
K-ballo: ~I had to change the program name from "clang-cl" to just "cl", otherwise clang wasn't recognizing "/std:c++latest" on linux/macos. I do not have an explanation for this, why would it depend on the environment? @mizvekov ?~

https://github.com/llvm/llvm-project/blob/4d32ea87673adaa5252545ca82fa6cd58b134ff9/clang/lib/Driver/ToolChain.cpp#L439

src/lib/MrDocsCompilationDatabase.cpp
alandefreitas: #1047 says other options are "simpler". Do you mean more robust, instead? Because if they're even simpler than copy/pasting that thing, why not just do it now? Does @mizvekov know of a good solution to this?

K-ballo: I mean simpler. I do agree they're more robust too.

alandefreitas: If they're even simpler than copy/pasting that thing, why not just do it now?

K-ballo: This PR fixes a fundamental bug in the tool, the suggested change introduces an unrelated change of behavior (in a different part of the tool), conflating them under a single PR is a poor engineering practice.

K-ballo:
I just found out about [clang-cl's `/clang:` option](https://github.com/llvm/llvm-project/blob/550b2ef041ba16ee8b5f55b5f2307f501b2c15a0/clang/include/clang/Driver/Options.td#L9278-L9279). Using it still requires conditionals, and it makes working with multiple token options difficult, but we could in principle do:

```
new_cmdline.emplace_back(is_clang_cl ? "/clang:-std=c++23" : "-std=c++23");
new_cmdline.emplace_back(is_clang_cl ? std::format("/clang:-isystem=\"{}\"", inc) : std::format("-isystem=\"{}\"", inc));
```

The benefit for these downsides is that mrdocs would not need to know about msvc flags. Thoughts?

src/test/TestRunner.cpp
K-ballo: This didn't work out either. Is it possible to use CL mode on linux/macox? Does it require explicit `--driver-mode`?

src/lib/MrDocsCompilationDatabase.cpp
alandefreitas: OK. No problem.

My intuition was that

```
            new_cmdline.emplace_back(is_clang_cl ? "-std:c++latest" : "-std=c++23");
```

makes the std option vary between "latest" and  "c++23" for an arbitrary reason. So *this* looked like an "unrelated change of behavior" introduced by the PR, even though I don't question your "engineering practices".

To which you would probably respond that the previous behavior (rather than the "expected" behavior) wasn't working anyway, so it can't be a change of behavior.

I'll resolve this conversation then and we can avoid this loop. Could you open an issue for that?

alandefreitas:
> Thoughts?

That's a fantastic discovery! 

In most cases, things would get much more readable because we would be expressing directly in code that we explicitly want both options in the ternary operator to represent the same thing.

We can even write a lambda or something that optionally wraps the flags with it, and the code would get much simpler.

mizvekov:
Oh yes I knew about the `/clang:` redirection flag, sorry that I didn't think to bring it up.

Using the redirection flag skips any driver specific behavior for that flag, the driver simply forwards it to another driver.

This could be, as one simple example, checking that one flag is compatible with another one.

I'd say this is best used for flags which don't have a specific counterpart in the cl driver.

I suppose it's not a big deal if we use it with a limited set of flags and use cases we have vetted are ok, but then this wouldn't be a big advantage versus simply not using it.

alandefreitas:
Oh... OK. So it's best to reserve this for flags with no counterpart. Or maybe flags with no obvious counterpart, at least.

src/lib/MrDocsCompilationDatabase.cpp
K-ballo: Good point, I'll change it to `-c++23preview`.

The issue is #1047

alandefreitas: > Good point, I'll change it to -c++23preview.

Oh... Interesting. I thought it'd be just "-c++23" for both but that's great. I'm closing this.

src/test/TestRunner.cpp
alandefreitas: @mizvekov Ideas?

alandefreitas:
> @alandefreitas @mizvekov what's the procedure?

My procedure is to ask @mizvekov so you already did my procedure :)

K-ballo:
> I'd try to preprocess and isolate it first.

@mizvekov does that mean the clang-cl fix will be on hold until the clang leak is fixed upstream and mrdocs bumps its target llvm commit?

mizvekov:
Yeah, but you could simply not commit the libc++ tests at this point, send that to a separate PR and lets fix it there.

The problem is pre-existing and only happens on new tests, so I don't see a problem. @alandefreitas ?

K-ballo:
> The problem is pre-existing and only happens on new tests,

The problem happens on existing tests, when run under a new configuration. So I take it the way to proceed is to leave clang-cl tests out on non-windows platforms.

alandefreitas:
> so I don't see a problem. @alandefreitas ?

In general, I agree with the "feeling", so to speak. But at the same time, unblocking PRs is not our end goal in itself. We need to have clarity about the conditions under which we're seeing the bug, including which tests it appears in, what came before the PR, what's causing it, and how complex it would be to fix. With this information, we can make an informed decision on what's more beneficial to the project. It could be just fixing it or opening an independent issue that "unblocks" the PR.

mizvekov:
This was finished in https://github.com/cppalliance/mrdocs/pull/1069

PR #1046, opened 2025-10-01
ci: improve the 'Publish website' step

### LINKED ISSUE #1039: ci: compress everything and deflate it on the server
[Uploading the website](https://github.com/cppalliance/mrdocs/actions/runs/18111852278/job/51541092312) is now taking forever…  

There’s more than the website on this server. The LLVM releases are also there. Still, storage and processing must be unrelated because the CPU is not busy, and because the LLVM release uploads (which are much larger) do not take this long.

This is likely because the website consists of numerous small pages. This tends to confuse servers. We should compress everything and then deflate it on the server.

mizvekov:
This makes that step use rsync to upload the website and demos.

Besides supporting compression, rsync also only sends the changes
over the wire.

This improves the performance of that step by a wide margin, reducing
normal upload times from almost 9 minutes down to about 20s for the most
common changes.

Note we have seen some times this step take almost two hours, but it seems
circumstantial with unknown cause, and we are not sure what kind of improvement
this would have in that case.

This uses rsync's ability to also remove stale files, to make sure
we don't leave old files behind. (This ignores the directories which are
handled separately, like llvm+clang and demos).

There were some pretty stale things left behind that this now removes,
such as a sitemap.xml which was generated last year.

Fixes #1039

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -581,7 +581,7 @@ jobs:
         uses: alandefreitas/cpp-actions/package-install@v1.8.12
         id: package-install
         with:
-          apt-get: build-essential asciidoctor cmake bzip2 git
+          apt-get: build-essential asciidoctor cmake bzip2 git rsync
 
       - name: Clone MrDocs
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


alandefreitas:
👏

K-ballo (approved):
LGTM

sdarwin (approved):

Issue #1048, opened 2025-10-03
Keep record members in the corpus


K-ballo:
- keep a list of members in `RecordInfo`: `vector< tuple< Access, InfoKind, SymbolID > > Members; ` (that is part of the corpus),
- preserve order of declared members

~~note some transformations like overload folding will have to happen in two places (or maybe the record interface should be computed later, after all transformations).~~

compute the record interface (which is not part of the corpus) once all transformations have finalized

vinniefalco:
This issue describes a proposed change to how metadata is stored. It doesn't immediately strike me as a bad idea, yet these types of architectural changes would be easier to understand and digest if they provided the motivation. Why do we want to change it to this?

alandefreitas:
> This is not necessarily the order in which clang visits them. If we need to sort by line-number to reflect the actual C++ order then that's what we'll do.

That's great. So that's already implemented. So ~preserve order of declared members~ is done and the issue we have left is "replace record tranches with vector of tuples". What problem are we tackling with this change?

vinniefalco:
Not so fast. Sorting by line number is a nice hack but what happens with:
```
struct T {
  void f0(); void f1(); void f2();
};
```

`f0`, `f1`, and `f2` all have the same line number. What is the order?

PR #1049, opened 2025-10-03
Add a requirements section to usage.adoc

gennaroprota:


docs/modules/ROOT/pages/usage.adoc
```diff
@@ -195,6 +195,11 @@ compilation-database: ../CMakeLists.txt
 output: ../docs/reference
 ----
 
+== Requirements
+
+MrDocs requires input to be well-formed C++ code. If the input is ill-formed, MrDocs will attempt to emit at least one diagnostic. However, constructs classified
```
ashtum: I think AsciiDoc files have issues with `C++` strings (they emit the plus signs), so we have to use {cpp} instead.

gennaroprota: Indeed. I had noticed in the docs preview that C++ had become C without classes again :-). Thanks, I'll fix it right away.

alandefreitas:
Isn't that information like everywhere in the documentation over and over again already?

alandefreitas:
@vinniefalco That's the pattern I was attempting to communicate earlier today, which also just happened in #1048 and many others: a design discussion about problem X, where solution Y is proposed, is (reasonably) interpreted as an *order* to implement Y regardless of X. 

The issues and PRs don't even bother to mention what problem X they're attempting to solve, and any attempt to even understand what's happening is futile because "Vinnie told me so" is, to be fair, a somewhat reasonable answer.

alandefreitas:
I completely agree the requirements should be clear to user. It's extremely important for users to understand this difference from doxygen as soon as possible. In fact, this content should probably move to the beginning of this page or be exposed much sooner.

However, the problem I was previously mentioning with this issue is not the that requirements should not be listed. The problems I mentioned were:

1) The form the section is included in the page: In this particular issue is the requirement list has a single requirement and this very page already includes a gigantic section about this requirement. This PR completely ignores what already exists. The gigantic section is already explaining not only this requirement but also includes all common strategies in MrDocs to deal with it. At the very least, the new content should make sense in the context of the content that already exists and not be some copy/paste section content completely unrelated to the flow of the text. Otherwise, we have two sequent sections talking about the same thing.
2) And then the more general recurring problem of these "because-vinnie-told-me-so" issues and PR that are impossible to review, but I'm not going to be repetitive here because I believe we're already working on this problem in parallel.

PR #1051, opened 2025-10-06
don't mark unnamed classes as implementation defined

### LINKED ISSUE #1035: Explore documentation issues with variables and unnamed struct
In 

```
/// Documentation
constexpr struct
{
} single {};
```

the documentation is attached to the unnamed struct rather than the variable.

The unnamed struct is also set as implementation defined and not included in the corpus.

We might need extra logic for the variable to inherit the documentation in these cases.

mizvekov:
These classes can appear as user written code and needs to be documented, as show in the new test case.

In general, for implementation defined declarations, we can detect them as the clang AST exposes them as 'implicit', but this is possible for any declaration, and can be handled more generally.

This is related to https://github.com/cppalliance/mrdocs/issues/1054: We have no way to mark these classes as implementation defined with this change.

This is also related to https://github.com/cppalliance/mrdocs/issues/1055, as the HTML generator has some issues printing unnamed entities.

Fixes #1035

### DIFF
--- a/src/lib/AST/ASTVisitor.cpp
+++ b/src/lib/AST/ASTVisitor.cpp
@@ -644,15 +644,6 @@ populate(
     RecordInfo& I,
     CXXRecordDecl const* D)
 {
-    // Unnamed structs and classes are implementation-defined
-    // or dependencies at best. There are never pages for them.
-    if (D->getIdentifier() == nullptr)
-    {
-        I.Extraction = mostSpecific(
-            ExtractionMode::ImplementationDefined,
-            I.Extraction);
-    }
-
     if (D->getTypedefNameForAnonDecl())
     {
         I.IsTypeDef = true;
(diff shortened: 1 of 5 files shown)


alandefreitas:
From what we discussed so far, I think this is good to go as soon as we move the fixtures from the snippets directory. Please  open new issues to decide what to do about and to fix the other cases: 1) when the user documents the variable rather than the record, 2) when the record is show in the member's table of the parent namespace (it seems like it's simply not there right now), and 3) when the user wants to make the record implementation defined (for now, users have to give it a name to mark them as implementation defined / maybe besides the documentation tag, the thing we just removed could be an option to make unnamed records regular or implementation defined by default).

mizvekov:
@alandefreitas I am not sure what you mean by:
```
2) when the record is show in the member's table of the parent namespace (it seems like it's simply not there right now)
```

I don't see a problem here, can you clarify?

PR #1058, opened 2025-10-08
Fix alias sfinae

### LINKED ISSUE #1057: Out-of-bounds Bit access while extracting SFINAE info
```
#include <memory>
#include <type_traits>

template <typename _Tp>
struct shared_ptr {
    template<typename _Yp, typename _Del, typename _Res = void>
	using _UniqCompatible = std::enable_if_t<
	  std::is_convertible<typename std::unique_ptr<_Yp, _Del>::pointer, _Tp*>::value, _Res>;

    template<typename _Yp, typename _Del>
	using _UniqAssignable = _UniqCompatible<_Yp, _Del, shared_ptr&>;
};
```

This is blocking #1042

K-ballo:
fixes #1057 

When SFINAE detection involves an allias, the implementation confuses the controlling params of the underlying type with those of the alias, which only works when both have equivalent template parameters. This PR makes an attempt to handle the most basic cases by mapping between the two template parameter sets, and avoids an out-of-bounds access in the other cases.

### DIFF
--- a/src/lib/AST/ASTVisitor.cpp
+++ b/src/lib/AST/ASTVisitor.cpp
@@ -2366,19 +2366,7 @@ getSFINAEControlParams(
             Arguments,
             [&](TemplateArgument const& Other)
             {
-                if (Arg.getKind() != Other.getKind())
-                {
-                    return false;
-                }
-                if (Arg.getKind() == TemplateArgument::Type)
-                {
-                    return context_.hasSameType(Other.getAsType(), Arg.getAsType());
-                }
-                if (Arg.getKind() == TemplateArgument::Expression)
-                {
-                    return context_.hasSameExpr(Other.getAsExpr(), Arg.getAsExpr());
-                }
-                return false;
+                return context_.isSameTemplateArgument(Arg, Other);
             });
         bool const found = It != Arguments.end();
         return found ? It - Arguments.begin() : static_cast<std::size_t>(-1);


src/lib/AST/ASTVisitor.cpp
```diff
@@ -2362,22 +2362,26 @@ getSFINAEControlParams(
...
-                return context_.hasSameType(Other.getAsType(), Arg.getAsType());
+                if (Arg.getKind() == TemplateArgument::Type)
+                {
+                    return context_.hasSameType(Other.getAsType(), Arg.getAsType());
+                }
+                if (Arg.getKind() == TemplateArgument::Expression)
+                {
+                    return context_.hasSameExpr(Other.getAsExpr(), Arg.getAsExpr());
+                }
+                return false;
```
mizvekov: This doesn't handle the other template argument kinds. I think you want `isSameTemplateArgument`.

But comparing template arguments like that only makes sense if the parameters are equivalent, as otherwise they'll have different conversions.

K-ballo: This only attempts to handle the boolean case. Given

```
template <bool C, typename T>
using enable_if_t = typename std::enable_if<C, T>::type;
```

the goal here is to match the `C` that is an argument to `std::enable_if` to the `C` that is a parameter of `enable_if_t`.

The pre-existing case attempts to do something similar for `T`. Nothing else (no heroics).

Would `isSameTemplateArgument` be a better choice here? Or is there a simpler alternative?

mizvekov: `isSameTemplateArgument` does what this code does, and more, since it handles the other template argument kinds as well, such as template template arguments, the other kinds of converted constant template arguments, packs, and such.

What I mean is, if we are comparing template arguments to equivalent template parameters, then it makes sense to compare them like this.

If they are otherwise not equivalent, for example you are comparing a template argument for a constant template parameter of type `bool`, to another one of type `int`, then this won't work as expected.

PR #1062, opened 2025-10-08
ci: treat warnings as errors

mizvekov:


.github/workflows/ci.yml
```diff
@@ -79,7 +79,7 @@ jobs:
             llvm-archive-extension: {{#if (ieq os 'windows') }}7z{{else}}tar.bz2{{/if}}
             llvm-archive-filename: {{{ llvm-archive-basename }}}.{{{ llvm-archive-extension }}}
             llvm-sanitizer-config: {{#if (and (ne compiler 'clang') (ne compiler 'apple-clang'))}}{{else if ubsan}}Undefined{{else if asan}}Address{{else if msan}}MemoryWithOrigins{{/if}}
-            common-flags-base: {{#if (ieq compiler 'clang')}}-gz=zstd {{/if}}
+            common-flags-base: {{#if (ieq compiler 'clang')}}-gz=zstd {{/if}}{{#if (eq compiler 'msvc') }}/WX {{else}}-Werror {{/if}}
```
alandefreitas: We can make this PR simpler. I don't know if these flags are being passed to dependencies. But one thing we often do in boost libraries is just to enable warnings as errors directly in CMake when the cmake option for tests is enabled. Then we enable it for the main libraries and for the test executables. In any case, even if using another strategy, we shouldn't be checking for errors in lua. In fact, I think I'm removing this lua bundle in another commit in another PR.

mizvekov: The problem is that the lua sources are compiled as if they were part of mrdocs itself, and it would be annoying to have to provide different flags to different source files within mrdocs.

If you are removing the lua bundle anyway, then we might as well wait for that.

But otherwise, what's happening to it, are we going to be using the system lua implementation?

alandefreitas: The Lua integration has never been completely implemented. In any case, it should be an external dependency, like any other, for several reasons, and that's one of them. I don't know if that removing lua blocks this PR but the other PR I mentioned should be ready today.

alandefreitas: > as well wait for that

#1063

PR #1069, opened 2025-10-14
Fix support for clang cl-mode driver

### LINKED ISSUE #904: `--use-system-libc=false` is not setting default include paths

mizvekov:
Fixes #904

### DIFF
--- a/src/lib/MrDocsCompilationDatabase.cpp
+++ b/src/lib/MrDocsCompilationDatabase.cpp
@@ -10,6 +10,7 @@
 //
 
 #include "MrDocsCompilationDatabase.hpp"
+#include <lib/AST/ClangHelpers.hpp>
 #include <lib/ConfigImpl.hpp>
 #include <lib/Support/Debug.hpp>
 #include <lib/Support/ExecuteAndWaitWithLogging.hpp>
@@ -270,13 +271,14 @@ isValidMrDocsOption(
     return true;
 }
 
-static
-std::vector<std::string>
+static std::vector<std::string>
 adjustCommandLine(
     llvm::StringRef const workingDir,
     std::vector<std::string> const& cmdline,
+    bool is_clang_cl,
     std::shared_ptr<Config const> const& config,
-    std::unordered_map<std::string, std::vector<std::string>> const& implicitIncludeDirectories,
+    std::unordered_map<std::string, std::vector<std::string>> const&
+        implicitIncludeDirectories,
     std::string_view filename)
 {
     if (cmdline.empty())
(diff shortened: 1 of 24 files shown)


src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -300,92 +302,16 @@ adjustCommandLine(
...
     // Add flags to ignore all warnings. Any options that
     // affect warnings will be discarded later.
-    new_cmdline.emplace_back(is_clang_cl ? "/w" : "-w");
+    new_cmdline.emplace_back("-w");
     new_cmdline.emplace_back("-fsyntax-only");
 
-    // ------------------------------------------------------
-    // Target architecture
-    // ------------------------------------------------------
-    constexpr auto is_target_option = [](std::string_view opt) {
```
alandefreitas, line 326: Just out of curiosity, isn't this logic needed anymore (IIRC we had to include this to make it work on macOS), or is this logic reimplemented somewhere else?

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -411,17 +337,21 @@ adjustCommandLine(
...
+               || opt.starts_with("/std:"); // clang-cl options
     };
-    if (std::ranges::find_if(cmdline, is_std_option) == cmdline.end())
+    if (std::ranges::none_of(cmdline, is_std_option))
     {
         if (!isCCompileCommand)
         {
-            new_cmdline.emplace_back("-std=c++23");
+            new_cmdline.emplace_back(
+                is_clang_cl ? "-std:c++23preview" : "-std=c++23");
```
alandefreitas: I may not recall our previous discussions. Do these flags really mean the same?

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -411,17 +337,21 @@ adjustCommandLine(
...
         }
         else
         {
-            new_cmdline.emplace_back("-std=c23");
+            new_cmdline.emplace_back(is_clang_cl ? "-std:c17" : "-std=c23");
```
alandefreitas: Do they mean the same thing? Maybe, if that's possible, this case and the case above would be a good case for using that flag that bypasses the clang-cl conversions? Because I'm not sure someone else would understand exactly why these mean the same thing. In other words, it looks like code someone will want to change in passing.

src/test/TestRunner.cpp
```diff
@@ -108,16 +123,50 @@ handleFile(
     std::shared_ptr<ConfigImpl const> config =
         ConfigImpl::load(fileSettings, dirs_, threadPool_).value();
 
+    auto parentDir = files::getParentDir(filePath);
+    std::unordered_map<std::string, std::vector<std::string>>
+        defaultIncludePaths;
+
+    // Test normally
```
alandefreitas, line 130: Oh... We're testing both modes with all files. Yes. That should be OK.

src/test/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -0,0 +1,347 @@
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (c) 2023 Vinnie Falco (vinnie.falco@gmail.com)
```
alandefreitas: That would be Matheus :)

alandefreitas: Nice file by the way.

test-files/golden-tests/symbols/concept/requires-clause.adoc
```diff
@@ -9,8 +9,8 @@
 [cols=1]
 |===
 | Name
-| link:#A-08[`A`] 
-| link:#A-0b[`A`] 
+| link:#A-09[`A`]
```
alandefreitas, line 12: Oh... The hashes changed with clang. Interesting. It's good to keep track of this.

src/lib/MrDocsCompilationDatabase.cpp
mizvekov: It's not reimplemented. Nothing broke without it as far as I know.

The specific problem for CL-mode here is that, when a target is specified which is different than the MSVC one, then the CL driver ignores a bunch of flags.

src/lib/MrDocsCompilationDatabase.cpp
mizvekov: The CL-mode driver only supports the "std:" spelling instead of "std=".

Just like MSVC, it interchangeably supports either "-" or "/" as a flag introducer.

And MSVC only recognizes the "c++23preview" option, even in the latest MSVC available in compiler explorer.
Clang follows suit here and doesn't recognize just "c++23" either.

src/lib/MrDocsCompilationDatabase.cpp
mizvekov: No they don't mean the same thing, these are different C standards, we should use the same regardless, I will fix this.

mizvekov: Ah I see, CL doesn't have a c23 option because MSVC doesn't either.

I think we could come up with a different approach here, we could switch to using the latest standard, or maybe not doing this at all, and relying on the user trusting the compiler default, or specifying his own.

mizvekov: Changed to use the latest standard for both.

alandefreitas: Oh... Much nicer. 

> relying on the user trusting the compiler default

Yes. I did try that for a while. People tend to have different expectations for documentation. 

Your final solution was better.

PR #1075, opened 2025-10-17
fix(docs): update Open Graph and Twitter meta tags for improved sharing

julioest:
This PR includs large preview cards for Mr.Docs links on X so shared URLs look better.

- Adds minimal `twitter:*` + OG fallbacks
- Sets `twitter:site=@cppalliance`; creator optional/overridable
- Uses absolute HTTPS og:image
- Keeps robots open so Twitterbot can fetch page/image
- X caches ~7 days; refresh by tweeting a fresh URL param

### Design Preview
image

@rbbeeston Lmk if we're okay with this design, I can make more variations - just lmk!


alandefreitas:
Wow!

PR #1077, opened 2025-10-20
use system libs by default

mizvekov:
This change makes mrdocs use the system includes by default instead of the bundled ones.

This helps mrdocs build projects which rely on differences between standard library implementations, since now mrdocs sees the source code closer to what the original compiler did, with only potential differences in the compilers themselves and some minor command line manipulations remaining.

It also stops bundling the clang resource directory, relying on the one installed with libclang.
This makes it so installing both to the same prefix be necessary, which is the same as all the other libclang based tools.

### DIFF
--- a/src/lib/MrDocsCompilationDatabase.cpp
+++ b/src/lib/MrDocsCompilationDatabase.cpp
@@ -303,9 +303,13 @@ adjustCommandLine(
         cmdLineCStrs.data() + cmdLineCStrs.size());
 
     char const* systemIncludeFlag = is_clang_cl ? "-external:I" : "-isystem";
-    // FIXME: No CL equivalent, but not really needed there?
+    // FIXME: No CL equivalent, and not really needed for the microsoft system
+    // headers, but other users could depend on this.
     char const* afterIncludeFlag = is_clang_cl ? "-external:I" : "-idirafter";
-
+    char const* latestCxxStandardFlag = is_clang_cl ? "-std:c++latest" : "-std=c++26";
+    char const* latestCStandardFlag = is_clang_cl ? "-std:clatest" : "-std=c23";
+    char const* noStdlibFlag = is_clang_cl ? "-X" : "-nostdinc++";
+    char const* noSystemLibFlag = is_clang_cl ? "-X" : "-nostdlibinc";
 
     // ------------------------------------------------------
     // Supress all warnings
@@ -351,12 +355,11 @@ adjustCommandLine(
     {
         if (!isCCompileCommand)
         {
-            new_cmdline.emplace_back(
-                is_clang_cl ? "-std:c++latest" : "-std=c++26");
+            new_cmdline.emplace_back(latestCxxStandardFlag);
         }
         else
         {
-            new_cmdline.emplace_back(is_clang_cl ? "-std:clatest" : "-std=c23");
+            new_cmdline.emplace_back(latestCStandardFlag);
         }
     }
 
@@ -399,7 +402,7 @@ adjustCommandLine(
         // implicit include paths and add the standard library
         // and system includes manually. That gives MrDocs
         // access to libc++ in a portable way.
-        new_cmdline.emplace_back(is_clang_cl ? "-X" : "-nostdinc++");
+        new_cmdline.emplace_back(noStdlibFlag);
         for (auto const& inc : (*config)->stdlibIncludes)
         {
             new_cmdline.emplace_back(systemIncludeFlag);
@@ -409,7 +412,7 @@ adjustCommandLine(
 
     if (!(*config)->useSystemLibc)
     {
-        new_cmdline.emplace_back(is_clang_cl ? "-X" : "-nostdlibinc");
+        new_cmdline.emplace_back(noSystemLibFlag);
         for (auto const& inc : (*config)->libcIncludes)
         {
             new_cmdline.emplace_back(afterIncludeFlag);


.github/workflows/ci.yml
```diff
@@ -91,7 +91,7 @@ jobs:
             common-flags-base: {{#if (ieq compiler 'clang')}}-gz=zstd {{/if}}
             common-flags: {{{ common-flags-base }}}{{#if msan }}-fsanitize-memory-track-origins {{/if}}
             common-ccflags: {{{ ccflags }}} {{{ common-flags }}}
-            mrdocs-flags: {{{ warning-flags }}}{{#if (and (eq compiler 'gcc') (not asan)) }}-static {{/if}}
+            mrdocs-flags: {{{ warning-flags }}}
```
alandefreitas: Why are we doing that? A huge problem we have had in the past is that the Linux release, which is actually an Ubuntu release, would not work on other versions of Ubuntu unless we set that. Otherwise, it would depend on a specific version of libc, and the release would fail. Is this going to break now?

.github/workflows/ci.yml
```diff
@@ -726,6 +726,26 @@ jobs:
       contents: write
 
     steps:
+      # This calculates a bunch of variables, which would normally go in to the regular matrix extra-values
+      # section, but which depend on paths not known at that point.
+      - name: Resolved Matrix
```
alandefreitas, line 731: Is this a new matrix of this kind or just code being moved? Because I remember we already had a step like that.

CMakeLists.txt
```diff
@@ -384,6 +375,7 @@ if (WIN32)
             mrdocs-core
             PUBLIC
             /permissive-    # strict C++
+            /Zc:__cplusplus # report C++ standard support
```
alandefreitas: What does that do?

.clang-format
```diff
@@ -227,3 +227,7 @@ IncludeCategories:
 # Comments
 FixNamespaceComments: true
 CommentPragmas: '^ clang-format'
+
+---
+Language: Json
```
alandefreitas, line 232: These kinds of changes can go in `style: *` commits.

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -278,7 +278,7 @@ It's also common for libraries to depend on the C++ standard library, the C stan
 
 That means unless `-nostdinc` is defined, all systems include paths are included. This is what allows the user to also use headers like `<Windows.h>` or `<linux/version.h>` without explicitly including anything else, even though they are not part of the C standard library. This is often seen as a convenience but can lead to portability issues.
 
-In this context, MrDocs provides the `use-system-stdlib` and `use-system-libc` options. Both are set as `false` by default, meaning MrDocs will compile the code as if the `-nostdinc&plus;&plus; -nostdlib&plus;&plus;` and `-nostdinc` flags were passed to Clang. Additionally:
+In this context, MrDocs provides the `use-system-stdlib` and `use-system-libc` options. Both are set as `true` by default; setting both to `false` results in MrDocs compiling the code as if the `-nostdinc&plus;&plus; -nostdlib&plus;&plus;` and `-nostdinc` flags were passed to Clang. Additionally:
```
alandefreitas, line 281: Nice catch. It's quite hard to remember to fix those details in the documentation. :)

docs/mrdocs.schema.json
```diff
@@ -621,7 +620,7 @@
       "type": "boolean"
     },
     "use-system-stdlib": {
-      "default": false,
+      "default": true,
```
alandefreitas, line 623: Out of curiosity, that's the default behavior for Windows and Linux, while the opposite would be the default behavior on MacOS, right?

docs/mrdocs.schema.json
```diff
@@ -611,7 +610,7 @@
       "type": "string"
     },
     "use-system-libc": {
```
alandefreitas, line 612: Should this option be called something else? Because it seems like the other option does relate to the C++ standard library, but it seems like this option is just about making the typical system paths available by default and not the C standard library in particular.

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -302,7 +302,10 @@ adjustCommandLine(
         cmdLineCStrs.data(),
         cmdLineCStrs.data() + cmdLineCStrs.size());
 
-    auto const systemIncludeFlag = is_clang_cl ? "-external:I" : "-isystem";
+    char const* systemIncludeFlag = is_clang_cl ? "-external:I" : "-isystem";
+    // FIXME: No CL equivalent, but not really needed there?
```
alandefreitas: I don't really understand this comment. It's quite weird to have new code that already has a fix me comment, especially when it's a single line of code. 😅

Maybe it's more about a limitation than something to be fixed. I'm not sure because I'm not completely sure I understand the problem that needs to be fixed here. 

If there is no CL equivalent, but it works for CL, maybe it's not needed, right? Or can't we just ensure it's included at the end by ourselves? 🙂

src/lib/MrDocsCompilationDatabase.cpp
```diff
@@ -406,10 +409,10 @@ adjustCommandLine(
 
     if (!(*config)->useSystemLibc)
     {
-        new_cmdline.emplace_back("-nostdinc");
+        new_cmdline.emplace_back(is_clang_cl ? "-X" : "-nostdlibinc");
```
alandefreitas: Should we put all this logic about what the arguments look like in a single place? For some flags, we're doing it at the beginning and already calculating what the arguments look like. For some other flags, we're doing it in-line.

test-files/golden-tests/snippets/sqrt.yml
```diff
@@ -0,0 +1,2 @@
+warn-if-doc-error: false
```
alandefreitas: What kind of errors are we getting here? Because we should indeed ensure that the snippets have nothing missing. That's very important because that's going to the website.

.github/workflows/ci.yml
mizvekov: Yes this step must be done per workflow that depends on resolved variables. We had two such workflows, now we add one more.

src/lib/MrDocsCompilationDatabase.cpp
mizvekov: The problem here is that MSVC doesn't have an equivalent of the driver flag which allows us to insert the system includes in the right order, but that does not seem to be really needed on MSVC because the windows system headers don't depend on the clang headers.

.github/workflows/ci.yml
alandefreitas: I think this is logically correct, but what we're doing here is some kind of overgeneralization of the problem. I think this step would be easier to understand if it were called just "Resolve paths" because the idea that this somehow has some necessary relationship with the matrix is false. 

- Of course, they depend on the matrix as input, but so do all other steps. 
- And in terms of the relationship with the matrix in the output, also, any other step could potentially want their results included on top of the matrix as a form of convenience. If one agrees this is convenient, this is, at best, more of a convenience than a necessity because it's not a necessity at all.

In other words, we're generalizing it in a direction that has no cognitive significance because any relationship this step has with the matrix is not more special than the relationship any other step has with the matrix. 

I think for someone looking at this for the first time, if this step were called "Resolve paths" and the output only included the paths, then it would be clear to anyone what this step does, and it would also be clear how we can work on that to come up with some new action in the repository of actions that simplifies that if that's something we're doing recurrently. 

Or yet, in other words, we're making this much more controversial than it has to be. 😅

src/lib/MrDocsCompilationDatabase.cpp
alandefreitas: I see. Well, I guess we still have to decide what to do about it, then? Because if that's the case, there's nothing to be fixed, and this comment doesn't make sense because this is just a regular implementation decision that has to be explained.

.github/workflows/ci.yml
mizvekov: Though that's the reason we need this step, the "Resolved Matrix" step produces variables that are not pure paths, like compiler flags for example. Just not the Resolved Matrix from this workflow, but the other workflows do have those.

These are basically things we would logically have put in the matrix, except they somehow depend on paths that are not known at that point.

So for example, there are things in the matrix right now which are not in the resolved matrix by pure coincidence, just because they don't depend on those paths right now.

So I think calling this the resolved matrix helps to reinforce that variables can arbitrarily be moved between these two steps.

docs/mrdocs.schema.json
mizvekov: I don't think so, why would we want a different default on MacOS?

MacOS makes a better example for why we would want to use the system libs, since they use a libc++ fork and we don't ship that.

docs/mrdocs.schema.json
mizvekov: I agree, but we could change this in a follow up though.

src/lib/MrDocsCompilationDatabase.cpp
mizvekov: Well, the MSVC system includes don't depend on the clang-headers and the 'include_next' extension, but other
system includes, even for example some custom user system includes, could theoretically depend on this.

I know there is a toolchain out there which uses clang-cl + mingw headers, that could also depend on this.

So the use case is a bit theoretical, but exists. Probably doesn't justify us doing anything at this point, but leaving a FIXME seems fair.

alandefreitas: But I think the main point is you're the only person who's going to understand this FIXME. None of what you just said here is in the comment, and I'm sure a random person coming across this comment won't understand it without thiscontext either.

.github/workflows/ci.yml
alandefreitas: LGTM

docs/mrdocs.schema.json
alandefreitas: > I agree, but we could change this in a follow up though.

Yes. It's just something to keep track of (an issue to explore it).

docs/mrdocs.schema.json
alandefreitas: Okie dokie

alandefreitas (approved):
LGTM 👍

PR #1101, opened 2025-10-30
docs: remove gradients, use solid background, and clean up styles

julioest:
- Set a consistent blue background color to home page and antora documentation
- Remove footer's li broken css rule

**Preview**
Screenshot 2025-10-30 at 5 14 06 PM

### DIFF
--- a/docs/ui/src/css/base.css
+++ b/docs/ui/src/css/base.css
@@ -9,15 +9,7 @@ html {
   font-size: var(--body-font-size);
   height: 100%;
   scroll-behavior: smooth;
-  background:
-    radial-gradient(ellipse 100% 100% at var(--x1) var(--y1), rgba(89, 36, 99, 0.8), transparent 50%),
-    radial-gradient(ellipse 80% 120% at var(--x2) var(--y2), rgba(138, 214, 236, 0.5), transparent 55%),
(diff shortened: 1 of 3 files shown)


alandefreitas:
Looks really nice 😊 It also looks great visually 💙

Since we don’t have the gradient anymore, maybe it’d be a good idea to extend the blue background from the main content area so it matches the same blue used behind the breadcrumbs (which are currently a bit lighter)? 🤔

There’s just one small thing about the “style” tag in commits I used to get confused by that, because in web projects it can seem like “style” is meant for CSS changes, while other commit types are for the backend. But that threw me off since lots of non-web projects use that tag as well. And even CSS-heavy projects don’t use “style” for everything, which made it even more confusing 😅

After digging into it a bit more, I found that the “style” tag is usually for non-functional code changes — things like formatting/linting. The idea is to keep those commits separate so they don’t end up in the changelog. CSS-related changes, on the other hand, use whatever tag matches their scope in the project. In our case, that’d be “docs” ❤️

(That's just a comment for future PRs. Don't worry too much about it)

rbbeeston (approved):

julioest:
I appreciate the clarity regarding the commit + tags best practices 🙏

PR #1120, opened 2025-12-01
fix: don't produce warnings for symbols copied from base classes

### LINKED ISSUE #1119: mrdocs produces bogus diagnostic when inheriting from an external class
Reduced test case from recent #1077 [failure](https://github.com/cppalliance/mrdocs/actions/runs/19715290161/job/56485986054?pr=1077): 

```C++
namespace ns {
  struct Foo {
    /// bar
    int bar();
  };
}
/// Baz
struct Baz : ns::Foo {};
```

Assumes `mrdocs.yml` includes some other namespace, but doesn't include `ns`:
```
include-symbols:
  - 'mrdocs::**'
```

For this test case, mrdocs produces the following diagnostic:
```
test.hpp:4:5:
    1) Baz::bar: Missing documentation for return value
  4 |     int bar();
    |     ^~~~~~~~~~
```

The documentation for `bar` produces a diagnostic even though it is external to the project (member of an excluded namespace).

mizvekov:
Fixes #1119

### DIFF
--- a/include/mrdocs/Metadata/Symbol/SymbolBase.hpp
+++ b/include/mrdocs/Metadata/Symbol/SymbolBase.hpp
@@ -71,6 +71,11 @@ struct MRDOCS_VISIBLE Symbol {
     */
     ExtractionMode Extraction = ExtractionMode::Dependency;
 
+    /** Whether this a copy of an inherited method, as produced when
+        `inherit-base-members` is not `never`.
+    */
+    bool IsCopyFromInherited = false;
+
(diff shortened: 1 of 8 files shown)


alandefreitas (commented):
Good work exploring this issue. 😃

The warning issue and issue where symbols in the global namespace are extracted both very merit. 👍

Let's be careful about the contract of `inherit-base-members` and the golden-tests in config that broke. 😅

alandefreitas (approved):
LGTM!

PR #1130, opened 2025-12-09
refactor(metadata): integrate Boost.Describe

gennaroprota:
This introduces Boost.Describe and Boost.Mp11 and applies them across metadata types (symbols, enums, inline elements) to replace hand-written sequences of `io.map()` calls with reflection-based code. This improves maintainability without altering the public API or user-facing features.

@alandefreitas: Please have a look at `normalizeMemberName()` in SymbolDescribeMapper.hpp: Some names are special-cased there. If you prefer, I can remove the special cases and modify the Handlebars templates to use the canonical names.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -615,7 +615,7 @@ jobs:
           package-dir: packages
           package-generators: ${{ matrix.mrdocs-package-generators }}
           package-artifact: false
-          ctest-timeout: 3600
+          ctest-timeout: 9000
 
       - name: Check YAML schema
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -433,6 +433,44 @@ jobs:
         run: |
           rm -r ../third-party/llvm-project
 
+      - name: Install Boost.Mp11
+        uses: alandefreitas/cpp-actions/cmake-workflow@v1.8.12
+        with:
+          source-dir: ../third-party/boost_mp11
+          git-repository: https://github.com/boostorg/mp11
+          git-tag: boost-1.90.0
+          patches: |
+            ./third-party/patches/boost_mp11/CMakeLists.txt
+            ./third-party/patches/boost_mp11/boost_mp11-config.cmake.in
+          build-dir: ${sourceDir}/build
+          cc: ${{ steps.setup-cpp.outputs.cc }}
+          cxx: ${{ steps.setup-cpp.outputs.cxx }}
+          build-type: Release
+          shared: false
+          install: true
+          install-prefix: ${sourceDir}/install
+          run-tests: false
+          trace-commands: true
+
+      - name: Install Boost.Describe
+        uses: alandefreitas/cpp-actions/cmake-workflow@v1.8.12
+        with:
+          source-dir: ../third-party/boost_describe
+          git-repository: https://github.com/boostorg/describe
+          git-tag: boost-1.90.0
+          patches: |
+            ./third-party/patches/boost_describe/CMakeLists.txt
+            ./third-party/patches/boost_describe/boost_describe-config.cmake.in
+          build-dir: ${sourceDir}/build
+          cc: ${{ steps.setup-cpp.outputs.cc }}
+          cxx: ${{ steps.setup-cpp.outputs.cxx }}
+          build-type: Release
+          shared: false
+          install: true
+          install-prefix: ${sourceDir}/install
+          run-tests: false
+          trace-commands: true
+
       - name: Install JerryScript
         uses: alandefreitas/cpp-actions/cmake-workflow@v1.9.2
         with:
@@ -567,6 +605,8 @@ jobs:
             -D LUA_ROOT="${{ steps.rmatrix.outputs.third-party-dir }}/lua/install"
             -D Lua_ROOT="${{ steps.rmatrix.outputs.third-party-dir }}/lua/install"
             -D lua_ROOT="${{ steps.rmatrix.outputs.third-party-dir }}/lua/install"
+            -D boost_mp11_ROOT="${{ steps.rmatrix.outputs.third-party-dir }}/boost_mp11/install"
+            -D boost_describe_ROOT="${{ steps.rmatrix.outputs.third-party-dir }}/boost_describe/install"
             ${{ runner.os == 'Windows' && '-D LibXml2_ROOT=../third-party/libxml2/install' || '' }}
           export-compile-commands: true
           run-tests: true
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,6 +223,10 @@ endif()
 
 set(CMAKE_FOLDER Dependencies)
 
+# Boost.Describe + Boost.Mp11 (header-only, fetched via bootstrap.py recipes)
+find_package(boost_mp11 REQUIRED)
+find_package(boost_describe REQUIRED)
+
 # LLVM + Clang
 if (LLVM_ROOT)
     # LLVM_ROOT is absolute
@@ -320,6 +324,7 @@ target_compile_definitions(
     PRIVATE
     -DMRDOCS_TOOL
 )
+target_link_libraries(mrdocs-core PRIVATE Boost::describe Boost::mp11)
 
 # Dependencies
 target_include_directories(mrdocs-core
--- a/docs/mrdocs.yml
+++ b/docs/mrdocs.yml
@@ -13,6 +13,14 @@ implementation-defined:
   - 'mrdocs::detail'
   - 'mrdocs::report::detail'
   - 'mrdocs::dom::detail'
+exclude-symbols:
+    # Symbols defined when using the Boost.Describe macros.
+  - 'mrdocs::**::boost_base_descriptor_fn'
+  - 'mrdocs::**::boost_public_member_descriptor_fn'
+  - 'mrdocs::**::boost_protected_member_descriptor_fn'
+  - 'mrdocs::**::boost_private_member_descriptor_fn'
+  - 'mrdocs::**::boost_enum_descriptor_fn'
+  - 'mrdocs::**::should_use_BOOST_DESCRIBE_NESTED_ENUM'
 multipage: true
 generator: adoc
 cmake: '-D MRDOCS_DOCUMENTATION_BUILD=ON'
--- a/include/mrdocs/Metadata/DocComment/Inline/ImageInline.hpp
+++ b/include/mrdocs/Metadata/DocComment/Inline/ImageInline.hpp
@@ -65,13 +65,7 @@ tag_invoke(
     dom::LazyObjectMapTag t,
     IO& io,
     ImageInline const& I,
-    DomCorpus const* domCorpus)
-{
-    tag_invoke(t, io, dynamic_cast<Inline const&>(I), domCorpus);
-    tag_invoke(t, io, dynamic_cast<InlineContainer const&>(I), domCorpus);
-    io.map("src", I.src);
-    io.map("alt", I.alt);
-}
+    DomCorpus const* domCorpus);
 
 /** Return the @ref ImageInline as a @ref dom::Value object.
 */
--- a/include/mrdocs/Metadata/Expression.hpp
+++ b/include/mrdocs/Metadata/Expression.hpp
@@ -4,6 +4,7 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
 // Copyright (c) 2023 Krystian Stasiowski (sdkrystian@gmail.com)
+// Copyright (c) 2025 Gennaro Prota (gennaro.prota@gmail.com)
 //
 // Official repository: https://github.com/cppalliance/mrdocs
 //
@@ -13,11 +14,15 @@
 
 #include <mrdocs/Platform.hpp>
 #include <mrdocs/ADT/Optional.hpp>
+#include <mrdocs/Dom/LazyObject.hpp>
+#include <mrdocs/Dom/Value.hpp>
 #include <concepts>
 #include <optional>
 #include <string>
 
 namespace mrdocs {
+
+class DomCorpus;
 
 /** Represents an expression
 */
@@ -88,6 +93,39 @@ static void merge(
     {
         I.Value = std::move(Other.Value);
     }
+}
+
+/** Map an ExprInfo to a @ref dom::Value object.
+
+    @param v The output parameter to receive the dom::Value.
+    @param expr The expression info to convert.
+*/
+inline
+void
+tag_invoke(
+    dom::ValueFromTag,
+    dom::Value& v,
+    ExprInfo const& expr,
+    DomCorpus const*)
+{
+    v = expr.Written;
+}
+
+/** Map an ExprInfo to a @ref dom::Value object.
+
+    @param v The output parameter to receive the dom::Value.
+    @param expr The expression info to convert.
+*/
+inline
+void
+tag_invoke(
+    dom::LazyObjectMapTag,
+    dom::Value& v,
+    ExprInfo const& expr,
+    DomCorpus const*
+)
+{
+    v = expr.Written;
 }
 
 } // mrdocs
--- a/include/mrdocs/Metadata/Symbol/Namespace.hpp
+++ b/include/mrdocs/Metadata/Symbol/Namespace.hpp
@@ -105,19 +105,7 @@ tag_invoke(
     dom::LazyObjectMapTag,
     IO& io,
     NamespaceTranche const& I,
-    DomCorpus const* domCorpus)
-{
-    io.map("namespaces", dom::LazyArray(I.Namespaces, domCorpus));
-    io.map("namespaceAliases", dom::LazyArray(I.NamespaceAliases, domCorpus));
-    io.map("typedefs", dom::LazyArray(I.Typedefs, domCorpus));
-    io.map("records", dom::LazyArray(I.Records, domCorpus));
-    io.map("enums", dom::LazyArray(I.Enums, domCorpus));
-    io.map("functions", dom::LazyArray(I.Functions, domCorpus));
-    io.map("variables", dom::LazyArray(I.Variables, domCorpus));
-    io.map("concepts", dom::LazyArray(I.Concepts, domCorpus));
-    io.map("guides", dom::LazyArray(I.Guides, domCorpus));
-    io.map("usings", dom::LazyArray(I.Usings, domCorpus));
-}
+    DomCorpus const* domCorpus);
 
 /** Map the NamespaceTranche to a @ref dom::Value object.
 */
@@ -195,14 +183,7 @@ tag_invoke(
     dom::LazyObjectMapTag t,
     IO& io,
     NamespaceSymbol const& I,
-    DomCorpus const* domCorpus)
-{
-    tag_invoke(t, io, I.asInfo(), domCorpus);
-    io.map("isInline", I.IsInline);
-    io.map("isAnonymous", I.IsAnonymous);
-    io.map("members", I.Members);
-    io.map("usingDirectives", dom::LazyArray(I.UsingDirectives, domCorpus));
-}
+    DomCorpus const* domCorpus);
 
 /** Map the NamespaceSymbol to a @ref dom::Value object.
 */
--- a/include/mrdocs/Metadata/Symbol/NamespaceAlias.hpp
+++ b/include/mrdocs/Metadata/Symbol/NamespaceAlias.hpp
@@ -59,11 +59,7 @@ tag_invoke(
     dom::LazyObjectMapTag t,
     IO& io,
     NamespaceAliasSymbol const& I,
-    DomCorpus const* domCorpus)
-{
-    tag_invoke(t, io, I.asInfo(), domCorpus);
-    io.map("aliasedSymbol", I.AliasedSymbol);
-}
+    DomCorpus const* domCorpus);
 
 /** Map the NamespaceAliasSymbol to a @ref dom::Value object.
 */
--- a/src/lib/Metadata/Source.cpp
+++ b/src/lib/Metadata/Source.cpp
@@ -5,10 +5,13 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
 // Copyright (c) 2023 Vinnie Falco (vinnie.falco@gmail.com)
+// Copyright (c) 2025 Gennaro Prota (gennaro.prota@gmail.com)
 //
 // Official repository: https://github.com/cppalliance/mrdocs
 //
 
+#include <lib/Support/Reflection/Reflection.hpp>
+#include <lib/Support/Reflection/MapReflectedType.hpp>
 #include <mrdocs/Dom/LazyArray.hpp>
 #include <mrdocs/Dom/LazyObject.hpp>
 #include <mrdocs/Metadata/Symbol/FileKind.hpp>
@@ -18,22 +21,6 @@
 #include <ranges>
 
 namespace mrdocs {
-
-std::string_view
-toString(FileKind kind)
-{
-    switch(kind)
-    {
-    case FileKind::Source:
-        return "source";
-    case FileKind::System:
-        return "system";
-    case FileKind::Other:
-        return "other";
-    default:
-        MRDOCS_UNREACHABLE();
-    };
-}
 
 namespace
 {
@@ -107,12 +94,7 @@ tag_invoke(
     IO& io,
     Location const& loc)
 {
-    io.map("fullPath", loc.FullPath);
-    io.map("shortPath", loc.ShortPath);
-    io.map("sourcePath", loc.SourcePath);
-    io.map("line", loc.LineNumber);
-    io.map("column", loc.ColumnNumber);
-    io.map("documented", loc.Documented);
+    mapReflectedType(io, loc);
 }
 
 void
--- a/src/lib/Metadata/Symbol/Function.cpp
+++ b/src/lib/Metadata/Symbol/Function.cpp
@@ -271,24 +271,6 @@ getOperatorReadableName(
     MRDOCS_UNREACHABLE();
 }
 
-dom::String
-toString(FunctionClass const kind) noexcept
-{
-    switch(kind)
-    {
-    case FunctionClass::Normal:
-        return "normal";
-    case FunctionClass::Constructor:
-        return "constructor";
-    case FunctionClass::Conversion:
-        return "conversion";
-    case FunctionClass::Destructor:
-        return "destructor";
-    default:
-        MRDOCS_UNREACHABLE();
-    }
-}
-
 void
 merge(Param& I, Param&& Other)
 {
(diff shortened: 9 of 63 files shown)


include/mrdocs/Metadata/Symbol/EnumToString.hpp
```diff
@@ -0,0 +1,52 @@
...
+std::string
+toString(Enum e) noexcept
+{
+    std::string result;
+    boost::mp11::mp_for_each<
+        boost::describe::describe_enumerators<Enum>>(
+        [&](auto const& D)
+        {
+            using Descriptor = std::decay_t<decltype(D)>;
+            if (Descriptor::value == e)
```
pdimov: You should be able to use `D.value` directly here (and `D.name` below). This eliminates the need for `Descriptor`.

include/mrdocs/Metadata/DocComment/Inline/ImageInline.hpp
```diff
@@ -5,16 +5,19 @@
...
 // Copyright (c) 2025 Alan de Freitas (alandefreitas@gmail.com)
+// Copyright (c) 2025 Gennaro Prota (gennaro.prota@gmail.com)
 //
 // Official repository: https://github.com/cppalliance/mrdocs
 //
 
 #ifndef MRDOCS_API_METADATA_DOCCOMMENT_INLINE_IMAGEINLINE_HPP
 #define MRDOCS_API_METADATA_DOCCOMMENT_INLINE_IMAGEINLINE_HPP
 
+#include <boost/describe/class.hpp>
```
vinniefalco: includes should be ordered from most specific to most general. describe is general. std includes are most general

include/mrdocs/Metadata/Symbol/Namespace.hpp
```diff
@@ -91,6 +106,24 @@ allMembers(NamespaceTranche const& T)
             }
         }
     ) | std::ranges::views::join;
+
+#if 0
```
vinniefalco: this whole thing is commented out?

.github/workflows/ci.yml
```diff
@@ -432,6 +432,36 @@ jobs:
         run: |
           rm -r ../third-party/llvm-project
 
+      - name: Install Boost.Describe
+        uses: alandefreitas/cpp-actions/cmake-workflow@v1.8.12
+        with:
+          source-dir: ../third-party/boost_describe
+          git-repository: https://github.com/boostorg/describe
```
alandefreitas: Setting the url is cheaper in this case.

CMakeLists.txt
```diff
@@ -277,6 +277,20 @@ llvm_map_components_to_libnames(llvm_libs all)
 string(REGEX REPLACE " /W[0-4]" "" CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
 string(REGEX REPLACE " /W[0-4]" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 
+# Boost.Describe + Boost.Mp11 (header-only, fetched via bootstrap.py recipes)
+set(BOOST_DESCRIBE_ROOT "${CMAKE_SOURCE_DIR}/build/third-party/source/boost_describe")
```
alandefreitas: Hard-coded paths are a no-no. 😅 You can write a patch for each that installs and exports them as Boost::describe and Boost::mp11. This way, you still don't depend on the Boost root project. Also, I thought there would be dependency on Boost::config here.

include/mrdocs/Metadata/Expression.hpp
```diff
@@ -90,6 +96,39 @@ static void merge(
...
+inline
+void
+tag_invoke(
+    dom::LazyObjectMapTag,
+    dom::Value& v,
+    ExprInfo const& expr,
+    DomCorpus const*
+)
+{
+    v = expr.Written;
```
alandefreitas, line 128: It's great that you haven't broken anything in the old behavior. This allows working within boundaries at each step. 😀

In later issues and PRs, we want these descriptions to match the structure exactly so we have the same data structure for all generators.

include/mrdocs/Metadata/Symbol/EnumToString.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <mrdocs/Support/String.hpp>
+#include <string>
+#include <type_traits>
+
+namespace mrdocs {
+
+/** Convert a Boost.Describe'd enumerator to string form.
+*/
+template <typename Enum>
+    requires std::is_enum_v<Enum>
```
alandefreitas: Shouldn't we check if it's boost.described?

include/mrdocs/Metadata/Symbol/EnumToString.hpp
alandefreitas: Oh! Peter's here! Cool! 😎

include/mrdocs/Metadata/Symbol/Namespace.hpp
```diff
@@ -91,6 +106,24 @@ allMembers(NamespaceTranche const& T)
...
+    constexpr std::size_t N = boost::mp11::mp_size<Members>::value;
+
+    // Build an array of ref_views over each member vector
+    std::array arr = [&]<std::size_t... I>(std::index_sequence<I...>) {
+        return std::array{
+            std::ranges::ref_view{T.*boost::mp11::mp_at_c<Members, I>::pointer}...
+        };
+    }(std::make_index_sequence<N>{});
+
+    // Now flatten lazily: iota -> transform -> join
```
alandefreitas: Trying to come up with a cheaper version of [`std::ranges::views::concat`](https://en.cppreference.com/w/cpp/ranges/concat_view.html)? We've all been there. :)

include/mrdocs/Metadata/Symbol/Namespace.hpp
```diff
@@ -107,6 +140,8 @@ tag_invoke(
     NamespaceTranche const& I,
     DomCorpus const* domCorpus)
 {
+    mapWithDescribe(io, I, domCorpus);
+/*
```
alandefreitas: We can just remove these comments

include/mrdocs/Metadata/Symbol/NamespaceAlias.hpp
```diff
@@ -62,7 +71,7 @@ tag_invoke(
     DomCorpus const* domCorpus)
 {
     tag_invoke(t, io, I.asInfo(), domCorpus);
-    io.map("aliasedSymbol", I.AliasedSymbol);
+    mapWithDescribe(io, I, domCorpus);
```
alandefreitas: Can't mapWithDescribe just iterate the base classes itself?

src/lib/Support/Reflection/MapReflectedType.hpp
```diff
@@ -0,0 +1,176 @@
...
+#include <locale>
+#include <string_view>
+
+namespace mrdocs {
+
+class DomCorpus;
+
+namespace detail {
+
+/** Helper to determine if a member should be mapped based on its value.
```
alandefreitas, line 43: Can't we have all the Boost.Describe related stuff in a single header? And will these headers always be public? But that would make mp11 and describe the first public dependencies of mrdocs, and cmakelists.txt and mrdocs.yml would need to be adapted accordingly.

src/lib/Metadata/Source.cpp
```diff
@@ -107,12 +93,14 @@ tag_invoke(
     IO& io,
     Location const& loc)
 {
+    mapWithDescribe(io, loc, nullptr);
+    /*GPS
```
alandefreitas: These comments can go. :)

alandefreitas (commented):
Great job! 👏😀🎉

alandefreitas:
> @alandefreitas: Please have a look at normalizeMemberName() in SymbolDescribeMapper.hpp: Some names are special-cased there. If you prefer, I can remove the special cases and modify the Handlebars templates to use the canonical names.

Sorry. I forgot to answer this one. 😬

Yes. We're going to use canonical names for everything sooner or later. Since you already made the effort to split this into two tasks, we can do that later rather than sooner. Reflection will enable all kinds of simplifications in the code from now on. 

I leave it up to you, though.

include/mrdocs/Metadata/Symbol/Namespace.hpp
gennaroprota: Yes, sorry. I meant to re-examine it later, but then I forgot.

include/mrdocs/Metadata/Symbol/EnumToString.hpp
gennaroprota: Yes, that's better.

include/mrdocs/Metadata/Symbol/NamespaceAlias.hpp
gennaroprota: Should it always iterate all bases? I was afraid there was some special case somewhere for which only one base was mapped. If you confirm I can iterate all bases, always, I'll do that :-).

After writing the above comment, I found this:

```
struct ImageInline final
    : InlineCommonBase<InlineKind::Image>
    , InlineContainer
{
    /** Image source URL or path.
    */
    std::string src;
    /** Alternate text when the image cannot be shown.
    */
    std::string alt;

    /** Order images by source, alt text, and inline children.
    */
    auto operator<=>(ImageInline const&) const = default;
    /** Equality compares source, alt text, and contents.
    */
    bool operator==(ImageInline const&) const noexcept = default;
};

template <class IO>
void
tag_invoke(
    dom::LazyObjectMapTag t,
    IO& io,
    ImageInline const& I,
    DomCorpus const* domCorpus)
{
    tag_invoke(t, io, dynamic_cast<Inline const&>(I), domCorpus);
    tag_invoke(t, io, dynamic_cast<InlineContainer const&>(I), domCorpus);
    mapWithDescribe(io, I, domCorpus);
}
```

src/lib/Metadata/Source.cpp
gennaroprota: Oops, thanks.

src/lib/Support/Reflection/MapReflectedType.hpp
gennaroprota: What do you mean by "*all* the Boost.Describe related stuff"?

.github/workflows/ci.yml
gennaroprota: What do you mean?

alandefreitas:
Sometimes we get conflicts if we don't rename things for Handlebars, as you notice with "class" and "functionClass", etc. This is because of how handlebars works with recursion, and it's very annoying. I don't know if it solves your problem, but one solution I was going to propose at some point is adding an "@meta" key to all objects to describe the class name, etc (this can also help with other problems we've been having). This way, we would have the proper information to make the distinction in handlebars.

include/mrdocs/Metadata/Symbol/NamespaceAlias.hpp
vinniefalco: Image?

src/lib/Support/Reflection/MapReflectedType.hpp
vinniefalco: Authors of plugins don't need the Describe data (or do they) so it would be nice not to require plugins to have Boost installed.

alandefreitas: > What do you mean by "all the Boost.Describe related stuff"?

That there should be no #include <boost/whatever.hpp> in public files.

> or do they

The members are already "described" for extensions.  Authors of plugins are probably going to iterate the members of objects often, but this can be abstracted without exposing boost.

.github/workflows/ci.yml
```diff
@@ -575,7 +615,7 @@ jobs:
           package-dir: packages
           package-generators: ${{ matrix.mrdocs-package-generators }}
           package-artifact: false
-          ctest-timeout: 3600
+          ctest-timeout: 9000
```
alandefreitas, line 618: From 1 to 2.5 hours? What's happening here? In fact, 3600 was already a generous temporary workaround until we get some problems with tests fixed.

.github/workflows/ci.yml
```diff
@@ -433,6 +433,44 @@ jobs:
...
+          trace-commands: true
+
+      - name: Install Boost.Describe
+        uses: alandefreitas/cpp-actions/cmake-workflow@v1.8.12
+        with:
+          source-dir: ../third-party/boost_describe
+          git-repository: https://github.com/boostorg/describe
+          git-tag: boost-1.90.0
+          patches: |
+            ./third-party/patches/boost_describe/CMakeLists.txt
```
alandefreitas, line 462: Nice

docs/mrdocs.yml
```diff
@@ -13,6 +13,14 @@ implementation-defined:
   - 'mrdocs::detail'
   - 'mrdocs::report::detail'
   - 'mrdocs::dom::detail'
+exclude-symbols:
+    # Symbols defined when using the Boost.Describe macros.
+  - '**::boost_base_descriptor_fn'
```
alandefreitas: What namespace are they defined in?

include/mrdocs/Dom/LazyArray.hpp
```diff
@@ -211,6 +213,26 @@ TransformArray(T const& arr, F const& f)
...
+
+/** Map a vector of strings to a @ref dom::Value object.
+
+    @param v The output parameter to receive the dom::Value.
+    @param elems The vector of T's to convert.
+    @param domCorpus The DomCorpus used to resolve references.
+ */
+template <typename T>
+void
+tag_invoke(
```
alandefreitas: This is breaking encapsulation between vectors, DOM lazy arrays, and DOM values. It'd be better if the function that iterates members of mrdocs objects used a lazy array directly whenever the member is a vector.

include/mrdocs/Support/EnumToString.hpp
```diff
@@ -0,0 +1,41 @@
...
+#define MRDOCS_API_SUPPORT_ENUMTOSTRING_HPP
+
+#include <string>
+#include <type_traits>
+
+namespace mrdocs {
+
+/** Convert a Boost.Describe'd enumerator to string form.
+
+    @param e The enumerator to convert.
```
alandefreitas: What does it do when it's not described? How can the user of the public API know if it's described? Should this be public?

include/mrdocs/Support/EnumToString.hpp
```diff
@@ -0,0 +1,41 @@
...
+    @return The string form of the enumerator.
+*/
+template <typename Enum>
+    requires std::is_enum_v<Enum>
+std::string
+toString(Enum e);
+
+enum class ExtractionMode;
```
alandefreitas: Why the forward declarations?

src/lib/Support/Reflection/MapReflectedType.hpp
```diff
@@ -0,0 +1,201 @@
...
+*/
+template <typename T>
+constexpr bool
+shouldMapValue(T const& value)
+{
+    if constexpr (std::is_same_v<T, std::string>)
+    {
+        return !value.empty();
```
alandefreitas, line 51: Do we omit all empty strings?

src/lib/Support/Reflection/MapReflectedType.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+        // Then, automatically map all FunctionSymbol-specific members.
+        mapReflectedType(io, I, domCorpus);
+    }
+    @endcode
+*/
+template <typename IO, typename T>
+    requires boost::describe::has_describe_members<T>::value
+void
+mapReflectedType(
```
alandefreitas, line 184: Nice

src/lib/Metadata/Source.cpp
```diff
@@ -5,10 +5,13 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
 // Copyright (c) 2023 Vinnie Falco (vinnie.falco@gmail.com)
+// Copyright (c) 2025 Gennaro Prota (gennaro.prota@gmail.com)
 //
 // Official repository: https://github.com/cppalliance/mrdocs
 //
 
+#include "../Reflection.hpp"
```
alandefreitas: We typically don't use dot-dot segments in mrdocs

src/lib/Metadata/Source.cpp
```diff
@@ -107,12 +94,7 @@ tag_invoke(
     IO& io,
     Location const& loc)
 {
-    io.map("fullPath", loc.FullPath);
-    io.map("shortPath", loc.ShortPath);
-    io.map("sourcePath", loc.SourcePath);
-    io.map("line", loc.LineNumber);
-    io.map("column", loc.ColumnNumber);
-    io.map("documented", loc.Documented);
+    mapReflectedType(io, loc);
```
alandefreitas, line 97: Can't we have a single template function that does that for all described types with the trait from boost.describe?

src/lib/Metadata/Symbol/Function.cpp
```diff
@@ -387,9 +369,9 @@ merge(FunctionSymbol& I, FunctionSymbol&& Other)
 {
     MRDOCS_ASSERT(canMerge(I, Other));
     merge(I.asInfo(), std::move(Other.asInfo()));
```
alandefreitas, line 371: Automatically merging metadata is another use case for reflection, by the way. (not in this PR but we should have an issue)

src/lib/Metadata/Symbol/Record.cpp
```diff
@@ -165,12 +151,10 @@ tag_invoke(
     BaseInfo const& I,
     DomCorpus const* domCorpus)
 {
-    io.map("access", I.Access);
+    mapReflectedType(io, I, domCorpus);
     io.map("isPublic", I.Access == AccessKind::Public);
```
alandefreitas, line 155: These are conveniences for handlebars because comparisons as parameters are a little expensive there, but we can also create another minor issue for these cases because there are solutions to this.

src/lib/Support/Reflection/Reflection.cpp
alandefreitas, line 1: These source and private header files related to reflection could go into a subdirectory. The natural place would be support, but maybe it could just go into metadata if we know only metadata will be described.

src/lib/Support/Reflection/Reflection.cpp
```diff
@@ -0,0 +1,527 @@
...
+void
+tag_invoke(
+    dom::LazyObjectMapTag,
+    IO& io,
+    Symbol const& I,
+    DomCorpus const* domCorpus)
+{
+    MRDOCS_ASSERT(domCorpus);
+    mapReflectedType(io, I, domCorpus);
+    io.map("class", std::string("symbol"));
```
alandefreitas, line 30: We're going to need another proper solution to this (another issue). Handlebars often need information about the class and so on, and these often clash with object members no matter what name we decide for the key. I was thinking of something like "@meta" where we would always provide a lazy representation of the class name, etc, to handlebars.

src/lib/Support/Reflection/Reflection.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+template <typename IO>
+void
+doc::tag_invoke(
+    dom::LazyObjectMapTag t,
+    IO& io,
+    doc::ImageInline const& I,
+    DomCorpus const* domCorpus)
+{
+    tag_invoke(t, io, dynamic_cast<doc::Inline const&>(I), domCorpus);
```
alandefreitas: Boost.Describe can also iterate the base classes in the function.

src/lib/Support/Reflection/Reflection.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+template <typename IO>
+void
+tag_invoke(
+    dom::LazyObjectMapTag,
+    IO& io,
+    DocComment const& I,
+    DomCorpus const* domCorpus)
+{
+    io.defer("description", [&I, domCorpus] {
```
alandefreitas: Couldn't we describe this one?

src/lib/Support/Reflection/Reflection.hpp
alandefreitas, line 1: That's really nice. Is there a way to reduce compile time (or address another issue) so that any TU that includes this header doesn't have to parse all metadata-related headers? I don't know if there's a reasonable way to forward declare the struct descriptions.

alandefreitas (commented):
That's really nice. Most comments I left are either compliments or suggestions for future issues. Regarding your question about generalizing the process and variable names, I would also leave that to another smaller issue.

.github/workflows/ci.yml
gennaroprota: I don't know. The Clang 21 + MSan test is usually taking between 1h 50m and 2h 10m. It's the only one to be so slow.

docs/mrdocs.yml
gennaroprota: They are defined by the `BOOST_DESCRIBE_STRUCT()` (or `..._CLASS()`) and `BOOST_DESCRIBE_ENUM()` macros and end up in whatever namespace is lexically enclosing the macro invocation.

include/mrdocs/Dom/LazyArray.hpp
gennaroprota: Good point. Done.

src/lib/Support/Reflection/MapReflectedType.hpp
gennaroprota: So it seems. I trusted the tests :-).

include/mrdocs/Support/EnumToString.hpp
gennaroprota: To use them in the subsequent `extern template` declarations.

include/mrdocs/Support/EnumToString.hpp
gennaroprota: The implementation in EnumToString.cpp static_asserts that the enum is described. The constraint just uses `is_enum_v`, instead, to avoid a public dependency on Boost.Describe. Yes, maybe we could put the `extern template` declarations in ExtractionMode.hpp, FunctionClass.hpp, etc., and remove EnumToString.hpp. Please let me know if you prefer that (note that EnumToString.hpp is public but the implementation is not exposed anyway).

src/lib/Metadata/Source.cpp
gennaroprota: I checked before using them :-). They were used in three places, although those are all in the test files. Please let me know if you want to change this.

.github/workflows/ci.yml
alandefreitas: OK. Well... I guess we can come up with some strategy for this later.

docs/mrdocs.yml
alandefreitas: mm... I see. My concern is this kind of pattern might be somewhat expensive to evaluate by MrDocs. If they are in whatever namespace is lexically enclosing the macro invocation, then maybe we could at least include that namespace in the excluded symbols. This way, MrDocs can filter more things more easily. Something like `whatever namespace_is_lexically_enclosing_the_macro_invocation::**::boost_base_descriptor_fn` would be much cheaper to evaluate.

src/lib/Metadata/Source.cpp
alandefreitas: I think it's best to change this because the few times I pushed `..` segments, I heard a lot of complaints from others 😆 And they're not completely wrong because when you use them, anyone reading the code loses context without looking at other things. It's simpler to describe the path relative to the directory with source files, just like all other includes.

include/mrdocs/Support/EnumToString.hpp
alandefreitas: I see nice. Maybe we can just include a brief, standard comment there so the AI knows what it's doing when it reads it. 😅 I'm happy I can blame this on AI now 😆

include/mrdocs/Support/EnumToString.hpp
alandefreitas: I just noticed GitHub is terrible at showing me the context of the comment so I can look at it again. Maybe there is a way, and I just know how to do it. In any case, I'm very confused about static assets or any compile-time assets that don't go in the public header because we simply don't have that information at compile-time no matter what implementation pattern is being used. 

If this is a public function, we don't know what user is going to call that with at compile-time. What I guess is going to happen here is this is only going to be defined for our described classes, and the user will get a weird link error otherwise. 

In any case, from the perspective of the user, this is not a general utility to convert any enum to a string. This is a utility to convert any enum related to metadata in the library to a string because the user of the public API doesn't have the tools to make this function work for anything else, and they don't really have to. 

Of course I might just be completely misunderstanding what's happening here because I lost some context.

include/mrdocs/Metadata/Symbol/NamespaceAlias.hpp
alandefreitas: In the general case, I can't think of any case in the library where we do not include information about base classes when describing the properties of an object. I can't think of anything like that happening anywhere, even hypothetically. 

Sooner or later, this function will need options anyway, because the way it has to behave and some small details are always going to change from one kind of generator to the other. So if we find out that's the case at some point in the future, that's also not a problem.

src/lib/Support/Reflection/MapReflectedType.hpp
alandefreitas: Interesting, it does make sense. At a logical level, it seems we should use strings for strings that will always be there, and optional strings otherwise (even if those are compressed optional strings, where nullified and empty strings are the same). 

But the results you got also make sense because it's very likely that all cases where we have strings are nullable strings because everything that requires a value is already being represented by enumerations and things like that.

src/lib/Support/Reflection/Reflection.cpp
gennaroprota: I vote for Support/Reflection/.

src/lib/Support/Reflection/Reflection.hpp
gennaroprota: Hmm, I'm pretty sure the `BOOST_DESCRIBE_xxx()` macros require a defining declaration. Or did you have something specific in mind to work around this?

alandefreitas: I'm not sure I understand. Do you mean a forward declaration for the specializations those macros define is something that wouldn't work? I was thinking just of forward declarations so everything could be encapsulated. And of course we would replicate the macro that generates the forward declarations we need. Instead of completely describing something, we would just forward declare the classes that describing something generates. We could also have a macro that forward declares whatever classes describing would create. Doesn't that work?

gennaroprota: I've not dug into the implementation details of Boost.Describe, but I guess it generates hidden internal structures that use pointers to members, and one cannot take a pointer to member of an incomplete type.

However, the current design already provides, I think, the separation you are asking about:

- The struct definitions live in public headers.

- The `BOOST_DESCRIBE_STRUCT()` invocations live in Reflection.hpp, which is private (src/lib/).

- Only .cpp files in src/lib/ include Reflection.hpp. And there are only four of them (not one, but a small number anyway :-)).

alandefreitas: Yes, the requirements are met. My concern was only about compilation time. We can open an issue and just refer to the comment. We can investigate this later because it's not going to be that simple anyway.

include/mrdocs/Support/EnumToString.hpp
gennaroprota: Hmm... this `toString()` is used in inline implementations of `tag_invoke()` in ExtractionMode.hpp, FileKind.hpp, FunctionClass.hpp, RecordKeyKind.hpp, Using.hpp. We can make it private if we make those implementations non-inline. Would that be better for you?

src/lib/Support/Reflection/Reflection.cpp
gennaroprota: Yes.

src/lib/Support/Reflection/Reflection.cpp
gennaroprota: Yes, done.

PR #1138, opened 2025-12-20
fix: remove duplicate ellipsis in parameter pack expansion

### LINKED ISSUE #1108: Parameter pack rendered with two `...`
The signature is rendered with two ellipses after a parameter pack argument type:

```cpp
template<
    class T,
    class... Keys,
    class... Args>
T&
emplace(Args...&&... args)
requires get_key<T>::value;
```

Minimal example:

```cpp
template<class T>
struct get_key
{
    static constexpr bool value = true;
};

template<
    class T,
    class... Keys,
    class... Args>
T&
emplace(Args&&... args)
    requires get_key<T>::value
{
    // implementation not important
    static T obj{};
    return obj;
}
```

### LINKED ISSUE #1129: Parameter pack rendered as `Args...&&...` (double `...`)
This `make_neunique` function template is being rendered with an invalid parameter pack in the generated docs.

Source:

```cpp
template<class T, class... Args>
typename std::enable_if<
    !std::is_array<T>::value,
    neunique_ptr<T>>::type
make_neunique(Args&&... args);
```

Rendered output:

```cpp
template<
    class T,
    class... Args>
neunique_ptr<T>
make_neunique(Args...&&... args)
    requires !std::is_array<T>::value;
```

The `...` appears twice around `Args`, which is not valid C++. 

```cpp
Args...&&... args
```

alandefreitas:
Only set IsPackExpansion on the outermost Result type in TypeBuilder's terminal builder functions.

### DIFF
--- a/src/lib/AST/TypeBuilder.cpp
+++ b/src/lib/AST/TypeBuilder.cpp
@@ -113,7 +113,6 @@ buildDecltype(
     MRDOCS_ASSERT(Inner);
     *Inner = Polymorphic<Type>(std::in_place_type<DecltypeType>);
     (*Inner)->Constraints = this->Constraints;
-    (*Inner)->IsPackExpansion = pack;
 
     auto &I = (*Inner)->asDecltype();
     getASTVisitor().populate(I.Operand, T->getUnderlyingExpr());
@@ -134,7 +133,6 @@ buildAuto(
     MRDOCS_ASSERT(Inner);
     *Inner = Polymorphic<Type>(std::in_place_type<AutoType>);
     (*Inner)->Constraints = this->Constraints;
-    (*Inner)->IsPackExpansion = pack;
 
     auto &I = (*Inner)->asAuto();
     I.IsConst = quals & clang::Qualifiers::Const;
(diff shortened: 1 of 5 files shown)


alandefreitas:
@mizvekov Have a look at

https://github.com/cppalliance/mrdocs/pull/1138/checks?check_run_id=58598186176 and https://github.com/cppalliance/mrdocs/pull/1138/checks?check_run_id=58598186580

I think codecov assumes lower relative coverage always means new code hasn't been covered, which is true for most cases. 

However, it's impossible for absolute coverage to decrease in this PR because it only removes code. 

But one of these jobs says

> 86.17% (-0.02%) compared to 3462996

and then fails. I have no idea where this is coming from. 😅

mizvekov:
> @mizvekov Have a look at
> and then fails. I have no idea where this is coming from. 😅

See https://app.codecov.io/gh/cppalliance/mrdocs/pull/1138/indirect-changes
And see the `src/lib/Support/Chrono.hpp` coverage changes.

But this looks like noise from a test run that took more than an hour this time.

PR #1151, opened 2026-01-21
fix: replace the XML generator with a reflection-based one

gennaroprota:
@alandefreitas: Please let me know what you think and... if I have commented it too much :-).


alandefreitas:
I think we should be removing way more code than including for this PR to make sense. The test outputs haven't even changed. This should be a bug fix, not a refactor. The biggest problem with this generator is that it's not in sync with the metadata. And the reason this generator is out of sync is that we didn't have reflection, and manually keeping it in sync was too expensive and fragile. So what we want to do is let reflection handle the whole thing, and the whole generator will have only one or two functions that recursively iterate over everything to generate the metadata. 🥳

This means the XML outputs will change completely, as they now reflect everything in the metadata, which is what this generator was meant for. Our intention is not to recreate the broken version, but to recreate it with reflection. If we don't fix it, reflection would mean more code rather than less, as we now have to force reflection to recreate the errors we had without it (which is even harder than just keeping the original errors). And that's going to be extra code just to maintain it broken. 

I think another issue here is that we should probably address the other issue first, where we should remove those small workarounds from reflection to make it reflect into different key names. Only canonical names should be used. All generators should have access to the same fields with the same names. Making this happen is basically just the finalization of the previous issue we already merged. And maybe the @meta field is required to make that happen, since ambiguity might arise. Only after establishing the canonical field names can we use them to generate XML. 😅

alandefreitas:
What happened to this PR? I thought it would become more active after #1153 🚀

gennaroprota:
I removed the commit before pulling, after merging #1153, and GitHub closed it. I shouldn't have used the `develop` branch.

alandefreitas:
Oh 😅

.github/workflows/ci.yml
```diff
@@ -609,14 +611,29 @@ jobs:
...
+          # need test targets built so the reduced ctest step can run a
+          # subset. The cmake-workflow action sets
+          #   -D <configure-tests-flag>=OFF
+          # when run-tests is false, which would turn off BUILD_TESTING
+          # and thus MRDOCS_BUILD_TESTS. We redirect the flag to a dummy
+          # variable so BUILD_TESTING is never explicitly set, and force
+          # MRDOCS_BUILD_TESTS=ON unconditionally to ensure tests are
+          # always compiled.
+          run-tests: ${{ !matrix.is-bottleneck }}
+          configure-tests-flag: ${{ matrix.is-bottleneck && 'MRDOCS_RUN_ALL_TESTS' || 'BUILD_TESTING' }}
```
alandefreitas: This logic is great and correct. It can be simplified, though. 

The first issue is that the code hardcodes MRDOCS_BUILD_TESTS=ON (which already makes BUILD_TESTING a noop because its only role is to set MRDOCS_BUILD_TESTS), then this is what you have for everything else:

| Is Bottleneck | Flag (tests-flag) | Value (run tests) |
|----------|----------|----------|
| True    | MRDOCS_RUN_ALL_TESTS   | True |
| False    | BUILD_TESTING    |  False |

The true path is noop because there's no MRDOCS_RUN_ALL_TESTS in CMakeLists.txt, unless this was included in another commit. The false path is also a noop anyway because of this table. So this code makes BUILD_TESTING irrelevant twice. Omitting everything would have the same result. No need for `-D MRDOCS_BUILD_TESTS=ON`, no need for custom tests flag (see issue 2), and no need for custom run-tests (see third issue).

The second issue is that once tests are forced to True, the infrastructure is repurposed to specify the test flag for something else and rely on the action to circumstantially turn it on. Instead of relying on this chain of events to turn this other flag on and off, you can just specify it directly in extra-args as `-D MRDOCS_RUN_ALL_TESTS=${{ matrix.is-bottleneck && 'OFF' || 'ON' }}` (or come up with a more representative name). 

The third issue is that the logic reads as "I don't want to run tests," and then it creates another step that says "since I didn't run the tests before, let's run them now". But you don't need a separate step. If the extra-args already only built and registered the tests you need, then you already have only the ones you need and you already ran them.

To summarize, all you need is some `-D MRDOCS_EXPENSIVE_TESTS=${{ matrix.is-bottleneck && 'OFF' || 'ON' }}` (a single line changes in ci.yml) and in CMakeLists.txt skip registering (`add_test`) expensive tests with `if (NOT MRDOCS_EXPENSIVE_TESTS)`.

gennaroprota: Yes, my solution was needlessly convoluted. Thanks for steering me the right way.

Issue #1148, opened 2026-01-20
feat(dom): add `@meta` key to all objects for type identification


gennaroprota:
We should add a `@meta` key to all DOM objects that describes the class name and other metadata.

### Context

As noted by @alandefreitas in PR#1130:

> Sometimes we get conflicts if we don't rename things for Handlebars, as you notice with "class" and "functionClass", etc. This is because of how handlebars works with recursion, and it's very annoying. [...] one solution I was going to propose at some point is adding an "@meta" key to all objects to describe the class name, etc (this can also help with other problems we've been having). This way, we would have the proper information to make the distinction in handlebars.

### Proposed solution

Each DOM object would include a `@meta` key containing:

- `type` or `className`: The C++ class name.
- Possibly other metadata useful for template logic.

### Benefits

- Eliminates naming conflicts in recursive Handlebars contexts.
- Enables templates to make type-based decisions without relying on the presence/absence of specific keys.
- Provides a consistent pattern for type introspection in templates.

PR #1153, opened 2026-01-23
feat: add $meta.type and $meta.bases to all DOM objects

gennaroprota:
Closes issue #1148.

### DIFF
--- a/src/lib/Support/Reflection/MapReflectedType.hpp
+++ b/src/lib/Support/Reflection/MapReflectedType.hpp
@@ -83,36 +83,12 @@ inline
 std::string
 normalizeMemberName(std::string_view name)
 {
-    // Special cases.
-    if (name == "Constexpr")
+    std::string result(name);
+    if (!result.empty())
     {
-        return "constexprKind";
+        result.front() = std::tolower(result.front(), std::locale::classic());
     }
-    else if (name == "ReturnType")
-    {
-        return "return";
-    }
-    else if (name == "Noexcept")
-    {
-        return "exceptionSpec";
-    }
-    else if (name == "Explicit")
-    {
-        return "explicitSpec";
-    }
-    else if (name == "KeyKind")
-    {
-        return "tag";
-    }
-    else
-    {
-        std::string result(name);
-        if (!result.empty())
-        {
-            result.front() = std::tolower(result.front(), std::locale::classic());
-        }
-        return result;
-    }
+    return result;
 }
 
 /** Remove namespace qualifiers from a type name.
--- a/src/lib/Support/Reflection/MapReflectedType.hpp
+++ b/src/lib/Support/Reflection/MapReflectedType.hpp
@@ -11,6 +11,7 @@
 #ifndef MRDOCS_LIB_SUPPORT_REFLECTION_MAPREFLECTEDTYPE_HPP
 #define MRDOCS_LIB_SUPPORT_REFLECTION_MAPREFLECTEDTYPE_HPP
 
+#include <mrdocs/Dom/Array.hpp>
 #include <mrdocs/Dom/LazyArray.hpp>
 #include <mrdocs/Metadata/Expression.hpp>
 #include <mrdocs/Metadata/Specifiers/ConstexprKind.hpp>
@@ -103,10 +104,6 @@ normalizeMemberName(std::string_view name)
     {
         return "tag";
     }
-    else if (name == "Class")
-    {
-        return "usingClass";
-    }
     else
     {
         std::string result(name);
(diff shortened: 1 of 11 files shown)


src/lib/Support/Reflection/MapReflectedType.hpp
```diff
@@ -82,40 +82,93 @@ inline
...
+/** Get the unqualified type name at compile time.
+
+    Uses compiler-specific intrinsics to extract the type name
+    from __PRETTY_FUNCTION__ (GCC/Clang) or __FUNCSIG__ (MSVC).
+
+    E.g.: readableTypeName<FunctionSymbol>() -> "FunctionSymbol"
+*/
+template <typename T>
+constexpr std::string_view
+readableTypeName()
```
alandefreitas, line 118: Isn't this information already boost described?

src/lib/Support/Reflection/MapReflectedType.hpp
```diff
@@ -170,22 +241,26 @@ mapMember(
...
+template <typename IO, typename T, bool isMostDerived = false>
     requires boost::describe::has_describe_members<T>::value
 void
 mapReflectedType(
     IO& io,
     T const& obj,
     DomCorpus const* domCorpus)
 {
+    // Add _meta only for the most-derived type.
+    if constexpr (isMostDerived)
```
alandefreitas, line 304: Oh... I never thought about this problem. There are some cases where the templates are worried about the base class. For instance, some fields return whether something is a symbol, and the templates often query that, but the derived class would just be a record, a function or something like that. I don't know what the best solution for that would be. Maybe we just include the metadata for the superclass inside the metadata for the class.

alandefreitas (commented):
LGTM :) 

What was the process to get to this `_meta` key?

gennaroprota:
> LGTM :)
> 
> What was the process to get to this `_meta` key?

You mean what led to choosing the underscore as prefix?

src/lib/Support/Reflection/MapReflectedType.hpp
gennaroprota: AFAICS, Boost.Describe cannot provide the type name. However, we could use Boost.Core:

```
template <typename T>
std::string
readableTypeName()
{
    return std::string(removeNamespaceQualifiers(boost::core::type_name<T>()));
}
```

src/lib/Support/Reflection/MapReflectedType.hpp
gennaroprota: Hmm... if there are already templates that do that, why didn't the golden tests fail?

alandefreitas:
> You mean what led to choosing the underscore as prefix?

I mean the whole thing. This combination of prefix + keyword. Did we compare it with existing practices for that anywhere? Or the set of valid chars that could go there? Getting this convention wrong now could be annoying later.

gennaroprota:
> > You mean what led to choosing the underscore as prefix?
> 
> I mean the whole thing. This combination of prefix + keyword. Did we compare it with existing practices for that anywhere? Or the set of valid chars that could go there? Getting this convention wrong now could be annoying later.

I was going with `@meta`, which you suggested, but `@` is not allowed. So I considered a different prefix. It didn't seem something to spend too much time on (maybe I'm wrong), so I considered `$` and `_`, and quickly chose the latter (which I like more, maybe because I use C++ more than PHP :-)).

`_meta` also mirrors the OData v2/v3 `__metadata` convention, which was specifically designed for the same purpose: storing entity metadata alongside properties.

src/lib/Support/Reflection/MapReflectedType.hpp
alandefreitas: Is Boost.Core already available?

gennaroprota: No, we'd have to add the (private) dependency.

src/lib/Support/Reflection/MapReflectedType.hpp
alandefreitas: Just to be clear we're not misunderstanting each other, what I meant in practice is that templates often need to do something like `{{#if (eq symbol.type "symbol")}}do this{{ else if (eq symbol.type "name")}}do that{{else}}error!{{/if}}` because we have two ways to talk about identifiers (Symbol and Name) and we want the template to work for both of them.

- https://github.com/cppalliance/mrdocs/blob/develop/include/mrdocs/Metadata/Symbol.hpp
- https://github.com/cppalliance/mrdocs/blob/develop/include/mrdocs/Metadata/Name.hpp

So the problem is the current model doesn't give us the parent class. So both conditions would return false here and we would get into error. I don't know about how to fix that. But only representing the most derived type would fail.

> Hmm... if there are already templates that do that, why didn't the golden tests fail?

Because the previous code wasn't broken and the new code is not relying on $meta for this distinction yet? So it's still using the previous workaround, which doesn't need $meta. Probably the type is represented with "type" and the parent type is represented with "kind" or something. Whatever the describe function does for Symbol and Name.

gennaroprota: I also added `$meta.bases`. Please let me know if you like it.

alandefreitas: LGTM :)

src/lib/Support/Reflection/MapReflectedType.hpp
alandefreitas: Then nevermind :)

Issue #1156, opened 2026-02-04
Capy: Markdown inline formatting and bullet lists not recognized


gennaroprota:
In the documentation of `boost::capy::any_buffer_sink` (<https://develop.capy.cpp.al/capy/reference/boost/capy/any_buffer_sink.html>), the bullet list at the end of the "Description" paragraph is not rendered as such, and no bold text is produced in the list items. Similarly for `boost::capy::strand` ("Executor Concept" paragraph), `boost::capy::DynamicBufferParam` ("Description" paragraph) and `boost::capy::IoAwaitable` ("Syntactic Requirements" paragraph). Code spans are not recognized either.

Issue #1113, opened 2025-11-24
Explore unknowns


alandefreitas:
Based on the internal discussion held on October 21st, the project will concentrate on the following areas of uncertainty:

- [x] Mrdocs-as-compiler #1073
- [x] Scripting extensions #1128 #881
- [x] Plugins  #58 #1044 
- [x] JSON-only mrdocs 
- [x] Reflection #1114
- [x] Cross-linking #1072

This issue serves as a reference for the list and a platform for more lasting discussions about these topics.

alandefreitas:
**Update**: The idea of a MrDocs version that only outputs JSON has been discussed internally this week. I'm adding this to the list of unknowns because it's independent from the idea of MrDocs as a compiler, even though Peter Dimov proposed both at the same time. 

This unknown is similar to #678, but the pros and cons are inevitably different because some of the cons in #678 no longer apply, and we have other unknowns that could change the incentives here.

Also, there's some urgency to discussing this one because it completely changes the meaning of the unknown regarding extensions.

alandefreitas:
**Update**

@vinniefalco @mizvekov @gennaroprota 

After investigating these unknowns for a little longer, here's my assessment of each unknown, ordered by their blocking relationship (and therefore priority):

- JSON-only mrdocs: It makes sense, but it depends on how extensions go. Tools can focus on what they're good at. Extensions become much more powerful post-hooks. Extensions can do anything we think of without bloating the binary. It's much easier to find devs for these extensions because they don't even need to know Clang/LLVM (though they do need to know some basic C++). Modularization/encapsulation improves. Different steps would have their own tests. The performance cons in #678 also no longer apply. It enforces good internal boundaries and encapsulation. If we want to keep the XML generator or even create a binary generator (as it would be useful for a mrdocs-as-compiler feature), that's also fine. Although we're calling this "JSON-only mrdocs", the relevant feature is actually "serializable-structured-data-formats-that-are-low-effort-to-maintain-only mrdocs".
[...]

PR #1157, opened 2026-02-06
fix: rendering of Markdown inline formatting and bullet lists (#1156)

gennaroprota:
**Note**: The comments below are out of date. Please refer to the PR commits for up to date info.

@alandefreitas: This patch covers the current Capy needs but, as I said on Slack, it is ad-hoc. The only long term solution is what Vinnie suggested: a Clang hook to take over comment processing entirely. A bit reluctantly, given the time I spent on it, I'd say we shouldn't merge it. We can extract the fix to inline.html.hbs from it, though, which is a genuine bug fix and doesn't depend on the Clang hook.

---

Markdown inline formatting (**bold**, *italic*, `code`) and bullet lists using "- " markers were not rendered in the HTML output.

Inline formatting:

- Add `appendMarkdownInlines()` to parse **bold**, *italic*, and `code` spans in text nodes, producing `StrongInline`, `EmphInline`, and `CodeInline` nodes respectively.

- Call it from `visitText()` so all text nodes are processed.

- Add a markup/b.html.hbs partial to render `<strong>` tags.

- Add doc/inline.html.hbs to dispatch text, strong, emph, and code inline kinds to their partials without emitting extraneous blank lines.

List detection and conversion:

- Add list marker detection functions that recognize "- " at the start of text, after newlines, after ":" or "." punctuation, and at trailing positions within already-started lists.

- Add `convertParagraphWithLists()` to split a paragraph containing list markers into a prefix paragraph and a `ListBlock`.

- Apply list detection in `visitParagraph()`, `@par` blocks, and `@li` blocks so that Markdown lists are converted in all contexts.

### DIFF
--- a/include/mrdocs/ADT/Optional.hpp
+++ b/include/mrdocs/ADT/Optional.hpp
@@ -962,6 +962,8 @@ public:
     constexpr
     bool
     /** Return true because Optional<T&> never allocates storage.
+
+        @return `true` always.
     */
     is_inlined() noexcept
     {
--- a/include/mrdocs/Metadata/DocComment.hpp
+++ b/include/mrdocs/Metadata/DocComment.hpp
@@ -227,6 +227,16 @@ requires std::same_as<T, doc::Block> || std::same_as<T, doc::Inline>
 void
 traverseImpl(T& baseNode, F&& func);
 
+template <bool bottomUp, class F>
+void
+traverseImpl(doc::ListItem& item, F&& func)
+{
+    for (auto& block : item.blocks)
+    {
+        traverseImpl<bottomUp>(block, func);
+    }
+}
+
 // Traverse a derived node
 template <bool bottomUp, DocCommentNode NodeTy, class F>
 requires(!std::same_as<NodeTy, doc::Block> && !std::same_as<NodeTy, doc::Inline>)
@@ -255,6 +265,15 @@ traverseImpl(NodeTy& N, F&& func)
     {
         traverseImpl<bottomUp>(N.exception, func);
     }
+
+    if constexpr (std::same_as<NodeTy, doc::ListBlock>)
+    {
+        for (auto& item: N.items)
+        {
+            traverseImpl<bottomUp>(item, func);
+        }
+    }
+
     if constexpr (bottomUp && std::invocable<F, NodeTy&>)
     {
         func(N);
--- a/src/lib/Metadata/Finalizers/DocCommentFinalizer.cpp
+++ b/src/lib/Metadata/Finalizers/DocCommentFinalizer.cpp
@@ -5,6 +5,7 @@
 //
 // Copyright (c) 2023 Krystian Stasiowski (sdkrystian@gmail.com)
 // Copyright (c) 2025 Alan de Freitas (alandefreitas@gmail.com)
+// Copyright (c) 2026 Gennaro Prota (gennaro.prota@gmail.com)
 //
 // Official repository: https://github.com/cppalliance/mrdocs
 //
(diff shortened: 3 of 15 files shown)


alandefreitas:
> The only long-term solution is what Vinnie suggested: a Clang hook to take over comment processing entirely

Yes. In theory, we could just bypass clang and parse everything ourselves, and this would be somewhat equivalent to a hook. The biggest issue blocking this right now is human resources: the current plan is to address #1113 first (although we're opening this exception for capy issues). We'll have reasonable solutions for the parsing problem once we get it. 😊

alandefreitas:
Oh... Regarding your solution, it looks a little verbose (and potentially duplicated), to be honest. 😅

We already parse Markdown elsewhere in the finalize step; this is completely unrelated to the clang parser. The algorithm takes an array of text nodes and converts them inline into nodes with the proper tags (there are even unit tests for that). We just need to complete the plumbing so that the list items also work with this functionality. It shouldn't be much code. 

The impression I have is that whatever procedure you used to figure out what to change wasn't aware that the logic already exists in the codebase. And that's understandable, because this codebase has become extremely large and complex, and it's hard to understand the context in which each thing happens. 🙃 

But we can come up with a reasonable solution to this problem. It's not that bad. Oh... and we also need tests, of course. 🙂

include/mrdocs/ADT/Optional.hpp
```diff
@@ -962,6 +962,8 @@ class Optional<T&> {
     constexpr
     bool
     /** Return true because Optional<T&> never allocates storage.
+
+        @return `true` always.
```
alandefreitas, line 966: Not your fault but... weird position for this doc comment.

include/mrdocs/Metadata/DocComment.hpp
```diff
@@ -255,6 +255,18 @@ traverseImpl(NodeTy& N, F&& func)
     {
         traverseImpl<bottomUp>(N.exception, func);
     }
+
+    if constexpr (std::same_as<NodeTy, doc::ListBlock>)
+    {
+        for (auto& item: N.items)
+        {
+            for (auto& block: item.blocks)
```
alandefreitas: traverseImpl<bottomUp>(item, func);

instead of going directly to the child.

src/lib/Gen/xml/XMLWriter.cpp
```diff
@@ -682,14 +682,20 @@ writeListItem(
     doc::ListItem const& node)
 {
     tags_.open("listitem");
-    // flatten blocks to write their inlines directly when possible
-    // this maintains the schema, but it would be best to
-    // just render the blocks directly
+    // Flatten inline containers to write their inlines directly,
```
alandefreitas: Oh... XML got updated. Good. Luckily we won't need to maintain any of this soon :)

src/lib/Metadata/Finalizers/DocCommentFinalizer.cpp
```diff
@@ -107,6 +262,14 @@ build()
         processRelates(I);
     }
 
+    // Convert Markdown-style bullet lists in text nodes.
+    // Must run before normalizeSiblings() merges per-line text nodes.
+    for (auto& I : infos)
+    {
+        MRDOCS_CHECK_OR_CONTINUE(I.doc);
+        convertMarkdownLists(*I.doc);
```
alandefreitas: This creates a new Markdown rendering pass instead of integrating with the existing finalizer Markdown rendering pass (IIRC, parseInlines, right?). Doing this is problematic, duplicates logic, is less efficient, and hard to maintain. Especially if it comes before the first pass, because we're trying to parse internal Markdown nodes before external ones.

alandefreitas (commented):
Great

include/mrdocs/Metadata/DocComment.hpp
gennaroprota: Yep, thanks.

src/lib/Metadata/Finalizers/DocCommentFinalizer.cpp
gennaroprota: Hmm... I moved the call to `convertMarkdownListsInBlocks()` into `parseInlines()`, so there's a single Markdown processing step now. The list conversion runs first within that function because it's a block-level restructure (splitting paragraphs into `ListBlock` nodes) that needs line boundaries intact, while `parse()` is a pure inline parser that operates on text strings. I think extending `parse()` to handle lists wouldn't work since it has no access to block-level structure. Does that make sense?

alandefreitas: Ideally, parseInlines would use an algorithm similar to the one in the CommonMark specification: first look for blocks, then for inlines and other blocks inside blocks, then inlines inside inlines, etc. It just becomes recursion after the first level of blocks. 

The only difference is we have this mess: we don't have string_view as input because another algorithm already tried to parse it, so we have to keep interpreting sequences of nodes as a single string, and the code becomes spaghetti. 🍝🤣 The spaguetti also makes the first level confusing, because another algorithm already parsed it, so we're already operating at the block level, because the other algorithm does find blocks. 

Anyhow, you're correct to assume lists have higher precedence because this is a block level. But lists are special nodes in CommonMark. It seems a little confusing at first. A list is a block level container, like other block level containers so they go in the higher level of the algorithm. But they contain items instead of inlines (it's a variation but this kind of variation is not so uncommon). But then each list item is also a block container (rather than a container of inlines) and this is the big variation. So when there's only text in the list item, that's actually a paragraph container in the list item. 

For instance, this is a list with one list item with 4 blocks:

```
- This is a paragraph.

  This is another paragraph.

  > A block quote

  - A nested list
```

And this is a list with one list item with 1 block (a paragraph).

```
- Just one line
```

The algorithm is derived naturally from that:

```
Document
└── List (block)
    ├── ListItem (block container)
    │   └── Paragraph (block)
    └── ListItem (block container)
        ├── Paragraph (block)
        └── List (block)
```

As usual, what makes our case complex is that we're actually (re)parsing nodes rather than strings, which is much more complex. At some point, we should probably just store the strings in the first step and let the post-processing step do all of the work. Because the clang parsing step is parsing only low hanging fruits (blocks that are `@ref and some text`) while making the post-processing step much harder to reparse, which is the step that actually understands ~80% of the features we want and have open issues about. (@mizvekov this is once again related to the topic we were discussing this week)

gennaroprota: Thanks for the context — that all makes sense. Just to be clear: are you happy to merge this as-is (with the understanding that the long-term fix is moving to raw strings in the finalizer), or would you rather hold off and address the Capy list rendering as part of that larger rework?

include/mrdocs/ADT/Optional.hpp
gennaroprota: Yeah, several functions in Optional.hpp have the doc comment in that position.

src/lib/Metadata/Finalizers/DocCommentFinalizer.cpp
alandefreitas: Not really. Perhaps I tried to provide too much detail about these layers work and didn't make myself clear. 😅

To be short and clear: The design I described (container of blocks -> containers of inlines/blocks) is not only valid for raw strings. It is also valid for reparsing. And it is also the model that the code is **currently** attempting to implement in this 2nd parsing layer. 

Thus, to be clear, I would **not** like this pattern to be broken by an implementation that bypasses it and introduces an additional 3rd parsing layer. That's what the previous version of the code did. Or I would not like any new code "pretending" to integrate with the 2nd layer, like a hypothetical function that calls a single function to do all the markdownListParsing from somewhere, even if that somewhere is inside the function that triggers the 2nd layer parsing, of course. 🙃

And we don't want that because if we want to merge these two layers into a single layer soon, we do NOT want to create a 3rd layer now. Also, because duplicating code is always a bad idea. 😁

> are you happy to merge this as-is

I don't know. I haven't read the new version of the code yet. I was hoping you could let me know whether the new code meets these criteria we've discussed and how it does it before I review it. This way, there's no need to put effort into reversing engineering code if you can just tell me how you integrated the layers. 😁

> the long-term fix is moving to raw strings in the finalizer

Yes. That would be great. But it's mostly unrelated to the issue described in this review thread. It's just something I mentioned in passing, so we have it in mind when integrating components into the second layer. It's not blocking or blocked by this issue.

gennaroprota: Yes, my solution used a 3rd layer. Thanks, I now understand your concerns.

Here's my analysis of the situation:

The list splitting **must** run on raw text nodes, before inline parsing, because `parse()` merges text across line boundaries. For example, the raw text "\`foo()\` - A dummy function" is a single TextInline — if inline parsing runs first, the backticks get parsed into a CodeInline, and the leftover TextInline(" - A dummy function.") starts with "- " after trim, causing a false-positive list marker match. So the ordering constraint is: list splitting → recurse into children → inline parsing, top-down.

The existing `bottomUpTraverse()` visits children before parents (bottom-up), which is the opposite of what list splitting needs. This means I can't just add list splitting into the bottomUpTraverse visitor — the ordering would be wrong.

I see two ways to do this as genuinely one pass:

Option A: Add a `topDownTraverseBlocks()` utility alongside `bottomUpTraverse()`.
It would use the same requires-clause pattern that `bottomUpTraverse()` already uses to discover `.blocks`, `.Document`, etc., but visit in top-down order. Then `parseInlines()` becomes:

```
void
DocCommentFinalizer::parseInlines(DocComment& doc)
{
    topDownTraverseBlocks(doc, [](auto& blocks) {
        splitParagraphsAtMarkers(blocks);
        for (auto& block : blocks)
        {
            auto* ic = dynamic_cast<doc::InlineContainer*>(&*block);
            if (ic)
            {
                parseInlinesInContainer(*ic);
            }
        }
    });
    if (doc.brief)
    {
        parseInlinesInContainer(*doc.brief);
    }
}
```

This automatically covers future additions and is genuinely one pass. The downside is adding a new traversal utility, but it mirrors the existing `bottomUpTraverse()` pattern.

Option B: Extend `bottomUpTraverse()` with a pre-visit hook.
Add a top-down pre-visitor parameter to `bottomUpTraverse()` so that list splitting runs before children are visited, and inline parsing runs in the existing post-visitor. Both happen in one traversal, and it automatically covers the full DocComment tree. The downside is modifying shared infrastructure.

alandefreitas: Maybe we're talking past each other, because your comment implies relationships among many of these moving parts that aren't obvious to me. In fact, some of these relationships ("list splitting (...) before inline parsing") are actually exactly what I said (that's what CommonMark and blocks before inlines mean). In any case, I don't want to block your PR because I'm unfamiliar with these details. As long as you understand my concern, I'll trust your judgment. 🙂

gennaroprota: Thanks Alan! To confirm: I've restructured the list splitting so it's integrated into the existing 2nd layer rather than being a separate traversal. The implementation uses a single `topDownTraverse()` call where the visitor does both list splitting and inline parsing at each node. Please have a look :-).

alandefreitas: Amazing. Thank you!

PR #1163, opened 2026-03-03
feat: add function object support

### LINKED ISSUE #564: identify niebloids
## Motivation

Many libraries, such as Boost.URL and Boost.Buffers include Niebloids. For instance:

https://github.com/cppalliance/buffers/blob/develop/include/boost/buffers/algorithm.hpp#L182

https://github.com/boostorg/url/blob/6c48a8c4e133d3b8d5b7113e6d8cae0ac48293cd/include/boost/url/grammar/digit_chars.hpp#L78

## Definition

MrDocs needs a strategy or feature to support documenting Niebloids. From the standard:

> The entities defined in the std​::​ranges namespace in this Clause are not found by argument-dependent name lookup ([basic.lookup.argdep](http://eel.is/c++draft/basic.lookup.argdep)). When found by unqualified ([basic.lookup.unqual](http://eel.is/c++draft/basic.lookup.unqual)) name lookup for the postfix-expression in a function call, they inhibit argument-dependent name lookup.
> 
> void foo() {
>   using namespace std::ranges;
>   std::vector<int> vec{1,2,3};
>   find(begin(vec), end(vec), 2);        // #1
> }
> The function call expression at #1 invokes std​::​ranges​::​find, not std​::​find, despite that (a) the iterator type returned from begin(vec) and end(vec) may be associated with namespace std and (b) std​::​find is more specialized ([temp.func.order]) than std​::​ranges​::​find since the former requires its first two parameters to have the same type.

Eric Niebler himself [suggested the name](https://twitter.com/tcanens/status/1063723279414890496?s=20) and wrote an [article](http://ericniebler.com/2014/10/21/customization-point-design-in-c11-and-beyond/) in 2014 explaining this concept.

[Cppreference](https://en.cppreference.com/w/cpp/algorithm/ranges) describes it as

> The function-like entities described on this page are niebloids, that is:
> 
> - Explicit template argument lists may not be specified when calling any of them.
> - None of them is visible to argument-dependent lookup.
> - When one of them is found by normal unqualified lookup for the name to the left of the function-call operator, it inhibits argument-dependent lookup.

## Properties of Niebloids

An inline variable pair and a functor are considered a niebloid if they have the following [properties](https://quuxplusone.github.io/blog/2019/08/02/the-tough-guide-to-cpp-acronyms/#cpo):

> ... a CPO is an object (not a function); it’s callable; it’s constexpr-constructible, [...] it’s customizable (that’s what it means to “interact with program-defined types”); and it’s concept-constrained.
> [...]
> If you remove the adjectives “customizable, concept-constrained” from the above, then you have a function object that turns off ADL — but is not necessarily a customization point. The C++2a Ranges algorithms, such as std::ranges::find, are like this. Any callable, constexpr-constructible object is colloquially known as a “niebloid,” in honor of Eric Niebler.

MrDocs can use these properties to attempt to automatically identify Niebloids.

## Niebloids in cppreference

Cppreference also documents niebloids as it own thing (not as `class/struct` or `function` overloads):

image

When we look at the [documentation](https://en.cppreference.com/w/cpp/algorithm/ranges/all_any_none_of) of these niebloids, the documentation is very similar to function overloads where the function names come from the inline variable and the function arguments and return type come from the functor. Pages documenting niebloids usually end with:

image

It's important to note the struct defining the niebloid and the constexpr inline variable are aggregated and documented as the same function-like entity (the niebloid). There's no individual page for the variable and for the functor.

## Niebloids in Doxygen 

Libraries such as range-v3 attempt to document niebloids as function overloads but the code is a mess. 

https://github.com/ericniebler/range-v3/blob/53c40dd628450c977ee1558285ff43e0613fa7a9/include/range/v3/algorithm/copy.hpp#L40

It depends on _lots_ of macros so that these symbols are identified as classes by the compiler and identified as function overloads by doxygen. The code becomes very hard to read.

## Proposed solutions

Neither doxygen nor mrdocs currently have any feature to make that easier. I think niebloids are going to be the first feature where we can confidently say mrdocs does something doxygen can’t do. 

MrDocs could implement a combination of the following solutions:

- When MrDocs builds the corpus, as a post-processing step, it would aggregate pairs of variables and functors that form Niebloids according to the properties above (constexpr inline variable + constexpr-constructible callable object whose all public members are overloads of `operator()`). A combination of these properties is very unlikely to be a false positive.
- This automatic detection could be disabled by a config option, in which case a special command would need to be included in the inline variable as a hint to identify it as a Niebloid. Although automatic detection should work most of the time (considering how specific the properties of Niebloids are), disabling automatic detection might be useful when the difference between a regular functor and a Niebloid might be a difference in intentionality. For instance, `std::hash<T>` shouldn't be documented as a niebloid even if it's a constexpr-constructible callable object whose all public members are overloads of `operator()` and there's a constexpr inline variable that instantiates it somewhere else.
- These pairs of inline variables and objects could be moved to a function overload set with an extra flag to identify it as a Niebloid or to a new Info type for Niebloids. The name of the object will come from the inline variable. In any case, the type should be compatible with function overload sets. The new Info type might inherit from the type for function overload sets.
- The documentation template pages for niebloids should include a message similar to the one in cppreference. Everything else would look similar to functions and function overload sets.

This is a general idea. Considering the complexity of the problem, implementation requirements can only be completely identified when we start working on it. We will certainly find more obstacles.

gennaroprota:
**Edit**: This description is now obsolete. Please refer to the commit messages for up to date info.

This PR adds detection of constexpr variables whose type is a record providing only operator() overloads, and documents them as callable objects in the Functions tranche. The type of the variable is hidden as implementation-defined, and breadcrumbs render in f() style for readability, but operator() signatures remain fully visible in the variable synopsis and each overload gets its own detail page---preserving the fact that the callable is a function object, not an ordinary function.

Detection triggers:

- Auto-detection (on by default, controlled by `auto-function-objects`) for records whose only public non-special members are operator() overloads, including class templates when the variable lives in an enclosing scope of the type.

- The `@function_object` documentation command for cases that fail auto-detection (e.g. types with extra members).

Key implementation details:

- FunctionObjectFinalizer runs before OverloadsFinalizer: hides the type of the variable, restores operator() to Regular extraction, re-parents overloads under the variable, and appends a fixed disclaimer note.

- VariableSymbol gains FunctionObjectOverloads and satisfies the SymbolParent concept, so MultiPageVisitor generates detail pages for each operator() overload.

- Handlebars templates render linked operator() signatures in the variable synopsis with "» more..." links to the detail pages.

Closes #564.


alandefreitas:
How does this solution relate to the proposals in https://github.com/cppalliance/mrdocs/issues/564? 🙂

(btw, #564 is from Apr 10, 2024 - it's an investigation that took a long time - not 2026 AI slop 😂)

Edit: nevermind. I think I already got it. :)

include/mrdocs/Metadata/Symbol/Variable.hpp
```diff
@@ -101,6 +102,22 @@ struct VariableSymbol final
     */
     ConstantExprInfo<std::uint64_t> BitfieldWidth;
 
+    /** Whether this variable is a function object.
+
+        A function object variable is a constexpr variable whose type
```
alandefreitas: I know I'm nitpicking, but I don't think it has to be `constexpr`.

share/mrdocs/addons/generator/common/partials/symbol/qualified-name-text.hbs
```diff
@@ -11,9 +11,11 @@
     See: https://mrdocs.com/docs/mrdocs/develop/generators.html#dom_reference
 --}}
 {{~#if (and parent parent.parent)~}}
-{{~> symbol/qualified-name-text parent ~}}::
+{{~> symbol/qualified-name-text parent ~}}{{~#unless parent.isFunctionObject~}}::{{~/unless~}}
 {{~/if~}}
-{{~#if name ~}}
+{{~#if parent.isFunctionObject ~}}
+()
```
alandefreitas: If functions don't have "()" as part of their qualified name, why should function objects have it?

include/mrdocs/Metadata/Symbol/Variable.hpp
```diff
@@ -101,6 +102,22 @@ struct VariableSymbol final
...
+        is a class or struct providing only operator() overloads. When true,
+        the variable is documented as a callable, and its type is hidden
+        as implementation-defined.
+    */
+    bool IsFunctionObject = false;
```
alandefreitas: Mmmm... I often got the recommendation to use a single `bool IsFunctionObject` in the variable, rather than creating a new symbol type for that. That is valid, but it's not anything but simple. For instance, 

- How do I know if something is only the "impl" of a function object? Does it also get a boolean?
- And if I don't, how do I remove it from indices in the end? 
- And how do we ensure the generator doesn't generate a page for them?
- And since function objects have a completely different way of being presented to the user, will we create a new template with ways of presenting them to the user that look more like functions (signature templates, etc), or will we sprinkle a lot of if/else throughout the templates?
- How do we get the impl object from the variable?
- If we don't, how do we render information from the impl object in the variable page if we have to? That could happen very often.

I'll probably get some of these answers as I review the PR, but I'm leaving them here so I don't lose context.

include/mrdocs/Metadata/Symbol/Variable.hpp
```diff
@@ -116,6 +133,20 @@ struct VariableSymbol final
...
+    For non-function-object variables this returns an empty range,
+    so the symbol is effectively a leaf. For function objects it
+    returns the operator() overload IDs, which makes the variable
+    a SymbolParent so the visitor generates detail pages for them.
+*/
+inline
+auto&
+allMembers(VariableSymbol const& T)
+{
+    return T.FunctionObjectOverloads;
```
alandefreitas: So the overloads are both members of the impl object and the variable? Is there a mechanism to ensure visitors don't go through them twice?

share/mrdocs/addons/generator/common/partials/symbol/qualified-name-title.hbs
```diff
@@ -21,9 +21,11 @@
     See: https://mrdocs.com/docs/mrdocs/develop/generators.html#dom_reference
 --}}
 {{~#if (and parent parent.parent)~}}
-{{~> symbol/qualified-name-title parent is-qualified-name-parent=true ~}}{{#if parent.name }}::{{/if}}
+{{~> symbol/qualified-name-title parent is-qualified-name-parent=true ~}}{{#if parent.name }}{{~#unless parent.isFunctionObject~}}::{{~/unless~}}{{/if}}
```
alandefreitas: There's something wrong with this pattern. Functions work without it. My intuition and the PR is forcing this template to do something is was not meant to do. qualified names in the standard don't include ()

share/mrdocs/addons/generator/common/partials/symbol/signatures.hbs
```diff
@@ -16,6 +16,16 @@
 {{#> markup/code-block~}}
 {{> symbol/signature }}
 {{/markup/code-block~}}
+{{#if isFunctionObject}}
```
alandefreitas: So we treat it as an overload? If so, why not reuse the code just below?

src/lib/AST/ASTVisitor.cpp
```diff
@@ -1944,6 +1945,20 @@ addMember(UsingSymbol& I, Symbol const& Member) const
     addMember(I.ShadowDeclarations, Member);
 }
 
+void
+ASTVisitor::
+addMember(VariableSymbol& I, Symbol const& Member) const
+{
+    // VariableSymbol satisfies SymbolParent only for traversal
```
alandefreitas: So this function will never run? Why does it exist then? And if it's impossible to make this not exist (doubt it, because it didn't), the convention for runtime code that shouldn't exist is a MRDOCS_ASSERT.

src/lib/AST/ExtractDocComment.cpp
```diff
@@ -1004,6 +1004,21 @@ class DocCommentVisitor
             MRDOCS_UNREACHABLE();
 
         default:
+            // Custom commands registered at runtime.
+            if (llvm::StringRef(cmd->Name) == "function_object")
```
alandefreitas: Nice

src/lib/AST/ExtractDocComment.cpp
```diff
@@ -1004,6 +1004,21 @@ class DocCommentVisitor
...
+            {
+                jd_.IsFunctionObject = true;
+                // Process the paragraph content as a regular paragraph
```
alandefreitas: We can have a single source of truth for all custom MrDocs flags, then. They should all behave this way unless there's a way to have a flag without a paragraph.

src/lib/Metadata/Finalizers/FunctionObjectFinalizer.cpp
```diff
@@ -0,0 +1,454 @@
...
+#include <algorithm>
+
+namespace mrdocs {
+
+namespace {
+
+// Check whether a parameter's type is a (possibly cv-qualified)
+// lvalue or rvalue reference to the given record type.
+bool
+isReferenceToRecord(
```
alandefreitas: I have an intuition that we already have a function for this somewhere in the codebase. Maybe these functions for types need to be better aggregated somewhere.

src/lib/Metadata/Finalizers/FunctionObjectFinalizer.cpp
```diff
@@ -0,0 +1,454 @@
...
+    }
+    return (**pointee).isNamed()
+        && (**pointee).namedSymbol() == recordId;
+}
+
+// A special member function is a default constructor, copy/move
+// constructor, copy/move assignment operator, or destructor
+// ([special]).
+bool
+isSpecialMemberFunction(
```
alandefreitas: Same thing. I just remember doing this somewhere already.

src/lib/Metadata/Finalizers/FunctionObjectFinalizer.cpp
```diff
@@ -0,0 +1,454 @@
...
+        }
+    }
+    return result;
+}
+
+void
+FunctionObjectFinalizer::
+markImplementationDefined(RecordSymbol& R)
+{
+    R.Extraction = ExtractionMode::ImplementationDefined;
```
alandefreitas, line 148: Oh... This answers one of my questions above. Good strategy :)

src/lib/Metadata/Finalizers/FunctionObjectFinalizer.cpp
```diff
@@ -0,0 +1,454 @@
...
+    for (SymbolID const& childId : allMembers(R.Interface))
```
alandefreitas, line 149: What about potentially deeper levels?

src/lib/Metadata/Symbol/Variable.cpp
alandefreitas, line 62: All these merge functions could use some reflection. That's one of the strongest use cases. I don't know if this should be incremental or in a single issue though.

src/lib/Support/Reflection/Reflection.hpp
```diff
@@ -190,7 +190,7 @@ BOOST_DESCRIBE_STRUCT(
     (Type, Template, Initializer, StorageClass, IsInline,
      IsConstexpr, IsConstinit, IsThreadLocal, Attributes, IsMaybeUnused,
      IsDeprecated, HasNoUniqueAddress, IsRecordField, IsMutable, IsVariant,
-     IsBitfield, BitfieldWidth)
+     IsBitfield, BitfieldWidth, IsFunctionObject, FunctionObjectOverloads)
```
alandefreitas: :)

alandefreitas: Maybe both of these could be replaced with `FunctionObjectImpl`. Unset values would mean not a function object (IsFunctionObject being false), and the FunctionObjectOverloads can always be retrieved from the FunctionObjectImpl. This also gives us extra information about the function object. 

IsFunctionObject and an empty IsFunctionObject are already redundant anyway right now.

alandefreitas (commented):
Isn't "functors" formal enough to be used here? It would be a much shorter command and a more concise way to describe it over and over again without needing a "_".

The fixtures should go to better directories, though. Maybe the directories need better documentation, but you can see they shouldn't go there by comparing with the siblings (and you'll see it's definitely not "core", which is used for really core functionality that's not optional). Symbol types and their features are tested in the symbols directory for each symbol. And since there's an option involved, every new option should be tested in the options directory. Then we have new features for variables that should be tested. Then we have a new field in variables, and that's not reflected there. Then we have a folder for javadoc commands that should be tested, and we need a test there because we introduced a new command.

The test cases in symbols/variable should also explore what happens when there is non-conflicting type information, but the variable is forced to be treated as a function object anyway. There are real-world patterns like this. Another pattern we see very often is the combination of template parameters and arguments: impl function with template, different variables with the same function but different params, and so on.

In the existing test, I'm not sure I understand the mechanism through which it goes to the list of functions in the interface. I'm not sure if cppalliance does that or shows it as a separate kind of entity. Because they do behave differently. All niebloids also have a special banner describing how that function works differently. Another thing to explore is how all of that works when one is visible, and the other is an implementation detail, and so on. Another thing to test and design is what happens when documentation goes on the other symbol.

In the output of the existing test, all functors look wrong. They look nothing like niebloids in cppreference. Many of them look arbitrary and not like valid C++ syntax at all. For instance, it tends to look like:

```
bool single_overload();
```

(not very different from an overload set with the main function name replaced from operator() to the variable name and with const removed).

rather than:

```
constexpr /* implementation-defined */ single_overload = {};

bool
operator()() const;
```

One more thing missing is the presentation of this feature on the landing page. The snippets on the landing page are meant to showcase features that only Mrdocs can support, and this is definitely one of them. 

Good work here overall, though. This is a big design decision.

include/mrdocs/Metadata/Symbol/Variable.hpp
gennaroprota: Yes. The `operator()` functions are removed from the record's interface and re-parented under the variable. So visitors won't encounter them twice.

include/mrdocs/Metadata/Symbol/Variable.hpp
gennaroprota: Oh, right.

share/mrdocs/addons/generator/common/partials/symbol/qualified-name-text.hbs
gennaroprota: Hmm, right. I always use "()" after function names, but MrDocs doesn't.

src/lib/Metadata/Finalizers/FunctionObjectFinalizer.cpp
gennaroprota: Aha. I think you mean the code in DocComment/Function.hpp. I see some issues with that:

- It doesn't handle default constructors with all-default parameters (only no parameters).

- It doesn't handle by-value copy assignment.

- It doesn't check for non-template.

I could fix these issues and re-use it. Should we extract those helpers (which are currently in an unnamed namespace) to a shared location? (BTW, unnamed namespaces in include files are bad! :-))

src/lib/Metadata/Finalizers/FunctionObjectFinalizer.cpp
gennaroprota: Yup, fixed :-).

src/lib/Metadata/Finalizers/FunctionObjectFinalizer.cpp
alandefreitas: Yes. I think so. Whatever you think is best, though. You understand the problem much better than I do. :)

src/lib/AST/ExtractDocComment.cpp
```diff
@@ -98,6 +98,29 @@ dumpCommentContent(
...
+struct CustomFlagCommand
+{
+    char const* name;
+    bool DocComment::* flag;
+};
+
+constexpr CustomFlagCommand customFlagCommands[] = {
+    {"functionobject", &DocComment::IsFunctionObject},
+    // For consistency, the name of the command must be "functionobject"
+    // and not "function_object; but that's not very readable, so we
```
alandefreitas: We don't need to describe what we're not doing. :)

include/mrdocs/Metadata/DocComment.hpp
```diff
@@ -171,6 +175,10 @@ struct MRDOCS_DECL DocComment {
 inline
 void merge(DocComment& I, DocComment&& other)
 {
+    // Merge the function-object flag before comparing so
+    // the flag alone doesn't cause spurious inequality.
+    I.IsFunctionObject |= other.IsFunctionObject;
```
alandefreitas, line 180: We need to replace these merge functions over time :)

src/lib/AST/ExtractDocComment.cpp
```diff
@@ -1004,6 +1027,23 @@ class DocCommentVisitor
...
 
         default:
+            // Custom flag commands registered at runtime:
+            // set the corresponding flag and preserve any
+            // paragraph text in the doc comment.
+            for (auto const& custom : customFlagCommands)
+            {
+                if (llvm::StringRef(cmd->Name) == custom.name)
+                {
+                    jd_.*(custom.flag) = true;
```
alandefreitas, line 1055: Cool design :)

src/lib/Metadata/Finalizers/FunctionObjectFinalizer.cpp
```diff
@@ -0,0 +1,414 @@
...
+// note admonition to a symbol's doc comment.
+void
+appendFunctionObjectDisclaimer(Optional<DocComment>& doc)
+{
+    if (!doc)
+    {
+        doc.emplace();
+    }
+    doc::ParagraphBlock para;
+    para.children.push_back(
```
alandefreitas: I think we should do this at the handlebars template level before we get complaints so the user can customize this to do whatever they'd like with it.

test-files/golden-tests/javadoc/function-object/function-object.cpp
```diff
@@ -0,0 +1,59 @@
...
+    */
+    int
+    operator()() const;
+
+    /** Extra public member that prevents auto-detection. */
+    char const*
+    name() const;
+};
+
+/** @functionobject
```
alandefreitas, line 29: Should we document that the command has no effect when auto-function-object is true?

test-files/golden-tests/javadoc/function-object/function-object.cpp
```diff
@@ -0,0 +1,59 @@
...
+struct undocumented_fn
+{
+    /** Call operator.
+        @return A value.
+    */
+    int
+    operator()() const;
+};
+
+/** Variable-level documentation. */
```
alandefreitas: Should we document that any comment on the variable or on the struct gets ignored? Or did I get the logic wrong?

test-files/golden-tests/symbols/variable/function-objects.cpp
```diff
@@ -0,0 +1,450 @@
...
+    A template constructor is not a special member function
+    per the C++ standard, so it counts as a non-operator()
+    public member that prevents auto-detection.
+*/
+struct template_ctor_fn
+{
+    template_ctor_fn() = default;
+
+    template<class U>
+    explicit template_ctor_fn(U);
```
alandefreitas, line 357: There are many variants of this pattern that use templates at this level. It's a reasonably common pattern to support templates in this case (for instance, all range algorithms). This will probably bite us on capy soon because it's really common. 

The other pattern, where the template goes in the struct, is a little rarer and harder to implement. It's also harder to implement because you have to actually instantiate the thing.

It's up to you, but I'd recommend implementing the easier one and opening a "Explore ..." issue for the hard one.

alandefreitas (commented):
Really nice tests. Everything about this PR is really cool 😎😄

src/lib/AST/ExtractDocComment.cpp
gennaroprota: I wanted to explain why we have two commands for the same thing.

include/mrdocs/Metadata/DocComment.hpp
gennaroprota: Yes. I worked on that a bit, and am waiting to merge this PR before opening a new one for that.

src/lib/Metadata/Finalizers/FunctionObjectFinalizer.cpp
gennaroprota: Great idea.

test-files/golden-tests/javadoc/function-object/function-object.cpp
gennaroprota: It still has effect if used on a type that fails auto-detection (for instance because it has extra public members).

test-files/golden-tests/javadoc/function-object/function-object.cpp
gennaroprota: Yes, I added some documentation. (We don't have documentation for the doc comment commands, though, so I added it to src/lib/ConfigOptions.json, where `auto-function-objects` is documented. That's not ideal.)

test-files/golden-tests/symbols/variable/function-objects.cpp
gennaroprota: Template `operator()` is already supported (the finalizer doesn't care whether `operator()` is templated or not). The test case here is specifically about a template constructor which isn't a special member function per the standard, so it blocks auto-detection. Are you suggesting we should treat template constructors (and template assignment operators) as "harmless" for auto-detection purposes?

alandefreitas: Oh... OK. Sorry. Yes. In the other case, we can force it with the `@functionobject` tag.

test-files/golden-tests/javadoc/function-object/function-object.cpp
alandefreitas: > where auto-function-objects is documented

Yes. That's exactly what I had in mind. Sorry I wasn't more explicit.

test-files/golden-tests/javadoc/function-object/function-object.cpp
alandefreitas: Oh... Duh... Of course. Sorry.

PR #1166, opened 2026-03-16
fix: Windows build

gennaroprota:


CMakeLists.txt
```diff
@@ -550,11 +551,15 @@ if (MRDOCS_BUILD_TESTS)
...
     # XML lint
     #-------------------------------------------------
-    if (MRDOCS_BUILD_STRICT_TESTS)
-        # Strict mode expects xml-lint to run; require LibXml2.
-        find_package(LibXml2 REQUIRED)
-    else()
-        find_package(LibXml2)
+    # Try config mode first (e.g. libxml2 installed by bootstrap.py),
+    # then fall back to module mode (e.g. system libxml2-dev on Linux).
+    find_package(LibXml2 CONFIG)
```
alandefreitas, line 556: I'm not sure, but I think there's a way to make config fall back to regular mode. If so, it would be more like:

    if (MRDOCS_BUILD_STRICT_TESTS)
        # Strict mode expects xml-lint to run; require LibXml2.
        find_package(LibXml2 CONFIG REQUIRED)
    else()
        find_package(LibXml2 CONFIG)
     endif()

If we use this pattern often, we can also create a REQUIRED_IF_STRICT_TESTS variable.

gennaroprota: We could do:

```
set(CMAKE_FIND_PACKAGE_PREFER_CONFIG TRUE)
if (MRDOCS_BUILD_STRICT_TESTS)
    find_package(LibXml2 REQUIRED)
else()
    find_package(LibXml2)
endif()
```

The variable is documented since [CMake 3.15](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_PACKAGE_PREFER_CONFIG.html). MrDocs requires CMake >= 3.13, though, so we'd need to check if 3.15 is acceptable; but, realistically, any current toolchain has it.

Downside: `CMAKE_FIND_PACKAGE_PREFER_CONFIG` is global, so it would affect all subsequent `find_package` calls in the same scope. We could use a save/restore.

alandefreitas: Exactly. Great idea.

I'm looking at this again, and I suspect #1168 will end up obsoleting this PR, though. Because the new bootstrap script had to cover all of that to actually work in CI. So it had to fix the root of the problem.

PR #1167, opened 2026-03-16
refactor: auto-merge symbol fields via Boost.Describe

gennaroprota:


alandefreitas (commented):
The problem we have (we have / not your PR has) is that we can only use reflection in source files, but reflection is a compile-time feature, and it has much less value if we can’t use it to define the reflection operations. 

For instance, this feature would typically be implemented as:

```cpp
template <canMerge T>
void
merge(T& I, T&& Other)
{
    // logic of mergeReflected(I, Other); directly
}
```

only **once**. 

But instead we have to define:

```cpp
void
merge(EnumSymbol& I, EnumSymbol&& Other)
{
    MRDOCS_ASSERT(canMerge(I, Other));
    mergeReflected(I, Other);
}
```

for each symbol that has this operation, because the information can't be available at compile time. We can't do it only once, and it's very repetitive (in fact, this is so repetitive that there could even be a DEFINE_MERGE macro to make this a little easier to maintain).

I can't say reflection is completely useless because we can still define `mergeReflected` in the source file. But it never lets us write code in a way that's easy to maintain with a single source of truth.

I think if the current direction is that we don't want public dependencies at all (which is not unreasonable), the best solution would be to bundle a single amalgamated `Support/Describe.hpp` file with the same contents of [boost/describe.hpp](https://github.com/boostorg/describe/blob/develop/include/boost/describe.hpp) we want iteratively include (with proper copyright notices, etc, the part we want removed, a short explanation of the changes: it could become MRDOCS_DESCRIBE in the headers until we have native reflection in C++, and so on). This is not different from what we've been doing for many other C++ features in this project.

Then we can create a *single* public custom `Support/Reflection.hpp` header with mrdocs operations for described objects, such as merge or whatever else (generators will probably need for_each operations, someone will probably need enum_to_string, and so on). With that, we can finally eliminate these workarounds and actually use reflection to remove code. So far, reflection is helping with correctness but not eliminating much code, and it's helping with maintenance costs.

Of course, the biggest concern then becomes the cost of maintaining this. I evaluated that, and Boost.Describe doesn't change significantly. It's been years. We're not using all of it, and the parts (on better, the connection) that depend on Boost.mp11 can be easily replaced with other simpler list types (like tuple or a short typedef) and just a couple of operations. Even if Boost.Describe changes significantly, which never happened, the probability that this is something that affects us is very low: we're just defining lists of pointers to member functions. Even the cost of maintaining it once every few years or creating it the first time is very low, because this is very repetitive code that AI would be really good at doing in about 20 minutes. Another consideration is that real C++ reflection will likely be available way before any of that can even happen, so we don't need to worry about what follows. The cost of maintaining N variants of each function we want "reflection" on is much higher than adjusting this once every few years, even if we have to.

Anyway, I think this is more of a comment on the state of reflection than on this PR. This PR is great as it is, considering our current constraints.

PR #1168, opened 2026-03-18
ci: dependency steps use bootstrap script

alandefreitas:
CI now uses the bootstrap script to install all dependencies instead of doing it inline. The script writes dependency paths and compiler flags to an env file that the build step picks up.

The documentation included 5 different methods for installing MrDocs from source, tvery developer had their own way of setting up MrDocs locally, and the CI workflow had its own completely separate procedure. When one changed, the other drifted. New contributors would follow the docs, hit errors, and waste hours figuring out what the CI did differently. The CI workflow itself was over 1300 lines of bash, YAML, and Handlebars templates that very few people understood.

Now there's one way to set up dependencies: `python bootstrap.py`. It works the same on a developer's laptop and in CI. If a dependency version changes, it changes in one place. If a build flag is wrong, it's fixed in one place. New contributors clone the repo, run the script, and get the same environment CI uses on their platform.

Bootstrap handles all dependency installation in CI via `--cache-dir` and `--env-file`, which were the two new features we implemented so that the bootstrap script interacts properly with CI cache. libc++ runtimes are always built automatically for clang sanitizer configurations. The LLVM recipe now includes the correct build options (Ninja, tablegen, tools off). CI uses devcontainer images where possible, removing setup steps for cmake, ninja, and system packages. `install.adoc` now documents bootstrap as the primary build path.

Also fixed recipe install paths, Windows preset suffix, build_dir template, and MSan cache key. Added `--cache-dir`, `--env-file`, `--cflags`/`--cxxflags`/`--ldflags`, prerequisite detection, and 316 bootstrap unit tests. Most "new" code in this PR is the bootstrap script being modularized and tested, which git will consider new lines of code. Most code removed came from the old ci.yml script.


.github/workflows/ci.yml
```diff
@@ -593,23 +355,23 @@ jobs:
...
-            ${{ runner.os == 'Windows' && '-D LibXml2_ROOT=../third-party/libxml2/install' || '' }}
+            -D CMAKE_EXE_LINKER_FLAGS="${{ env.BOOTSTRAP_LDFLAGS }}"
+            -D LLVM_ROOT="${{ env.LLVM_ROOT }}"
+            -D jerryscript_ROOT="${{ env.jerryscript_ROOT }}"
+            -D LUA_ROOT="${{ env.Lua_ROOT }}"
+            -D Lua_ROOT="${{ env.Lua_ROOT }}"
+            -D lua_ROOT="${{ env.Lua_ROOT }}"
+            -D boost_mp11_ROOT="${{ env.boost_mp11_ROOT }}"
+            -D boost_describe_ROOT="${{ env.boost_describe_ROOT }}"
+            ${{ env.LibXml2_ROOT && format('-D LibXml2_ROOT={0}', env.LibXml2_ROOT) || '' }}
```
gennaroprota: Should `{0}` be quoted to allow for paths with spaces?

util/bootstrap/src/core/process.py
```diff
@@ -26,6 +26,15 @@
 
 from .ui import TextUI, get_default_ui
 
+# Track env vars already exported during a dry-run session so each
+# unique key=value pair is printed only once (at its first occurrence).
+_dry_run_exported: set = set()
+
+
+def reset_dry_run_exports():
```
gennaroprota: It seems that this function is defined but never called.

util/bootstrap/src/presets/generator.py
```diff
@@ -364,23 +393,39 @@ def create_cmake_presets(
...
+    _TOOL_ROOT_VARS = {"GIT_ROOT"}
+    if valid_package_root_vars is not None:
+        valid_roots = set(valid_package_root_vars)
+        for preset in user_presets.get("configurePresets", []):
+            cache = preset.get("cacheVariables", {})
+            stale_keys = [
+                k for k in cache
+                if k.endswith("_ROOT")
+                and k not in valid_roots
+                and k not in _TOOL_ROOT_VARS
```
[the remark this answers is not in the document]
alandefreitas: Fixed. The cleanup now only removes stale _ROOT vars from the preset bootstrap created, not from other  user-managed presets in the same file.

util/bootstrap/src/installer.py
```diff
@@ -349,20 +439,54 @@ def install_dependencies(self):
...
+            if (
+                not self.options.dry_run
+                and self.options.cache_dir
+                and not self.options.force
+                and not self.options.clean
+                and self.is_non_empty_dir(recipe.install_dir)
+                and not is_recipe_up_to_date(recipe, resolved_ref)
+            ):
+                self.ui.ok(f"[{recipe.name}] found existing cache at {self.ui.maybe_shorten(recipe.install_dir)}. Writing stamp and skipping build.")
+                write_recipe_stamp(recipe, resolved_ref, self.options.dry_run, self.ui)
```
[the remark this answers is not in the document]
alandefreitas: Fixed. Removed the legacy cache path entirely — a non-empty directory without a valid stamp is no longer trusted. Additionally, stamps now include a content hash covering both the recipe definition and runtime parameters (sanitizer, compiler, flags), so any change to how a dependency is built invalidates the stamp and triggers a rebuild.

util/bootstrap/src/presets/generator.py
```diff
@@ -323,6 +349,9 @@ def create_cmake_presets(
         new_preset["cacheVariables"]["CMAKE_C_FLAGS"] = cc_flags.strip()
     if cxx_flags:
         new_preset["cacheVariables"]["CMAKE_CXX_FLAGS"] = cxx_flags.strip()
+    if ld_flags:
+        new_preset["cacheVariables"]["CMAKE_EXE_LINKER_FLAGS"] = ld_flags.strip()
+        new_preset["cacheVariables"]["CMAKE_SHARED_LINKER_FLAGS"] = ld_flags.strip()
```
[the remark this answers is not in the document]
alandefreitas: Fixed. User --ldflags are now merged with the injected clang toolchain flags instead of replacing them.

.github/workflows/ci.yml
alandefreitas: Done. :)

util/bootstrap/src/core/process.py
alandefreitas: Removed. :) Thank you.

PR #1173, opened 2026-04-08
fix: missing dependent array bounds in the output

gennaroprota:
[fix: missing dependent array bounds in the output](https://github.com/cppalliance/mrdocs/pull/1173/changes/0d528ea0fe7c68c415b0af9700003974178d0bbf)
This fixes the rendering of e.g.:

template<typename T, std::size_t N>
void
f(T(&)[N]);

which was previously rendered without the bound `N` in all the output
formats. For Adoc and HTML, the fix is just to fall back to the written
expression when the numeric value is absent. For XML, the fix is adding
ExprInfo handling in writeElement, *and* this adds several missing
elements in the XML output (e.g. initializers and requires clauses), not
just array bounds.


src/lib/Gen/xml/XMLWriter.cpp
```diff
@@ -177,9 +178,18 @@ XMLWriter::writeElement(std::string_view tag, T const& value)
         { if (!value) return; }
     else if constexpr (is_vector<Type>::value)
         { if (value.empty()) return; }
+    else if constexpr (std::is_base_of_v<ExprInfo, Type>)
+        { if (value.Written.empty()) return; }
 
     // Primitives inline, compounds wrapped.
-    if constexpr (std::is_same_v<Type, bool> ||
+    if constexpr (std::is_base_of_v<ExprInfo, Type>)
```
alandefreitas, line 185: Oh.... I thought this already used reflection.

alandefreitas (approved):
Looks good to me!

PR #1177, opened 2026-04-10
refactor: replace per-type operator==() and operator<=>() with a single generic template

gennaroprota:
Add a reflection-based operator==() and operator<=>() template pair that compare bases then members using the MRDOCS_DESCRIBE metadata.

This removes almost all per-type overloads.


alandefreitas (approved):
Looks amazing to me!

gennaroprota:
Thanks. Unfortunately, that MSVC bug doesn't allow removing more overloads.

alandefreitas:
> As a consequence, the operator<=>() overloads in namespace mrdocs (Symbol, Type, Name, Template types, etc.) cannot be replaced yet

I'm a little confused by the limitation. So the operator<=>() overloads in namespace mrdocs can't be replaced, but the ones in mrdocs::doc can? Or both can't? And if both can't, what can? I think from the description alone I couldn't mentally visualize all the instances of the problem described.

gennaroprota:
The overloads in `mrdocs` can't. The ones in `mrdocs::doc` can.

alandefreitas:
> The overloads in `mrdocs` can't. The ones in `mrdocs::doc` can.

Oh... I really don't understand this bug then. :)

I'll study it a little better.

alandefreitas:
> I'll study it a little better.

Got it! :) What if `merge` becomes a niebloid or something? This bug would make reflection unusable for so many things. Or we could use a windows machine to just iterate on this problem until we find a solution. I'm not sure. Do we have the error output in a gist or something?

In the extreme worst case we could so:

```
  #define MRDOCS_REFLECTED_COMPARISON(Type)                                  \
      inline std::strong_ordering operator<=>(Type const& a, Type const& b)  \
      { return describe_ops::operator<=>(a, b); }                            \
      inline bool operator==(Type const& a, Type const& b)                   \
      { return std::is_eq(a <=> b); }

  // Usage:
  namespace mrdocs {
      struct Symbol { /*...*/ };
      MRDOCS_REFLECTED_COMPARISON(Symbol)
  }
```

Or the same technique could be used to even define it inside the class. But I'm not sure this is a great idea. It's definitely worse than reflections to being available.

gennaroprota:
I tried making `merge()` a Niebloid but that didn't help. It seems that the mere presence of an `operator<=>()` constrained on `has_describe_members<T>::value` in `namespace mrdocs` poisons MSVC's evaluation of that same constraint for `merge()`.

Here are the errors from MSVC: <https://gist.github.com/gennaroprota/94404edb327916f3cb506549a7295bf0>.

gennaroprota:
Hey, I found a workaround!

alandefreitas:
> Hey, I found a workaround!

🥳

Issue #1162, opened 2026-03-02
`@par` not processed correctly when front matter is absent


akrzemi1:
Given the following comment
```c++
/**
    
    @par Effects 
    Initializes storage value with expression `MP::marked_value()`.
    
    @post `has_value()` is `false`.
  */
  AK_TOOLKIT_CONSTEXPR markable() AK_TOOLKIT_NOEXCEPT_AS(MP::marked_value())
    : _storage(MP::store_representation(MP::marked_value())) {}
```

MrDocs renders the following Asciidoc output

<img width="1107" height="627" alt="Image" src="https://github.com/user-attachments/assets/8984dfc5-2143-4849-9edd-7bf9ff9df280" />

Note that the text following `@par` is displayed on top.

This does no longer happen, when I put some text before `@par`:

```c++
/**
    
    Default constructor.
    
    @par Effects 
    Initializes storage value with expression `MP::marked_value()`.
    
    @post `has_value()` is `false`.
  */
  AK_TOOLKIT_CONSTEXPR markable() AK_TOOLKIT_NOEXCEPT_AS(MP::marked_value())
    : _storage(MP::store_representation(MP::marked_value())) {}
```

In that case, I get the intuitive output:

<img width="1057" height="680" alt="Image" src="https://github.com/user-attachments/assets/f0a31763-16b4-4a85-a511-704c175d8001" />

alandefreitas:
Fixed. Auto-brief extraction stops at a heading now, so a leading @par is processed correctly (8556f23c7, which cites this issue). Closing as done.

PR #1179, opened 2026-04-22
fix: stop auto-brief extraction at headings

gennaroprota:
When a symbol's documentation begins with a `@par Title` block (with no front matter text before it), parsing emits a `HeadingBlock` for the title followed by a `ParagraphBlock` for the body. The auto-brief pass walked past the heading looking for the first paragraph, promoted the `@par` body to the brief, and left the title rendered as a section with no content.

The fix is to stop iterating at the first HeadingBlock: content under a heading is the body of that section, not the symbol's introductory prose. This is tightly scoped---`HeadingBlock` is only emitted from `@par` today--- and preserves the existing behavior when a front matter is present.

The commit adds a regression test covering both shapes (no front matter, and explicit brief paragraph before the `@par` section).

Closes issue #1162.


gennaroprota:
Closed, since I opened it on 'develop' by mistake. Will reopen on a feature branch.

PR #1180, opened 2026-04-22
feat: render long noexcept specifications as noexcept(see-below)

### LINKED ISSUE #1103: Translate noexcept conditions
Mr. Docs needs to extract the noexcept condition and translate it into human readable terms

https://mrdocs.com/demos/develop/nlohmann-json/multi/adoc-asciidoc/nlohmann/swap-0e.html

The metadata should have a list of exception attributes, and the template should render them as a nice list in the exposition instead of in the declaration, which is ugly

gennaroprota:
MrDocs used to render the full `noexcept` operand inline, so a declaration like

    void swap(reference, reference) noexcept(
      std::is_nothrow_move_constructible<value_t>::value &&
      std::is_nothrow_move_assignable<value_t>::value &&
      std::is_nothrow_move_constructible<json_value>::value &&
      std::is_nothrow_move_assignable<json_value>::value);

buried the `noexcept` condition in a mostly-unreadable slop.

This replaces dependent operands with an italic "see-below" placeholder in the declaration, and moves the actual condition to a dedicated "noexcept Specification" section of the exposition:

    void swap(reference, reference) noexcept(see-below);

    === noexcept Specification
    noexcept when `...long condition...`.

The section is intentionally separate from the existing "Exceptions" section, which continues to cover `@throws` documentation.

Closes issue #1103.

[danger skip docs].

---------

## Summary

This renders long `noexcept` specifications as `noexcept(see-below)` and pulls the actual condition into a dedicated "noexcept Specification" section of the exposition, so the function signature stays readable when the condition is a dependent expression. Closes #1103.

Before this PR, a declaration like

```cpp
void swap(reference, reference) noexcept(
  std::is_nothrow_move_constructible<value_t>::value &&
  std::is_nothrow_move_assignable<value_t>::value &&
  std::is_nothrow_move_constructible<json_value>::value &&
  std::is_nothrow_move_assignable<json_value>::value);
```

buried the `noexcept` keyword in a wall of trait expressions. After this PR, the same declaration renders as

```cpp
void swap(reference, reference) noexcept(see-below);
```

and the condition appears under a separate "noexcept Specification" subsection alongside the declaration. The section is intentionally separate from the existing "Exceptions" section, which continues to cover `@throws` documentation.

The implementation lives mostly in the Handlebars template layer. The C++ side adds a small helper surface in include/mrdocs/Support/Handlebars.hpp (implemented in src/lib/Support/Handlebars.cpp); the existing function-signature, declarator-suffix, and per-format symbol partials under share/mrdocs/addons/ are updated to consume it.

## Changes

- **Source**: New Handlebars helper added in src/lib/Support/Handlebars.cpp and exposed via the public include/mrdocs/Support/Handlebars.hpp. Templates updated under share/mrdocs/addons/: common/partials/symbol/signature/function.hbs and common/partials/type/declarator-suffix.hbs decide when to print `(see-below)`; adoc/partials/symbol.adoc.hbs and html/partials/symbol.html.hbs render the new "noexcept Specification" section in their respective output formats.
- **Tests**: A new src/test/lib/Metadata/NoexceptInfo.cpp covers the underlying noexcept-info classification: which operands stay inline vs. get "see-below".
- **Golden tests**: New test-files/golden-tests/symbols/function/noexcept.{cpp,xml,html,adoc} fixture exercises the see-below path end-to-end across all three output formats.
- **Breaking changes**: None configuration is unchanged. The generated XML / HTML / AsciiDoc output for functions with long, dependent `noexcept` specifications changes; downstream tooling that scrapes function signatures and expects the full inlined condition will need to also read the new "noexcept Specification" section.

## Testing

- src/test/lib/Metadata/NoexceptInfo.cpp is the unit-level coverage for the classification rule (when an operand is short enough to inline vs. long or dependent enough to redirect to the see-below section).
- test-files/golden-tests/symbols/function/noexcept.{cpp,xml,html,adoc} is the end-to-end coverage across all three output formats.

## Documentation

No user-facing documentation page is added (`[danger skip docs]`). The change is a rendering improvement for generated output: it does not surface a new option, flag, or configuration knob a user would need to look up. The new "noexcept Specification" section is self-describing in the generated output.


alandefreitas:
What criteria did we end up with to consider the condition long?

gennaroprota:
There's no length check. The rule ended up being `noexcept.kind == dependent`.

Non-dependent specifications do carry text but, in practice, that's empty or literally `true`/`false`, so the inline form stays short. Dependent operands are (again, in practice) the only ones that may grow long, and are currently always rendered with "see-below", even when "short". Do you want to tweak that?

alandefreitas:
I'm not clear on what this means. OK. Questions:

- Aren't these just specifications dependent on template parameters? 
- I don't understand what this is a proxy for. Are you trying to infer the size from the specification kind? 
- Why is this a good proxy for that (size?)? It seems very easy to come up with a counterexample.
- And why would we look at a proxy when the thing the proxy is a proxy for (size?) is readily available?
- Also, I don't understand how this compares to whatever cppreference does.

For example, if `noexcept(std::default_constructible<T>)` were dependent and rendered in a way that's much harder to read, I don't see how that's a good thing.

I mean, to some extent, the strategy ended up being, in practice, "almost always use see-below". And it's mostly coherent with cppreference it uses see-below even for things like `noexcept(std::is_nothrow_move_constructible_v<E>)`. https://en.cppreference.com/cpp/utility/expected/expected

But I'm not sure this is a good thing for mrdocs because cppreference inherited this format from the old times when concepts weren't a thing (so the other rules were already there as a rule at the bottom of the page) and is edited manually (so they can always go back and forth and don't have to justify their thresholds precisely). 

On the other hand, for MrDocs, when we look at the nature of user complaints, it's usually only about the cases when it's not legible. This is typically when it's too large to render or when it uses lots of implementation details. In cases where it's something like `noexcept(std::is_nothrow_move_constructible_v<E>)`, users actually see mrdocs as an improvement over cppreference.

What's interesting is if you look at https://en.cppreference.com/cpp/utility/expected/expected, you'll see the explicit specifications are actually inline when they're short. Their rule is not the same, and I assume it's because they don't have to maintain their own historical precedent in this case.

gennaroprota:
OK, fixing an arbitrary length limit didn't look "right" to me, so I wanted something more logical and not arbitrary, like "dependent"/"non-dependent". But that's a weak proxy. If length is what users complain about (and we have that as `info.Operand.size()`) let's switch to a length threshold (say 40 characters): Inline when <= 40, "see-below" when longer. Does that work for you?

alandefreitas:
Yes. That doesn't seem arbitrary at all. But please confirm what this looks like when rendered in adoc and html.

include/mrdocs/Metadata/Specifiers/ExplicitInfo.hpp
```diff
@@ -27,7 +28,7 @@ struct ExplicitInfo
 {
     /** Whether an explicit-specifier was user-written.
     */
-    bool Implicit = true;
+    bool IsUserWritten = false;
```
alandefreitas: Is this change really valuable? I think the other name came from Krystian, but I think it makes sense. Explicit/Implicit carries more semantic weight than UserWritten to me. Because the user chose to write what he chose to write anyway.

include/mrdocs/Metadata/Specifiers/NoexceptInfo.hpp
```diff
@@ -28,7 +29,13 @@ struct NoexceptInfo
 {
     /** Whether a noexcept-specifier was user-written.
     */
-    bool Implicit = true;
+    bool IsUserWritten = false;
```
alandefreitas: Same

include/mrdocs/Metadata/Specifiers/NoexceptInfo.hpp
```diff
@@ -28,7 +29,13 @@ struct NoexceptInfo
...
+
+    /** Whether the operand is long enough to warrant rendering as
+        `noexcept(see-below)` plus a separate specification section,
+        instead of inline in the signature.
+    */
+    bool OperandIsLong = false;
```
alandefreitas: Oh... This should be at the presentation level, not at this level in the handlebars templates. If we open this Pandora’s box, we can never close it again.

include/mrdocs/Support/MergeReflectedType.hpp
```diff
@@ -152,16 +151,6 @@ mergeByType(T& dst, T&& src)
         }
         return true;
     }
-    // Types with .Implicit flag (NoexceptInfo, ExplicitInfo):
```
alandefreitas, line 160: What happened here? Was this wrong?

share/mrdocs/addons/generator/adoc/partials/markup/see-below.adoc.hbs
```diff
@@ -0,0 +1 @@
+pass:q[_see-below_]
```
alandefreitas: What? see-below is not a markup command. Have a look at what kind of content we have in this directory.

share/mrdocs/addons/generator/adoc/partials/symbol.adoc.hbs
```diff
@@ -168,6 +168,14 @@
 {{/each}}
 |===
 {{/if}}
+{{/if}}
+{{! noexcept specification — only when the operand is long enough to
+    be rendered as noexcept(see-below) in the signature. Short
+    operands are shown inline and need no separate section. }}
+{{#if symbol.noexcept.operandIsLong}}
```
alandefreitas: Just use the handlebars helper for size here. This does not belong in the corpus.

share/mrdocs/addons/generator/common/partials/symbol/signature/function.hbs
```diff
@@ -31,7 +32,9 @@
 {{~#if isConst}} const{{/if~}}
 {{#if isVolatile}} volatile{{/if~}}
 {{#if refQualifier}} {{refQualifier}}{{/if~}}
-{{#if noexcept}} {{noexcept}}{{/if~}}
+{{~#if noexcept.isUserWritten~}}
+    {{~#if noexcept.operandIsLong}} noexcept({{> markup/see-below }}){{else if noexcept.operand}} noexcept({{noexcept.operand}}){{else if (eq noexcept.kind "true")}} noexcept{{/if~}}
```
alandefreitas: Why not just write "see-below" literally or with the str helper so handlebars encodes it for us?

test-files/golden-tests/config/extract-friends/extract-friends.xml
```diff
@@ -140,6 +140,10 @@
     <name>rhs</name>
   </param>
   <func-class>normal</func-class>
+  <noexcept>
+    <is-user-written>1</is-user-written>
```
alandefreitas: It seems like we're not representing booleans correctly in XML.

test-files/golden-tests/config/extract-friends/extract-friends.xml
```diff
@@ -140,6 +140,10 @@
...
+    <kind>true</kind>
```
alandefreitas: This also seems wrong.

test-files/golden-tests/config/extract-friends/extract-friends.xml
```diff
@@ -140,6 +140,10 @@
     <name>rhs</name>
   </param>
   <func-class>normal</func-class>
+  <noexcept>
```
alandefreitas: It seems wrong that the XMLs are changing at all for this PR.

test-files/golden-tests/symbols/function/noexcept.adoc
```diff
@@ -0,0 +1,159 @@
...
+[source,cpp,subs="verbatim,replacements,macros,-callouts"]
+----
+template&lt;typename T&gt;
+void
+f6(
+    T&,
+    T&) noexcept(pass:q[_see-below_]);
+----
+
+=== noexcept Specification
```
alandefreitas: Is this intentional? 

Should it be "noexcept specification"? Or maybe "`noexcept` Specification"? What does cppreference use?

test-files/golden-tests/config/extract-friends/extract-friends.xml
gennaroprota: Yes. See PR #1200.

test-files/golden-tests/config/extract-friends/extract-friends.xml
gennaroprota: That's actually correct, although confusing, because `NoexceptKind` is defined as:

```
enum class NoexceptKind
{
    /** Potentially-throwing exception specification
    */
    False = 0,
    /** Non-throwing exception specification
    */
    True,
    /** Dependent exception specification
    */
    Dependent
};
```

test-files/golden-tests/symbols/function/noexcept.adoc
gennaroprota: Cppreference puts everything under "Exceptions" (see e.g. <https://en.cppreference.com/cpp/utility/expected/expected>). In any case, MrDocs uses title case for section headings, so, if we keep that separate section, the title must say "Specification", with an uppercase "S".

alandefreitas:
> No user-facing documentation page is added ([danger skip docs]). The change is a rendering improvement for generated output: it does not surface a new option, flag, or configuration knob a user would need to look up. The new "noexcept Specification" section is self-describing in the generated output.

About this point, I noticed MrDocs supports lots of very nice C++ idioms that make life very convenient, and most of them are not documented. I'm noticing this as I'm working on this presentation about MrDocs. Maybe that's how this could go into the documentation. We could have a page or section for C++ idioms, and over time we can try to categorize them, and then we just have a reference of all the nice things MrDocs can do, and that's going to be one of them.

alandefreitas:
Hi, the repo was force-pushed 25 days ago. Is this ready for another look, or are you still working through it?

Issue #1146, opened 2026-01-19
Inline HTML causes warnings and eventually an error


jll63:
Putting a HTML table in javadoc:

```html
    //! <table>
    //!   <tr>
    //!     <th>Method Parameter Type</th>
    //!     <th>Overrider Parameter Type</th>
    //!     <th>Returns</th>
    //!   </tr>
    //!   <tr>
    //!     <td>T&</td>
    //!     <td>U&</td>
    //!     <td>U&</td>
    //!   </tr>
    //! </table>
    //!
```

Causes:

```
Building docs in custom dir...
[09:58:33.951] FATAL (antora): Command failed with exit code 1: /home/jll/.cache/antora/reference-collector/mrdocs/linux/develop/bin/mrdocs --config=/home/jll/dev/boost/libs/openmethod/doc/mrdocs.yml --output=/home/jll/.cache/antora/reference-collector/reference/openmethod/main --generator=adoc --multipage=true --tagfile=reference.tag.xml
warning: HTML <table> tag not followed by pure text (src/lib/AST/ExtractDocComment.cpp:511) at /home/jll/dev/boost/libs/openmethod/include/boost/openmethod/core.hpp (2862)

An issue occurred during execution.
If you believe this is a bug, please report it at https://github.com/cppalliance/mrdocs/issues
with the following details:
    MrDocs Version: 0.8.0 (Build: c136a465d068b4ceac8e4b033fb58287a4e1151e)
    Reported From: `src/lib/AST/ExtractDocComment.cpp` at line 631

warning: HTML <tr> tag not followed by pure text (src/lib/AST/ExtractDocComment.cpp:511) at /home/jll/dev/boost/libs/openmethod/include/boost/openmethod/core.hpp (2863)

An issue occurred during execution.
If you believe this is a bug, please report it at https://github.com/cppalliance/mrdocs/issues
with the following details:
    MrDocs Version: 0.8.0 (Build: c136a465d068b4ceac8e4b033fb58287a4e1151e)
    Reported From: `src/lib/AST/ExtractDocComment.cpp` at line 631

warning: HTML <tr> tag not followed by pure text (src/lib/AST/ExtractDocComment.cpp:511) at /home/jll/dev/boost/libs/openmethod/include/boost/openmethod/core.hpp (2868)

An issue occurred during execution.
If you believe this is a bug, please report it at https://github.com/cppalliance/mrdocs/issues
with the following details:
[...]

PR #1183, opened 2026-04-23
feat: support HTML table tags in doc comments

gennaroprota:
Previously, an HTML table in a doc comment triggered:

  warning: HTML `<table>` tag not followed by pure text

and then a fatal error: the existing HTML start-tag handler could only gather pure text between a tag and its matching end tag. The doc-comment extractor now recognizes `<table>`/`<tr>`/`<th>`/`<td>` and populates the existing `TableBlock`/`TableRow`/`TableCell` metadata.

The HTML and Adoc generators each gain a doc/block/table partial. The Adoc output uses the `[%header]` table attribute when the first row is a header; the HTML output emits plain `<table>`/`<tr>`/`<th>`/`<td>`. The XML generator already serialized the metadata generically.

Not (yet?) supported:

- HTML attributes on `<table>`/`<tr>`/`<th>`/`<td>` (`align`, `colspan`, `rowspan`, `class`, `style`, ...). Column alignment is therefore never populated (`TableBlock::Alignments` is always empty) and cells cannot span rows or columns.
- `<caption>`, `<col>`, and `<colgroup>`.
- Block-level content inside cells (paragraphs, lists, code blocks). Cells currently hold only inline content.

Closes issue #1146.


alandefreitas (approved):
LGTM

PR #1186, opened 2026-04-29
fix: render inline markup correctly in Adoc and HTML

### LINKED ISSUE #1185: HTML output lacks `<em>` , `<mark>`, `<sub>`, `<sup>`, `<del>`
Several kinds of inline markup in doc comments don't produce corresponding HTML tags.

For instance, test-files/golden-tests/javadoc/inline/styled.cpp contains `_italic_`, which produces a `doc::EmphInline` in the metadata, but the HTML golden renders it as the bare text "italic", with no `<em>` wrapper. The XML golden has `<emph>...</emph>` correctly; the AsciiDoc golden renders `_italic_` correctly.

### Cause

Each affected share/mrdocs/addons/generator/common/partials/doc/inline/<kind>.hbs invokes a `markup/i` partial:

```
{{#> markup/i }}{{> doc/inline-container }}{{/markup/i}}
```

For AsciiDoc, share/mrdocs/addons/generator/adoc/partials/markup/i.adoc.hbs exists and renders `_..._`, so all of these end up rendered identically as italic in Adoc, which happens to be visually correct for emphasis but is semantically wrong for the others. For HTML, share/mrdocs/addons/generator/html/partials/markup/i.html.hbs does not exist; the partial-block invocation produces nothing, and only the inner content survives, hence the missing wrappers.

The shared `markup/i` indirection conflates several semantically different inline kinds. Each kind should map to its own HTML element.

gennaroprota:
**Note**: Rebased onto current develop, where highlight and strikethrough already rendered correctly, so this PR now covers the remaining three kinds. It also documents subscript and superscript in the inline-commands reference with runnable examples.

---

Five kinds of inline markup in doc comments (emphasis, highlight, subscript, superscript, and strikethrough) were not rendered correctly:

- HTML output silently dropped the wrapping element, so emphasized text and the others appeared as bare text.
- AsciiDoc output rendered all five as italics, which is correct for emphasis but semantically wrong for the others (e.g., a subscript appeared as italic text, not lowered).

Both rendering paths now produce semantically appropriate markup: `<em>`, `<mark>`, `<sub>`, `<sup>`, `<del>` in HTML, and the matching constructs in AsciiDoc (with strikethrough using the role-based `[.line-through]#...#`, since AsciiDoc has no native syntax for it).

Doc comments may now also use the HTML-tag form for these kinds (`<strong>`, `<mark>`, `<del>`, `<s>`, `<sub>`, `<sup>`) alongside the previously-supported `<em>`. This is what most users reach for when they need a tight notation (e.g., `H<sub>2</sub>O`); the Markdown-marker forms (`~text~`, `^text^`, `==text==`, `~~text~~`) require whitespace flanks, which makes compact chemistry/math notation awkward.

Three other inline kinds (footnote-reference, image, and math) suffer from the same rendering issue but each requires its own non-trivial design decision and is left for separate fixes.

Closes issue #1185.


alandefreitas:
This is really nice. I think it's going to need a rebase because a few kinds of inline comments have already been fixed since then. We'll also need the examples for the new reference pages that didn't exist 40 days ago.

alandefreitas:
Minor: the parser now accepts `<s>` alongside `<del>` for strikethrough, and the golden exercises `<s>`, but the Strikethrough entry in `commands-registry.json` only lists `<del>text</del>` under syntaxes. Worth adding `<s>text</s>` there so the reference page matches what's accepted.

PR #1192, opened 2026-04-30
feat: support macros

### LINKED ISSUE #1127: Support macros
Macros can be extracted before the AST step via the preprocessor:

This is a new issue meant to obsolete https://github.com/cppalliance/mrdocs/issues/627 because of  https://github.com/cppalliance/mrdocs/issues/627#issuecomment-2187150299

gennaroprota:
This adds support for preprocessing macros: `#define` directives are captured via a `clang::PPCallbacks` subclass and exposed as `MacroSymbol` instances at the corpus root (since macros aren't in any C++ scope). Include guard macros are recognized and skipped.

Filters apply to macros, too:

- `extract-all`: when off, undocumented macros are dropped.
- `exclude-symbols`, `include-symbols`, `implementation-defined`, `see-below`: matched against the macro name.
- File-pattern filters: matched against the directive's source location.

Multipage mode produces a `macros.{ext}` page at the output root listing every macro with a link to its per-symbol page, and the global-namespace page gains a "See also: Macros" navigation hint (multipage-only) so the new page is discoverable. Macro `@param` blocks are validated.

## Changes

- **Source**: Macro capture in a new `clang::PPCallbacks` subclass, exposed as a corpus-root `MacroSymbol`. The Handlebars generators render the macro synopsis, the single-page section heading, the multipage `macros` page, and the "See also: Macros" link from templates.
- **Golden tests**: Macro fixtures covering extraction (happy path, edge cases, extraction modes, multi-file inputs, embedded macros, include guards), the filter paths, rendering, and multipage output.
- **Build**: `mrdocs.rnc` gains a `<macro>` element, a sibling of `<namespace>`, with `<source>` carrying the verbatim definition.
- **Breaking changes**: None for projects without macros. The output schema gains a new top-level `<macro>` element, so downstream tooling that enumerates top-level elements should add a clause for it.

## Testing

Covered by the golden tests above. No CI workflow changes are needed.

## Documentation

No user-facing documentation in this PR. A reference page describing what is captured, how the filters apply, and the new `macros` page is worth a follow-up so the feature is discoverable from the docs rather than the schema.

Closes #1127.


alandefreitas:
Thanks for exploring this. This is a complex issue.

> This adds support for preprocessing macros: #define directives are captured via a clang::PPCallbacks

Interesting. I thought it would be easier to have a preprocessor step just for macros. 

>  so we look up a preceding doc-comment by source location

This seems extremely problematic because the counterexamples where it doesn't work would be kind of obvious. Clang doesn't even attach comments to anything if there's not close to a symbol, which is probably the case when there's a macro.

My intuition is we can't really implement this feature unless clang wants to attach comments to macros or we implement our own parser (which we also intend to).

> Filters apply to macros, too

How do these compare to doxygen? Shouldn't we have some special options for macros? I'm not sure. I assume macros are somewhat special in terms of what's typically hidden or shown, but I haven't thought too much about it.

I think @mizvekov would probably be more qualified to review this issue and PR.

gennaroprota:
>> This adds support for preprocessing macros: #define directives are captured via a clang::PPCallbacks

> Interesting. I thought it would be easier to have a preprocessor step just for macros.

Hmm... I'm listening for `MacroDefined` events on the preprocessor instance that Clang is already running for the AST pass, which seems natural.

>>  so we look up a preceding doc-comment by source location

> This seems extremely problematic because the counterexamples where it doesn't work would be kind of obvious. Clang doesn't even attach comments to anything if there's not close to a symbol, which is probably the case when there's a macro.

> My intuition is we can't really implement this feature unless clang wants to attach comments to macros or we implement our own parser (which we also intend to).

Why? The current condition for accepting a preceding comment is that it is separated from the macro definition by blank lines only (and it is a doc-comment). We could tighten that to "no lines at all between the comment and the macro definition"; or maybe we can accept at most one line that is blank.

>> Filters apply to macros, too

> How do these compare to doxygen? Shouldn't we have some special options for macros? I'm not sure. I assume macros are somewhat special in terms of what's typically hidden or shown, but I haven't thought too much about it.

I surveyed Doxygen's options. They are:

- `ENABLE_PREPROCESSING`/`MACRO_EXPANSION`: master switches for Doxygen's preprocessor. MrDocs always runs a preprocessing pass.

- `PREDEFINED`: pre-defines macros. MrDocs has the `defines` config option.

- `EXPAND_ONLY_PREDEF`/`EXPAND_AS_DEFINED`: restrict macro expansion to predefined macros plus those of an explicit list. MrDocs has no equivalent.

- `SKIP_FUNCTION_MACROS`: heuristic that drops some function-like macros; it exists to keep some macros from confusing the Doxygen parser.

Would you like to add any of these to MrDocs?

alandefreitas:
> Why? The current condition for accepting a preceding comment is that it is separated from the macro definition by blank lines only (and it is a doc-comment). 

Oh... OK. That's good. I had only made the comment because your previous comment only said by "source location", which implied there could be content between the macro and the source location. But what you mentioned sounds great.

> Would you like to add any of these to MrDocs?

No. Great survey. Thanks. I think my original comment was more specific about filtering macros. But that’s great. If there are no specific options to filter macros by name, let's just go with what we have. Sounds great.

One more question: if clang doesn't attach comments to macros like it does to the Decls, how do we find these comments? Do we look for comments attached to something else that should be attached to macros and then move them to the macros? I don't really understand it yet.

gennaroprota:
Good question. Clang's `ASTContext` stores every comment it lexes in `ASTContext::Comments` (which, despite the type name `RawCommentList`, is really a per-file table). In the case of declarations, it also builds a second-layer mapping (`Decl` -> `RawComment`) by walking from each `Decl`'s location backward with some proximity rules; in the case of macros it doesn't, but the underlying per-file map still has every comment in the TU, including the ones above `#define`s. We just do the location-based lookup ourselves. This logic is in `findPrecedingMacroComment`, in ASTVisitor.cpp.

mizvekov:
FYI the need for this approach will soon be made obsolete by https://github.com/llvm/llvm-project/pull/198452

gennaroprota:
@mizvekov: Thanks for the heads-up! @alandefreitas: I think we can use the current approach for the moment being, and update it when the upstream patch lands and we switch to a new Clang version.

alandefreitas:
> I think we can use the current approach for the moment being, and update it when the upstream patch lands and we switch to a new Clang version.

I think the current approach could be reasonable. But I also think if https://github.com/llvm/llvm-project/pull/198452 gets merged in two weeks or so, using the next version will be easier and clearer. We won’t notice the difference; we have about 10 PRs in line.

gennaroprota:
> I think the current approach could be reasonable. But I also think if [llvm/llvm-project#198452](https://github.com/llvm/llvm-project/pull/198452) gets merged in two weeks or so, using the next version will be easier and clearer. We won’t notice the difference; we have about 10 PRs in line.

Yes, *if* we also plan to switch to a new Clang version as soon as it is released.

alandefreitas:
> new Clang version

It's more like a commit than a version, but yes. It'd be the same major. It's about time to upgrade anyway.

gennaroprota:
@alandefreitas: The patch has been merged upstream.

alandefreitas:
> @alandefreitas: The patch has been merged upstream.

🚀🔥

alandefreitas:
How's the situation with LLVM going so far?

PR #1200, opened 2026-05-13
fix: render true booleans as "true" in XML output

gennaroprota:
The XML writer emitted true boolean members as `<x>1</x>` (and omitted the element when false). It now emits the xsd:boolean primary lexical form, `<x>true</x>`. The omit-when-false convention is preserved, so the golden XMLs change only in the body of already-present boolean elements. The RELAX NG schema is updated accordingly.

-----

## Summary

This fixes a long-standing readability quirk in the XML output. The XML writer emitted true boolean members as `<x>1</x>` (and omitted the element when false). Reading such output meant the consumer had to know in advance that the member was actually a boolean, as `1` looks indistinguishable from an integer.

The writer now emits the `xsd:boolean` primary lexical form: `<x>true</x>`. The omit-when-false convention is preserved, so the golden XML fixtures change only in the body of already-present boolean elements. The RELAX NG schema is updated to match.

## Changes

- **Source**: One-line fix in src/lib/Gen/xml/XMLWriter.cpp; the boolean-serialization branch now writes `"true"` instead of `"1"`.
- **Golden tests**: 207 XML fixtures across test-files/golden-tests/ updated to reflect the new `<x>true</x>` form.
- **Build**: mrdocs.rnc updated so the schema accepts the new lexical form on boolean elements.
- **Breaking changes**: Downstream tooling that parses mrdocs XML output and matches boolean fields exactly against `"1"` will need to update to handle `"true"`. The omit-when-false convention is unchanged, so the handling of absent elements is unaffected.

## Testing

- The golden XML fixtures are themselves the regression coverage: any future regression that re-emits `"1"` for booleans (or omits an element when true) fails the golden tests.

## Documentation

The schema (mrdocs.rnc) is the canonical reference for the XML output format and is updated. No prose documentation describes the previous `"1"` encoding, so nothing else needs updating.


alandefreitas:
> The XML writer emitted true boolean members as <x>1</x> (and omitted the element when false). It now emits the xsd:boolean primary lexical form, <x>true</x>. The omit-when-false convention is preserved

I'm no XML expert, but that seems like a weird convention to make. If we omit when false, this is technically a flag. And flags don't have to say "true". I thought booleans were typically represented as attributes in XML.

gennaroprota:
> > The XML writer emitted true boolean members as 1 (and omitted the element when false). It now emits the xsd:boolean primary lexical form, true. The omit-when-false convention is preserved
> 
> I'm no XML expert, but that seems like a weird convention to make. If we omit when false, this is technically a flag. And flags don't have to say "true".

I was going to point out the same. `true` is currently redundant.

> I thought booleans were typically represented as attributes in XML.

Yes. For us, the form would be something like:

```
<function name="foo" is-const="true" is-variadic="false" ...>
  <param>...</param>
  ...
</function>
```

The MrDocs schema currently has almost no data attributes, so booleans-as-elements is at least consistent. I'd suggest switching to "empty element (e.g. `<is-variadic/>`) when true, nothing when false", as that avoids the enormous diff that would result from moving all booleans to attributes.

alandefreitas:
> The MrDocs schema currently has almost no data attributes

Yes. We used to have many.

But I think it's going to be more like `is-const` than `is-const="true"` because =true is also redundant here.

For `name="foo"`, then I'm not sure. We need research on typical usage and trade-offs.

alandefreitas:
Hi, checking in. We agreed on "empty element when true (`<is-variadic/>`), omitted when false" on 2026-05-22, but the branch still has the original `<x>true</x>` from 2026-05-13. Is this PR blocked by something else?

gennaroprota:
Hi. This, too, is ready for the second round of review.

alandefreitas:
Roger

alandefreitas:
Sorry, I didn't express myself well in my last message because I was passing by all PR checking on them and didn't give this the proper attention. I quoted your empty-element proposal as if we'd agreed on it, and you correctly implemented it. But my previous understanding is we hadn't agreed on that. Here's how I read the four steps of the thread:

- First message: I said `1` looked like a weird convention and that booleans are usually attributes in XML. What I was after was **research into the best convention, then using that, regardless of what we do today**.
- Your reply proposed the empty element, but I don't agree with the two reasons you mentioned. Consistency with itself doesn't help us much. A small diff isn't something we ever optimized for, least of all in a schema change. And this PR already touches every golden file anyway.
- Third message: I reinforced, although maybe not as clear this time, that I want the research on the best solution, whatever it is, regardless of existing practice or how much work it takes (this is work we can throw AI at). I also said I lean toward a bare attribute, `is-const` with no value, a tag. That was me guessing the likely outcome, not deciding it.
- Fourth message was a nudge where I didn't give much attention to context and ended up quoting your own proposal instead of restating the request from the previous review.

So the request from the previous review was: **do the research, lay out the trade-offs, then we pick**. My guess is we land on simple tags for booleans, but I'm not sure, and that's a side note. We need to research and see the trade-offs. When analyzing the tradeoffs, I think it's also important to notice the internal motivation for XML is for tests. Users are probably going to use JSON for their own tasks that need structured data.

PR #1203, opened 2026-05-15
fix: inherit base members from dependent template specializations

### LINKED ISSUE #1176: Inherited members are not shown in derived classes
I have a base class that defined `operator[]`, this operator doesn't appear in any derived class (unless it is explicitly redefined).

For example `subarray` defined `operator[]`: https://correaa.gitlab.io/boost-multi/multi/reference/boost/multi/subarray.html

`array_ref` derives from `subarray`, yet `operator[]` doesn't appear in its documentation.

I tried using ""copy-all", "reference", and "copy-dependency".

I am using this config file:

```
# MrDocs configuration for Boost.Multi
# https://github.com/cppalliance/mrdocs

source-root: include/

compilation-database: build-mrdocs/compile_commands.json

input:
  - include/boost/multi/array_ref.hpp
  - include/boost/multi/array.hpp

output: doc/modules/reference/pages

generator: adoc

multipage: true

extract-private: false

inherit-base-members: "copy-all"

base-url: https://gitlab.com/correaa/boost-multi/-/blob/develop/include/

include-symbols:
  - "boost::multi"
exclude-symbols:
  - "boost::multi::detail"

sort-members-by: location
sort-namespace-members-by: location

warn-no-paramdoc: false

verbose: true
```

full repo: https://gitlab.com/correaa/boost-multi

gennaroprota:
Fixes issue #1176.

Members defined in a class template base were silently dropped from derived class templates that inherited via parameter passthrough (e.g., `template<T> class derived : public base<T>`), even with `inherit-base-members: copy-all`. Concrete bases (`derived : public base<int>`) worked, dependent ones didn't, and the difference was invisible: no warning, no diagnostic (the base just got skipped).

The root cause was in `BaseMembersFinalizer.cpp`: when `extract-implicit-specializations` is on (the default), the code unconditionally replaced `Name::id` (primary template's ID) with `SpecializationName::specializationID`. The latter is set by `ASTVisitor` only for `ClassTemplateSpecializationDecl`s, i.e., concrete instantiations. For a dependent base, Clang's `TemplateSpecializationType::getAsCXXRecordDecl()` returns the primary template, the `dyn_cast` to CTSD fails, and `specializationID` stays `SymbolID::invalid`. The subsequent `MRDOCS_CHECK_OR_CONTINUE(baseID)` then quietly bailed out. The fix is to fall back to the primary's ID when `specializationID` is invalid, so dependent template-template inheritance picks up the primary's members with the primary's template parameter intact. Concrete cases still inherit from the implicit specialization as before.

## Changes

- **Source**: `BaseMembersFinalizer` now only switches from the primary template's ID to `SpecializationName::specializationID` when the latter is valid; dependent bases keep the primary's ID and inherit its members.
- **Tests**: new `template-base.cpp` fixture under `test-files/golden-tests/config/inherit-base-members/`, covering both the template-template passthrough case (`derived : public base<T>`) and a non-template-derived-from-concrete-specialization control (`concrete_derived : public base<int>`).
- **Golden tests**: generated `template-base.{xml,adoc,html}` to reflect the cure: `derived` now lists the inherited `method` and `operator[]` alongside the existing entries on `concrete_derived`. No other fixtures changed.

## Testing

The new `template-base` fixture is picked up automatically by the existing `mrdocs-golden-tests-{xml,adoc,html}` CTest targets, which already run on every CI build via the `Build & Test` matrix. The control case (`concrete_derived`) verifies that the previously-working concrete-specialization path stays unchanged, and the new case (`derived`) verifies the fix going forward. No CI workflow changes needed.

## Documentation

Not needed: the public behavior of `inherit-base-members` already documents what users expected; this PR makes it match its description.


alandefreitas (commented):
LGTM. My comment is that this title makes sense for the issue, but not for the commit. After the PR, we know exactly the case we're fixing. Issue numbers in the title are uncommon, too.

gennaroprota:
I'll remove the issue number from the title. As to the commit message subject mentioning the issue and not the fix, that was intentional: we report the subject in our release notes, and people may be scanning them for known problems. Would you like to mention the fix instead?

alandefreitas:
Oh, I see. I didn’t know it was intentional. Well, I think this would break the pattern used here for years. Changing the pattern is significant enough to warrant internal discussion. The impact I see in the kind of changelog we get:

- The changelog extension has an expectation of where the issues we reference go. This issue wouldn’t appear on the list of issues in the changelog or the corresponding issues for a change because the extension will look for the issue number in the body, as in conventional commit messages, and for the GitHub issue-tagging pattern. 
- The changelog would also become more redundant. Right now, the person knows what has been fixed from the change log (if their case is covered by the fix), and the issue provides a reference to the issue that led us to discover the problem (someone else’s case). But after this change to the changelog, the person scanning it doesn’t know what has been fixed, only the case that led us to find out about the necessity for the fix. The new format would remove the initial information about how it affects their case (for instance, cases involving specializations) and would include information about someone else’s case twice (when the person didn’t yet know the full category of fixtures affected). 

We could change the system. But I don't think the internal discussion about weather we should change this is worth blocking this PR. I think just changing the title until we have the discussion makes more sense even if we decide to change it.

correaa:
I am curious what is preventing this PR to be merged?

alandefreitas:
> I am curious what is preventing this PR from being merged?

It's still under review. And it moves somewhat slowly because we have 10 others under review.

correaa:
ok, thank you. I was wondering if it was something technical. Great work!

alandefreitas:
In fact, I think the only problem we had in the previous review round has already been fixed. I wasn't aware of that. All it needs is to be rebased then.

correaa:
I am using the mrdocs' binaries through develop's tar gz. I didn't try to compile mrdocs yet.

Issue #1210, opened 2026-05-19
Design the extension entry-point ladder beyond `transform_corpus`


gennaroprota:
This issue is the home for the design discussion that came out of PR #1196's review.

## Background

PR #1196 introduces the corpus-mutation extension system with a single entry point: a script defines a `transform_corpus(corpus)` global, and the host calls it. This is the simplest possible shape (let's call it **Option 1**) and it's what shipped in that PR.

During review, Alan raised the question of whether this shape is appropriate, particularly as we add:

- More lifecycle hooks (e.g. `before_extract`, `after_render`).
- Extensions that register multiple capabilities from one file (a corpus transform and a Handlebars helper).
- Extensions that bundle non-code files (templates, assets, configuration).
- Extension enabling/disabling.

The PR ships Option 1. This issue captures the trade-off analysis and lays out the ladder of richer options we can add as needs arise.

### Option 1: Reserved function names

Script defines a function with a known name (`transform_corpus`); the host calls it.

Examples: pytest, Sphinx.

Trade-off: minimal syntax. The host owns the names, so one file can expose many *different* known hooks, but a script can't add several capabilities of the same kind under names it picks itself.

My evaluation: right starting point for Mr. Docs, and it scales further than it looks. pytest recognizes dozens of `pytest_*` hooks, and a single `conftest.py` (or one Sphinx `setup()`) routinely defines many of them together. So "many capabilities in one file" is not a reason to leave this rung; the only thing it genuinely can't do is let a script name its own capabilities.

### Option 2: Top-level registration calls

Script calls `host.register_*(fn)` in top-level code; the host stores the registration and invokes the callback at the right time.

Exampled: Darktable, LLVM/Clang plugins.

Trade-off: the script passes a name, so it can add capabilities the host never pre-named, like several generators with author-chosen names from one file. The price: the host must run the script just to learn those names, then keep each registered callback alive until it calls it. (GDB's pretty-printers work this way: `register_pretty_printer` stores the callable, and GDB invokes it later, once per value.)

My evaluation: not needed yet. Reserved names already give us "many capabilities per file", so that's not the reason to climb. The real trigger is wanting scripts to *name their own* capabilities, most concretely one extension that adds several named generators. Until that's a concrete need, paying the run-at-discovery and keep-alive cost is premature.

### Option 3: Reserved `register` function + event emitter

Scripts export one reserved name (`register`); inside, it subscribes to host events.

Example: Antora.

[...]

PR #1218, opened 2026-06-03
feat: support script-driven generators

### LINKED ISSUE #1216: Generator extensions: user-defined output generators
Splitting this out per the #1128 / #1196 discussion. Transformation extensions (scripts that mutate the corpus) shipped in #1196. Generator extensions are the separate capability: letting users add or customize an output generator without touching MrDocs core.

The concrete first approach is tracked in #1082 (format-agnostic Handlebars generator: a base generator parameterized by format, templates under `share/mrdocs/addons/generator/<format>/`, an escape map, supplemental addon directories; e.g. a Markdown generator); it shipped as data-driven generators in #1197. This issue is the umbrella for the capability, which now has two flavors: data-driven (template) generators (#1197) and script-driven generators (#1218).

Open questions (resolved by #1197 and #1218):
- Template-only generators (the #1082 flavor) vs script-driven generators (interpreter hooks).
- How a custom format registers and how output file extensions map to template directories.
- Escaping rules per format.
- Interaction with addon layering / supplemental directories.

Distinct from #1210 (richer transformation entry points), #1213, and #1215, which are all about transformation extensions, not generators.

gennaroprota:
This adds the second flavor of addon-defined output generator, complementing the data-driven Handlebars generators from #1197: a generator backed entirely by a user Lua or JavaScript script, with no C++ and no compilation. The script owns the whole emit loop through a `generate(corpus, output)` entry point, so it can produce output shapes the per-page generators cannot, such as a single artifact aggregated across every symbol (a search index, for example).

The design is recorded in the #1216 design comment.

## Changes

- **Source**: A new src/lib/Gen/script/ module: `ScriptGenerator` (a `Generator` whose `build()` owns the emit loop and invokes the script's `generate(corpus, output)`), `OutputSink` (the file-writing API exposed to the script as `output.write`, resolved under the output directory and forbidden from escaping it), and the Lua and JavaScript runners. The manifest parser and the addon-directory walk are factored into a shared src/lib/Gen/GeneratorManifest.{hpp,cpp}, so the data-driven and script-driven discovery passes read the same mrdocs-generator.yml; src/lib/Gen/hbs/DataDrivenGenerators.{cpp,hpp} now consume it and skip directories that declare a `script:` entry. `script::discoverScriptGenerators` is wired into src/tool/GenerateAction.cpp and the test runner alongside the data-driven pass. The `generator` option's description in src/lib/ConfigOptions.json now mentions the script-driven flavor. A latent Lua-bridge bug is fixed in src/lib/Support/Lua.cpp: `domValue_push` now marshals `Undefined` (to `nil`) and `SafeString` (to a string) rather than aborting, so reading an absent field (for example the global namespace's empty name) no longer crashes a Lua script.
- **Tests**: A unit suite drives both runners against a synthetic corpus, exercises the output writer's path safety (writes under the root; rejects absolute and escaping paths), and checks that discovery installs a `ScriptGenerator`. A regression test reads a symbol with no name field, covering the `Undefined`-to-`nil` marshalling that a real corpus needs.
- **Documentation**: A new "Script-driven generators" section in `docs/modules/ROOT/pages/generators.adoc`, plus the regenerated docs/mrdocs.schema.json.
- **Breaking changes**: None. The script-driven flavor is additive; data-driven discovery and output are unchanged.

## Testing

- src/test/lib/Gen/script/ is the unit-level coverage: discovery, the output writer's path-escaping guards, and both the Lua and JavaScript runners producing the expected aggregated file.
- A script-driven generator's output is an arbitrary tree, which the per-file golden harness (one expected file per source) cannot check; the mechanism is therefore covered by the unit suite rather than by goldens.
- The `Undefined`-to-`nil` regression test guards the Lua-bridge fix, which also benefits the #1196 corpus-transform scripts that read symbol fields.
- No CI workflow changes are needed.

## Documentation

- A new "Script-driven generators" section in docs/modules/ROOT/pages/generators.adoc covers the `generate(corpus, output)` entry point, the `output.write` API, the `script:` discovery marker, and a search-index example.
- The `generator` option's description in src/lib/ConfigOptions.json now mentions script-driven generators.


alandefreitas:
Nice! Let me split this review into two points:

**On the contribution**

- In the documentation example, in what directory would output.write write? 
- How can the user see the configuration values and so on? 
- How can the extension receive its own parameters? 
- The extensions model we had envisioned (and discussed a lot in the last PR) is that a single extensions would contain all kinds of functionalities. The entry point is described as generate (in the same file, right?) but I don't see the word "generate" anywhere in the example script of the documentation. Why is that? Is it just my lack of skills with lua? The transform_corpus example does say `function transform_corpus`.
- As a side note, could you please put this new documentation content on a new page? I'll reorganize it later in nav.adoc. I'm working on documentation right now, and this will have its own page.

**On the use of AI**

I think this is a little out of hand, to be honest. 😅 And it's creating a bottleneck rather than being useful at this point.

> The design, and how each of its open questions was resolved, is recorded in the https://github.com/cppalliance/mrdocs/issues/1216 design comment

I don't find this very useful at all. Let me try to explain what I find useful first:

The way to make contributions more useful and less extractive is by having a **human in the loop**, adding human entropy, and being able to quickly answer questions about the human decisions. The https://llvm.org/docs/AIToolPolicy.html explains this well. Even the PR summary becomes less and less useful if there's no human in the loop, because it's harder and harder to read without intentionality. Only humans know what decisions were made by humans and why. Letting AI do it because AI doesn't know what was on the human's mind. AI doesn't know what you see as important or not. And the reviewer won't know it if AI generates the report. So that's the only part of the description that is actually useful to the contribution. Everything else could be generated by the reviewer just asking Claude to do it the same way. The reviewer was getting a *better* reply because Claude would make the answer according to the preferences of the reviewer. 

I think the word **loop** here is also very important. When asked something like "why did you do that", humans operate on improving the design of things that are highest ROI. Or, at the very least, humans would answer with the things that have the highest ROI first. AI doesn't know what has the highest ROI to humans, so it tends to create very long answers that are completely inappropriate for the audience, doesn't use the internal vocabulary, includes metaphors, doesn't order things by internal importance, or any of that. AI answers are low ROI: return is low because they don't answer what's important first and how to phrase it to the audience, because they can't, and investment is high because you have to read this very long text that gives you no information.

And, of course, there's value in the human taking accountability over the code that was generated. Because without that, the "contribution" becomes about throwing this work on the reviewer. The human should take accountability for the entire code being reviewed and be able to answer questions with a human response immediately and with confidence. It's very hard to pin down a single metric for that because it's very easy to perceive it when it's not there. If that's not the case, and the answers also come from AI, then there's no contribution here because, again, the reviewer could just ask AI the question directly and avoid the friction.

I know we're excited about AI because it generates a lot of code for us, but trying to bypass the human contribution and answering things with AI only creates more bottlenecks. We have 8 open PRs right now because code generation is faster, but CI and reviews are the bottleneck. We can't skip review; we can't merge code that's not good in this project, since it's not a new project. Even if we trust AI, we still have the specification to check. So only humans in the loop can break this bottleneck by helping prioritize with human entropy. Otherwise, we'll just have more and more blocked PRs, and IA will not help us move things forward.

Of course, AI can still help us with text-related tasks. As I type this, Grammarly is making grammar suggestions. But this is the gist of what I'm saying is human. I often also ask Claude to reorganize text into sections differently, and so on. But the content and its ROI are still accountable, reviewed, and aligned with the vocabulary we use internally. 

For instance, AI fluff paragraphs like:

> D3 - One context, one call. Build the script context once, evaluate the script once, and call generate once. This matches corpus transforms (a single whole-corpus call), not helpers (many per-node calls), so the existing single-shot extension machinery applies almost unchanged.

just wastes me half a minute to read without understanding. Then, it takes much more time because I have to reverse engineer these distant AI catch phrases ("One context, one call"), metaphors (AI never writes things descriptively, which is very inappropriate in technical contexts), expressions that are not part of the project vocabulary ("single-shot extension"), and these unnecessarily long and complex sentence hierarchies. The whole document would take 15 minutes to read, and it's mostly fluff before getting into the PR. And content-wise, I'm sure no human would say this is high ROI content in the context of this PR. And the way AI works (predicting the next word retroactively) makes it the opposite of a design document because it's retroactively justifying the decisions. And even if the text were good, it doesn't reflect our internal vocabulary at all, which is really weird. And even if all of that were false, the lack of human entropy (to communicate what's on the human's mind) would still mean the reviewer could get a better version of it by asking Claude locally. 

And the document is about 50x longer. Meaning it's all this pain times 50. We know that, in practice, most people read the first 10 words, notice it's AI, and ignore it completely. We need a human in the loop somewhere.

gennaroprota:
> **On the contribution**
> 
>     * In the documentation example, in what directory would output.write write?

In the configured output directory. The same the built-in generators use. I added this to the docs.

>     * How can the user see the configuration values and so on?

Hmm, they couldn't. Implemented now.

>     * How can the extension receive its own parameters?

Likewise. `generate` now has `config` and `params` as trailing parameters.
 
>     * The extensions model we had envisioned (and discussed a lot in the last PR) is that a single extensions would contain all kinds of functionalities.

Hmm... maybe I misunderstood you, but I think we are waiting for that (see issue #1210).

> The entry point is described as generate (in the same file, right?) but I don't see the word "generate" anywhere in the example script of the documentation. Why is that? Is it just my lack of skills with lua? The transform_corpus example does say `function transform_corpus`.

Yeah. It's a doc inconsistency, not your Lua :-). The example uses Lua's return-an-anonymous-function idiom (`return function(corpus, output)...`), which our Lua runner happens to accept, so the name never appears. It should be rewritten using `function generate(corpus, output)`. And, yes, the function should be in the file named by `script:` in mrdocs-generator.yml.

Given the above, I wonder: should we keep accepting the returned-function form? Or should we require `generate` (like we do for JavaScript)?

>     * As a side note, could you please put this new documentation content on a new page? I'll reorganize it later in nav.adoc. I'm working on documentation right now, and this will have its own page.

Yes, done.

> **On the use of AI**

Yeah. The PR summaries you generated with AI weren't particularly good either, so I've rewritten them by hand.

alandefreitas:
Hi! 

This needs rebase (more than others) because af3fc8337932caf832a5766c624a4160af12c324 is going to have an impact on this. The new nav.adoc will also have an impact on docs.

> Or should we require generate (like we do for JavaScript)?

I believe we should require it because one requirement of extensions is they need to be able to provide more than one extension. 

We don't have to fix that issue with the "layers of extensions" models in this PR, but it's quite urgent after this because we need to be able to represent this so we can provide meaningful examples for the user that show how to put things together.

alandefreitas:
In terms of design. I don’t understand the model we’re at. I think we were in level 1 going to level 2 (I mean, in the way you defined the levels in your issue) but now it seems like we went to some level 0.5. It seems like the other extension system comes from addons/extensions, and these ones come from addons/generator. But that can’t be different. Wasn’t the original idea that a extension could contain a transform_corpus and generate function? They can’t come from different places even if they weren’t related because the naming wouldn’t be appropriate.

It also seems like one of them requires a manifest and the other doesn’t but that doesn’t seem to make much sense. I think you tried to implement this system at level 2 and left the other system at level 1 if I understand correctly.

Also, it seems like everything I mentioned the user might need in an extension is being appended as an extra parameter but I’m not sure that’s the best design either.

As for the docs, it’s be nice to have the same model as the page for corpus extensions: scripts in both languages in tabs, sections explaining each thing the user can do, and examples interleaved with the concept being explained rather than singled out in a section.

PR #1237, opened 2026-06-22
feat: uniform attribute extraction

### LINKED ISSUE #1233: [[deprecated]] could be documented
I have this deprecated trait in a class:

```
	using extension_type [[deprecated("use extent_type")]] = extent_type;
```

There is no automatic mention of "deprecated" (of course, this can be done manually through normal documentation)

Image

Image

alandefreitas:
Reads attributes for every kind of symbol and shows them in the generated pages. Before, it only read attributes on functions and variables, and a deprecated symbol gave no sign that it was deprecated. For other symbols, it showed nothing. Now every symbol carries its attributes, deprecated symbols get a notice with the message in an admonition, and attributes appear in signatures and member tables. 

Fixes #1233.

## Changes

In the source code, attributes now live on the common symbol base class. Each attribute now contains its name and all its arguments. The templates render them in each signature, add a deprecation notice, and tag deprecated entries in member tables the same way constructors are tagged. The output schema and the golden tests were updated to match.

## Testing

New metadata fixtures cover attribute extraction for each symbol type. A new generator example renders the attributes to AsciiDoc and HTML.

## Documentation

The data model reference section now describes the new attribute fields.

### DIFF
--- a/docs/modules/ROOT/pages/extensions/dom-reference.adoc
+++ b/docs/modules/ROOT/pages/extensions/dom-reference.adoc
@@ -76,6 +76,64 @@ The `Symbol` object represents a symbol extracted from the source code. The symb
 | `Any`
 | The documentation for the symbol.
 
+| `attributes`
+| `<<attribute-fields,Attribute Object[]>>`
+| The C++ attributes attached to the symbol, for example `[[deprecated]]` or `[[nodiscard]]`.
+
+|===
+
+[#attribute-fields]
+=== Attribute
+
+The `Attribute` object represents a single C++ attribute. Every attribute object has these common fields:
+
+|===
+|Property |Type| Description
+
+| `kind`
+| `string`
+| The recognized standard attribute, identified independently of spelling: one of `deprecated`, `nodiscard`, `maybe-unused`, `no-unique-address`, `noreturn`, `carries-dependency`, `fallthrough`, `likely`, `unlikely`, `assume`, `indeterminate`, or `other` for an unrecognized attribute.
+
+| `name`
+| `string`
+| The attribute name. Recognized attributes use the normalized spelling (e.g., `deprecated`, `no_unique_address`); unrecognized ones keep the written spelling, including any scope (e.g., `gnu::custom`).
+
+| `balancedTokens`
+| `string[]`
+| The arguments as a balanced-token sequence, one per element (e.g., `["printf", "1", "2"]` for `[[gnu::format(printf, 1, 2)]]`). String-literal arguments keep their quotes. Empty for attributes that take no arguments.
+
+|===
+
+Some attribute kinds add a field for an evaluated argument. Templates should branch on `kind` and read these directly (for example, the deprecation notice prints `message`) rather than parsing `balancedTokens`.
+
+When the attribute kind is `deprecated`, the attribute object has the following additional property:
+
+|===
+|Property |Type| Description
+
+| `message`
+| `string`
+| The deprecation message, without quotes (e.g., `use bar instead` for `[[deprecated("use bar instead")]]`). Empty when no message was given.
+|===
+
+When the attribute kind is `nodiscard`, the attribute object has the following additional property:
+
+|===
+|Property |Type| Description
+
+| `reason`
+| `string`
+| The reason the result should not be discarded, without quotes (the C++20 `[[nodiscard("reason")]]` form). Empty otherwise.
+|===
+
+When the attribute kind is `assume`, the attribute object has the following additional property:
+
+|===
+|Property |Type| Description
+
+| `expression`
+| `string`
+| The assumed expression (the C++23 `[[assume(expression)]]` form).
 |===
 
 Handlebars generators extend each symbol with the following fields:
@@ -217,10 +275,6 @@ When the symbol kind is `function`, the symbol object has the following addition
 | `bool`
 | Whether the function is deleted as written.
 
-| `isNoReturn`
-| `bool`
-| Whether the function is noreturn.
-
 | `hasOverrideAttr`
 | `bool`
 | Whether the function has the override attribute.
@@ -240,10 +294,6 @@ When the symbol kind is `function`, the symbol object has the following addition
 | `isFinal`
 | `bool`
 | Whether the function is final.
-
-| `isNodiscard`
-| `bool`
-| Whether the function is nodiscard.
 
 | `isExplicitObjectMemberFunction`
 | `bool`
@@ -292,10 +342,6 @@ When the symbol kind is `function`, the symbol object has the following addition
 | `requires`
 | `string`
 | The `requires` expression of the function.
-
-| `attributes`
-| `string[]`
-| The attributes of the function.
 |===
 
 When the symbol kind is `typedef`, the symbol object has the following additional properties:
@@ -352,10 +398,6 @@ When the symbol kind is `variable`, the symbol object has the following addition
 | `initializer`
 | `string`
 | The initializer of the variable.
-
-| `attributes`
-| `string[]`
-| The attributes of the variable.
 |===
 
 When the symbol kind is `field` (i.e. non-static data members), the symbol object has the following additional properties:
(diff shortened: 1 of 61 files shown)


docs/modules/ROOT/pages/extensions/dom-reference.adoc
```diff
@@ -76,6 +76,28 @@ The `Symbol` object represents a symbol extracted from the source code. The symb
 | `Any`
 | The documentation for the symbol.
 
+| `attributes`
+| `<<attribute-fields,Attribute Object[]>>`
+| The C++ attributes attached to the symbol (e.g., `[[deprecated]]`, `[[nodiscard]]`). Present for every symbol kind; absent when the symbol has no attributes.
+
```
gennaroprota, line 82: "Present for every symbol kind" but then "absent when..." sounds strange :-). Perhaps we could say "Applies to every symbol kind"?

src/lib/AST/ASTVisitor.cpp
```diff
@@ -1738,26 +1735,149 @@ populate(
...
 }
 
-template <std::derived_from<Symbol> InfoTy>
+namespace {
+// Split the text inside an attribute's parentheses into its individual
+// arguments, breaking on top-level commas while respecting nested
+// brackets and string/character literals.
+std::vector<std::string>
+splitAttributeArgs(std::string_view const s)
+{
```
gennaroprota, line 1743: Do we normally use top-level `const` on function parameters?

test-files/golden-tests/snippets/options/extract-private-virtual/extract-private-virtual.adoc
```diff
@@ -36,6 +36,7 @@ class circle
 | Render the shape&period;
 |===
 
+
```
gennaroprota: Why the extra empty lines?

include/mrdocs/Metadata/Symbol/Function.hpp
```diff
@@ -189,10 +183,10 @@ MRDOCS_DESCRIBE_STRUCT(
     (ReturnType, Params, Template, FuncClass, Noexcept, Requires,
      IsVariadic, IsDefaulted, IsExplicitlyDefaulted, IsDeleted,
      IsDeletedAsWritten, IsNoReturn, HasOverrideAttr, HasTrailingReturn,
-     IsNodiscard, IsExplicitObjectMemberFunction, Constexpr,
+     IsExplicitObjectMemberFunction, Constexpr,
```
gennaroprota, line 183: I think you forgot to remove `IsNoReturn`.

src/lib/AST/ASTVisitor.cpp
```diff
@@ -1738,26 +1735,149 @@ populate(
...
+        cur += c;
+    }
+    if (std::string_view const last = trim(cur); !last.empty())
+    {
+        args.emplace_back(last);
+    }
+    return args;
+}
+} // (anon)
+
```
gennaroprota, line 1892: Very nice implementation. It's only con is it doesn't correctly handle raw string literals (very rare, I guess, in an attribute, but I thought to mention that).

gennaroprota:
Hi Alan, very good PR. I only left some minor comments. The most important one is about the duplication of `IsNoReturn`.

share/mrdocs/addons/generator/common/partials/symbol/section/deprecated-admonition.hbs
```diff
@@ -0,0 +1,6 @@
+{{! Deprecation admonition, shown when the symbol carries a [[deprecated]] attribute. }}
+{{~#each symbol.attributes}}
+{{~#if (eq name "deprecated")}}
```
gennaroprota: I think this won't match `[[gnu::deprecated]]` (not sure if it will match `__declspec(deprecated)`).

alandefreitas: I noticed this one when writing it. I thought the template would become a mess if we tried to detect all kinds here, with little benefit. Because people who don’t use these other deprecated attributes usually do so because they don’t have access to C++20, but MrDocs does. 

So if they use [[gnu::deprecated]], they probably are using macros for mrdocs and other C++20 compilers to see and document [[deprecated]] instead. So I thought accepting other kinds was just reinforcing the error somehow.  So the template would become giant for very little benefit.

The solution I had in mind for this was to propose an optional feature that normalizes non-standard attributes if people complain. This would happen long before the templates. But I can’t really propose this now because I’m not so sure how that would go and if people actually care.

What do you think?

include/mrdocs/Metadata/Symbol/Function.hpp
alandefreitas: Done

share/mrdocs/addons/generator/common/partials/symbol/section/deprecated-admonition.hbs
gennaroprota: Yes, I agree, normalization is the way to go; but let's wait to see if people actually care.

test-files/golden-tests/snippets/options/extract-private-virtual/extract-private-virtual.adoc
alandefreitas: There's a template there but handlebars also creates an extra line when the template is empty. Let me see what I can do about that.

src/lib/AST/ASTVisitor.cpp
alandefreitas: 😮

src/lib/AST/ASTVisitor.cpp
alandefreitas: Yes. This is the literal expression because clang doesn't give us the individual arguments in the general case. 😢 

This is literally parsing `std::string const text = getSourceCode(attr->getRange());`. 

On the other hand, I think we need the literal expression anyway. Otherwise, we would have to reconstruct the attribute on the other side.

I'll investigate the actual grammar and come up with something here.

docs/modules/ROOT/pages/extensions/dom-reference.adoc
alandefreitas: Fixed

test-files/golden-tests/snippets/options/extract-private-virtual/extract-private-virtual.adoc
alandefreitas: Fixed

src/lib/AST/ASTVisitor.cpp
alandefreitas: I studied the C++ grammar a little better and came up with a good representation for this. We have one representation that matches the C++ grammar and another with the semantics implied by clang. In fact, the comma is just a convention people use. Even that is not really official, let alone evaluations, quoted strings, and all of that. So the C++ grammar allows us to replicate it in the synopsis and the clang interpretation gives us unquoted strings and so on, which we can use for other purposes, such as admonitions.

share/mrdocs/addons/generator/common/partials/symbol/section/deprecated-admonition.hbs
alandefreitas: https://github.com/cppalliance/mrdocs/pull/1237#discussion_r3454376312

This also solved this issue because we now reuse the semantic interpretation from clang to determine what to render. :) Clang converts of all these to the same attribute internally.

alandefreitas:
@gennaroprota I believe everything is handled. Thank you so much. Those were good comments. 🙏

correaa:
i am curious how this will look like in the produced docs.

alandefreitas:
> i am curious how this will look like in the produced docs.

Well... the golden tests already include all the possible variants.

correaa:
is there a link you can point to. or a screenshot?

alandefreitas:
https://github.com/cppalliance/mrdocs/blob/develop/test-files/golden-tests/generator/hbs/attributes/attributes.adoc

correaa:
Excellent!

image

https://correaa.gitlab.io/boost-multi/multi/reference/boost/multi/array-0ad.html

PR #1239, opened 2026-06-25
fix: omit the synthesized initializer for default-constructed variables

### LINKED ISSUE #1238: Spurious initializer shown for default-constructed variables
When documenting mp-units, variables that are default-constructed instances of a type ~~having their same name~~ are rendered with an initializer (which doesn't appear in the source). For example, the synopses show:

```
inline constexpr one one = one;
inline constexpr percent percent = percent;
inline constexpr pi pi = pi;
inline constexpr pi_c pi_c = pi_c;
```

The corresponding declarations are:

```
inline constexpr struct one final : detail::derived_unit_impl<> {} one;
inline constexpr struct percent final : named_unit<"%", mag_ratio<1, 100> * one> {} percent;
inline constexpr struct pi final : named_constant<symbol_text{u8"π", "pi"}, mag<pi_c> * one> {} pi;
inline constexpr struct pi_c final : mag_constant<symbol_text{u8"π", "pi"}, std::numbers::pi_v<long double>> {} pi_c;
```

So `one` should render as `inline constexpr one one;`, with no `= one`, and likewise for the others.

gennaroprota:
Default-constructed variables (`T t;`) were rendered with a spurious initializer (`= t`). That's because Clang represents them with an implicit construction expression located at the variable name; so the extracted initializer text was the variable name itself.

Therefore, skip the initializer when it is an implicit default construction.

Closes #1238.

## Changes

- **Source**: ASTVisitor.cpp now skips the synthesized initializer provided by Clang when the variable is default-constructed.

## Testing

A new golden test, with XML output, tests a handful of variable declarations.

## Documentation

No documentation needed. This is a bug fix.


alandefreitas:
It looks good to me. The only thing I don't like is the comment in the AST visitor. The comment is huge, and it's obviously AI slop because it's huge and it contains very little information. No human would go into this level of detail because none of the details provided there are relevant in the context. It's very easy to describe what this function does in this AST context with less than one line of code. Still, AI always expands on things way more than we need to. If we keep doing that, the documentation becomes very incoherent because AI is only focused on this change, but the rest of the file doesn't follow this structure. Anyone reading the file without looking at this specific comment will stop here and think this change is way more important or tricky than it is. I mean, anyone, including another AI agent, could be confused by that. In a way, I'm kind of doing the same here, but the only reason I'm talking about this so much is that it's a recurring comment I'm making on all the PRs lately.

gennaroprota:
Oh, actually the comment was meant to explain what, for me, was indeed a tricky point: the fact that a variable with no written initializer still gets an implicit initializer in the AST which is located at the variable name. It probably isn't so tricky for someone more used to how Clang represents things.

alandefreitas:
Oh. OK. Fair enough. Thanks.

PR #1241, opened 2026-07-02
chore: bump LLVM to 77e43ec1 (May 21, 2026)

gennaroprota:
The new pin includes llvm/llvm-project#198452, which attaches documentation comments to macro definitions. That is the prerequisite for reading them from Clang directly instead of scanning the source, as we were initially doing in PR #1192.

## Changes

- **Third-party**: Bump LLVM to 77e43ec1 (llvm/llvm-project#198452).
- **CI**: Update the hardcoded LLVM hash in the cache key (ci-matrix.yml) and its mirror unit test (test_cache_keys.py) to match the new pin, so the key misses and LLVM builds fresh instead of restoring the old install. Build an instrumented libc++ for MSan only (the newer libc++ needs the latest Clang, and ASan works with the normal libc++), and prune the LLVM source tree's stray .profraw fixtures.
- **Source**: The pin was about six months behind, so this also adapts MrDocs to the Clang API changes across that range:
    - USRGeneration.h moved from clang/Index/ to clang/UnifiedSymbolResolution/; the `clang::index` namespace and signatures are unchanged.
    - `clang::DiagnosticConsumer::finish()` was removed, so the collecting diagnostic consumer relies on the forwarded `EndSourceFile()` instead.
    - The driver option table moved to clang/Options/Options.h, with its enumerators now in `clang::options` and `getDriverOptTable()` in `clang`.
    - `cl::getRegisteredOptions()` now returns a `DenseMap`, so its entries expose `.second` instead of `getValue()`.
    - `SubstituteConstraintExpressionWithoutSatisfaction` was re-synced with Clang's current implementation.

  `getRawCommentForDeclNoCache`, the per-declaration comment lookup `getDocumentation` relied on, is no longer public. Its only public replacement, `getRawCommentForAnyRedecl`, follows the redeclaration chain and so cannot return a single declaration's own comment. `getDocumentation` keeps the per-declaration behavior by porting Clang's declaration-path comment search (`getLocsForCommentSearch` and `getRawCommentNoCacheImpl`) as file-local helpers.

  Parsing the newer libc++ also needs a new vcruntime_new.h stub and C++ guards so stdbool.h, stdalign.h, and threads.h stop redefining `bool`, `alignas`, and `thread_local`, which are keywords, not macros, in C++.
- **Golden tests**: The regenerated goldens change only because Clang's output drifted over the six months: constrained templates and destructors get different symbol IDs from USR generation, though their structure is identical (requires-clause, dtor-overloads), and a few declaration locations shifted by a line or column (auto, param-types, function-template).

## Testing

No new tests are added; this is a toolchain bump. The existing unit and golden suites already cover the affected behavior and run in CI on every build, and the goldens whose output changed have been regenerated as described above.

## Documentation

No documentation changes are needed: this is an internal toolchain bump with no user-facing API changes.


.github/workflows/ci-build.yml
```diff
@@ -156,6 +156,14 @@ jobs:
           build-dir: build
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      # The extracted LLVM source tree ships its own .profraw test fixtures,
+      # in an older raw-profile version that llvm-profdata rejects. They are
+      # not ours, so drop them before the coverage step globs build/ for
+      # *.profraw.
+      - name: Prune third-party profraw
+        if: matrix.coverage
+        run: find build/third-party -name '*.profraw' -delete
```
alandefreitas: What’s this profraw issue? The explanation is more descriptive than explanatory of the motivation of the command. It seems to me like if this is happening and coverage is globing the wrong directory, that’s what should be fixed instead of a second workaround to accommodate the first mistake.

.github/workflows/ci-matrix.yml
```diff
@@ -42,11 +42,16 @@ jobs:
           standards: '23'
           main-entry-factors: |
             clang Coverage
+          # MSan builds an instrumented libc++ from the pinned LLVM sources,
```
alandefreitas: I don’t like some of the excessive comments that likely come from using AI. The thing with AI is that it doesn’t know how to not say things. For instance, you tell AI to not say “X” and very often it gives you text like “I’m not even going to bring up X”.

Because it’s a text-generating machine. It has no button for not generating text.

For instance, in this particular PR, it’s obvious AI was moving MSan to latest-factors and felt like it just to justify this in a comment. The comment is coherent with what it had in its context but it’s not coherent with the rest of the code. After this is merged, a person reading this file doesn’t know this context. This looks like it’s documentation for latest-factors and makes no sense there. The user wouldn’t even know this option was not in latest-factors before. The user won’t understand why msan and not latest-factors is explained here. They won’t understand why this is the only factor singled-out and so on.

.github/workflows/ci-matrix.yml
```diff
@@ -75,14 +80,8 @@ jobs:
           append-common-cxxflags: |
             clang: -gz=zstd
           extra-values: |
-            # libc++ runtimes: clang + (ASan or MSan) needs instrumented libc++
```
alandefreitas, line 78: Why are these gone? Weren't they being used? Interesting. I thought we used the custom libc++ whenever we had msan or asan. It it because the logic for this necessity is now being checked in the bootstrap script?

share/mrdocs/headers/libc-stubs/new.h
```diff
@@ -12,29 +12,7 @@
...
-void* operator new[]( std::size_t count, const std::nothrow_t& tag );
-void* operator new  ( std::size_t count, std::align_val_t al,
-                      const std::nothrow_t& tag ) noexcept;
-void* operator new[]( std::size_t count, std::align_val_t al,
-                      const std::nothrow_t& tag ) noexcept;
-
-// Non-allocating placement allocation functions
-void* operator new  ( std::size_t count, void* ptr );
-void* operator new[]( std::size_t count, void* ptr );
+#include <vcruntime_new.h>
```
alandefreitas, line 15: Interesting. Is this something being fixed in passing or do we have something about the new version that interacts with the libc stubs? In fact, I would have this question about all other changes to the libc stubs. It's unexpected to me that they interacted in an LLVM bump.

test-files/golden-tests/symbols/function/function-template.xml
```diff
@@ -315,7 +315,7 @@
       <short-path>function-template.cpp</short-path>
       <source-path>function-template.cpp</source-path>
       <line-number>22</line-number>
-      <column-number>6</column-number>
+      <column-number>1</column-number>
```
alandefreitas, line 318: These changes to line numbers and column numbers are curious. Since the fixtures didn't change, we need to check whether they were broken before or whether they're broken now. Or maybe they're referencing a different part of the symbol.

util/bootstrap/src/recipes/builder.py
```diff
@@ -45,7 +46,7 @@ def needs_libcxx_runtimes(sanitizer: str, compiler_id: str) -> bool:
     if not sanitizer:
         return False
     san = sanitizer_flag_name(sanitizer.lower())
-    if san not in ("address", "memory"):
+    if san != "memory":
```
alandefreitas: We could discuss this with Matheus, but I believe the changes to this file are incorrect. At the very least we lose some std lib annotations, and so on. ASan doesn't need sanitized dependencies as much as MSan but it's less effective without sanitized dependencies. And we're building+caching our own libc++ anyway so the cost is low.

alandefreitas (commented):
Nice bump :)

Reimplementing rawCommentForDecl ourselves ended up a little verbose. We could add the pinned commit (77e43ec1) to the comment on locsForCommentSearch so the next bump can compare the port against the exact revision it came from. We could also check why they removed it in their PRs, issues and git blame (doing that might even reveal a better way of doing things) and we should consider filing an issue upstream asking to re-expose a per-declaration lookup (making getRawCommentNoCache public would do). If Clang takes it, we can delete the port on a later bump.

share/mrdocs/headers/libc-stubs/vcruntime_new.h
```diff
@@ -0,0 +1,41 @@
...
+//
+// Copyright (c) 2024 Alan de Freitas (alandefreitas@gmail.com)
+//
+// Official repository: https://github.com/cppalliance/mrdocs
+//
+
+#ifndef MRDOCS_SHARE_HEADERS_LIBC_STUBS_VCRUNTIME_NEW_H
+#define MRDOCS_SHARE_HEADERS_LIBC_STUBS_VCRUNTIME_NEW_H
+
+// libc++ includes this Microsoft runtime header (through <__new/align_val_t.h>)
```
alandefreitas: This whole comment, especially "Provide just enough of it for parsing." seems redundant because this is true for all stubs by definition. They're headers libc++ includes for its own functionality and we provide only enough for parsing.

gennaroprota:
> Reimplementing rawCommentForDecl ourselves ended up a little verbose. We could add the pinned commit (77e43ec1) to the comment on locsForCommentSearch so the next bump can compare the port against the exact revision it came from. We could also check why they removed it in their PRs, issues and git blame (doing that might even reveal a better way of doing things) and we should consider filing an issue upstream asking to re-expose a per-declaration lookup (making getRawCommentNoCache public would do). If Clang takes it, we can delete the port on a later bump.

Good news! None of the above applies! :-) Turns out that I misread ASTContext.h and `getRawCommentNoCache` is public (the declaration appears to be in what seems a private section at first glance, but it's public)! So I'm using that now. Good that we re-examined this.

.github/workflows/ci-build.yml
gennaroprota: The issue is that the coverage step scans build/ for *.profraw, which happens to include the *.profraw test fixtures in the LLVM extracted source tree; but they are in an older format, which makes the llvm-profdata merge fail. I'm not sure why this issue didn't occur when the cache was cold, the first time. Yes, this would better be fixed in the action.

.github/workflows/ci-matrix.yml
gennaroprota: Since this was an unrelated change, I restored them. Anyway, the decision of whether to build an instrumented libc++ is made in bootstrap, even on develop.

test-files/golden-tests/symbols/function/function-template.xml
gennaroprota: It's an upstream change: in a declaration like `int f();` the line and column number used to be those of the 'f' (the beginning of the function name); now they are those of the 'i' in "int" (the beginning of the declaration).

share/mrdocs/headers/libc-stubs/new.h
gennaroprota: Good question. They do interact with the bump:

- The newer libc++'s `<__new/align_val_t.h>` now does `#include <vcruntime_new.h>` under the VC runtime ABI.
- `stdbool.h` / `stdalign.h` / `threads.h` used to define C++ keywords as macros; that's a bug which surfaced with the bump.

share/mrdocs/headers/libc-stubs/vcruntime_new.h
gennaroprota: Fair.

util/bootstrap/src/recipes/builder.py
gennaroprota: Agreed. I reverted that.

PR #1243, opened 2026-07-24
feat: pass extra compiler arguments from the config file

### LINKED ISSUE #9: additional Executor arguments from YML
There should be a YML key that lets the mrdox.yml file specify additional command line arguments to be appended to the passed command line arguments, which are then passed to the Executor of the front end action.

gennaroprota:
This adds support for an `extra-compiler-args` option (a list of strings whose values are appended verbatim to each translation unit's compile command). It is an escape hatch for compiler flags MrDocs does not model directly.

## Changes

- **Source**: `adjustCommandLine()` is extended to append any such arguments as specified in the config file.
- **Golden tests**: A new golden test fixture ensures `-D<macro_name>` reaches the compiler.

## Testing

A new golden test includes a config file that adds the command-line argument `-DMRDOCS_EXTRA_ARG`; the corresponding .cpp file uses that macro to guard a function declaration, and the test checks that the function is included in the XML output.

## Documentation

ConfigOptions.json provides the documentation for the new option.

Closes #9.

### DIFF
--- a/docs/mrdocs.schema.json
+++ b/docs/mrdocs.schema.json
@@ -153,6 +153,15 @@
       "title": "Symbol patterns to exclude",
       "type": "array"
     },
+    "extra-compiler-args": {
+      "default": [],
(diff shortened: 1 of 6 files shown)


alandefreitas:
https://github.com/cppalliance/mrdocs/issues/9#issuecomment-5075569976

PR #1256, opened 2026-08-11
feat: support plugins

### LINKED ISSUE #58: Add plugins/generators via DLL

gennaroprota:
This PR adds support for plugins, i.e. shared libraries written in C++ that can be written and compiled without needing Clang/LLVM headers.

## Changes

- **Source**:
  - A public header, mrdocs/Plugin.hpp. It says what a plugin has to export (an entry point, the interface version, a toolchain tag) and what MrDocs passes in (`PluginContext`). It also provides `MRDOCS_PLUGIN_MAIN`, a macro that writes those exported functions for the plugin author.
  - The loader. It finds the libraries in each addon root, in order, checks the version and the toolchain, and calls the entry point.
  - One call to the loader while the tool starts up, before any generator is looked up by id.
  - Export attributes. They now expand to a real attribute when MrDocs itself is built. They also move to a header of their own, because Platform.hpp includes Assert.hpp before it could define them.
- **Tests**: discovery unit tests, a stats example plugin run by ctest, and a link test under tests/plugin-api.
- **Build**:
  - The mrdocs target sets `ENABLE_EXPORTS`. It also lists, on its interface, the include directories and compile options a plugin needs.
  - The example plugin and the link test are built as module libraries. The example is built into an addon root that its own test passes to MrDocs.

## Testing

Each test checks something the others cannot.

- **The discovery unit tests** check which files are loaded, in what order, and that a library reachable through two roots is loaded once. Those rules are part of the documented behavior, so they should not change by accident.
- **The stats example** runs MrDocs with a configuration that asks for `generator: stats`. That name only exists if the plugin was found, loaded, and installed its generator, so the test fails if any step in that chain breaks.
- **The link test** checks that the API a plugin uses is really exported. A plugin compiles its own copy of every inline and template member, and those copies call functions that live in the tool and need an export attribute. If one of them is missing, the plugin does not link, and no other test would notice.

Writing the link test found a real problem. The comparison operators of the symbol kinds were not exported, so a plugin that sorts members could not link. The first commit fixes that.

The two ctest tests run in suites that CI already runs. The link test has no ctest entry, because it does not need one: it is a library, and a missing symbol makes the build fail. So CI needs no change.

The link test only detects a missing export on Windows, where a plugin gets its symbols from the tool's export table. On Linux and macOS, it still checks that the public headers compile the way a plugin compiles them, and that such a library links against the executable.

One thing is not tested: the cases where the loader refuses a plugin, that is, a wrong interface version, a wrong toolchain, or a missing entry point. Testing those needs a library built the wrong way on purpose.

## Documentation

A new Plugins page under Extensions covers where plugins live, the entry point, the generator, and how to build one. Its code comes from the example through tag includes, so the page cannot drift from something that compiles. The plugins directory that ships with MrDocs gets a README saying what belongs in it.

Closes #58.


docs/modules/ROOT/pages/extensions/plugins.adoc
```diff
@@ -0,0 +1,121 @@
+= Plugins
+:url-native: https://github.com/cppalliance/mrdocs/tree/{page-origin-refname}/examples/generators/native
+
+A plugin is a shared library that Mr.Docs loads as it starts up, handing it a context. Through that context, the plugin installs what it provides, so the rest of the run sees it as if it had shipped with the tool: generators, which turn the corpus into output, and transforms, which change the corpus before any generator sees it.
+
+Reach for a plugin when the output needs {cpp}: a format whose rules are awkward to express in templates, a generator that depends on a system library, or one whose work is heavy enough to justify compiled code. The lighter options remain xref:extensions/data-driven-generators.adoc[data-driven generators] (templates only) and script-driven generators (Lua or JavaScript, see xref:extensions/corpus-extensions.adoc[Extensions]). The difference from xref:extensions/as-library.adoc[Mr.Docs as a library] is that a plugin runs inside the ordinary `mrdocs` tool: you keep the command line, the configuration, and everything else Mr.Docs does, and you do not build Mr.Docs or LLVM.
```
alandefreitas, line 6: Another interesting use case is linking C/C++ libraries. For instance, a plugin could use sqlite or something like that.

docs/modules/ROOT/pages/extensions/plugins.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+== Where plugins live
+
+Mr.Docs looks in the `plugins` subdirectory of every addon root, that is, under both the xref:configuration/reference.adoc#addons_option[`addons`] directory and each xref:configuration/reference.adoc#addons-supplemental_option[`addons-supplemental`] directory, in that order:
```
alandefreitas, line 10: The sections seem a bit out of order. I've been following this page trying to build a plugin to test the system but the section on how to build it is the last section and this is already about how to place it. The first step is always setting up the directories and the boilerplate. That can't be a detail left to the end.

docs/modules/ROOT/pages/extensions/plugins.adoc
```diff
@@ -0,0 +1,121 @@
...
+----
+<1> `.dll` on Windows, `.so` on Linux, `.dylib` or `.so` on macOS.
+
+Every library in such a directory is loaded, in name order within a root, and roots are visited in the order the configuration lists them. One library reachable through several roots is loaded once. Anything that is not a library is ignored, so a directory can hold a README, an import library, or debug information without confusing the search.
+
+A library that cannot be loaded, does not export the entry points, or reports an error of its own stops the run.
+
+== The entry point
+
+The `MRDOCS_PLUGIN_MAIN` macro defines the function Mr.Docs calls, naming the cpp:PluginContext[] parameter it receives:
```
alandefreitas, line 26: The macro needs to be linked to the reference documentation.

docs/modules/ROOT/pages/extensions/plugins.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+.`plugin.cpp`
+[source,cpp]
+----
+include::example$examples/generators/native/stats/plugin.cpp[tag=main]
+----
+
+The context is how a plugin reaches Mr.Docs: cpp:PluginContext::installGenerator[] adds a generator, cpp:PluginContext::installTransform[] adds a transform, and cpp:PluginContext::config[] returns the configuration the run was started with, already loaded and normalized, in case what the plugin installs depends on it. The reference is valid for the duration of the call.
+
+Along with the entry point, the macro defines a function that reports the version of the plugin interface the library was compiled against. Mr.Docs compares it with its own and refuses to call a plugin that was built against a different one. A plugin therefore has to be rebuilt when it is used with a Mr.Docs whose plugin interface has changed.
```
alandefreitas, line 36: Shouldn't we mention the macro for the plugin version be mentioned and linked here? And how does the versioning system work, etc? How do we developers ensure that this MRDOCS_PLUGIN_API_VERSION is always correct?

examples/generators/CMakeLists.txt
```diff
@@ -39,3 +39,30 @@ foreach (script_driven IN ITEMS search-index json)
...
+if (MRDOCS_BUILD_PLUGIN_MODULES)
+    add_library(mrdocs-stats-plugin-example MODULE native/stats/plugin.cpp)
+    target_link_libraries(mrdocs-stats-plugin-example PRIVATE mrdocs)
+    target_compile_features(mrdocs-stats-plugin-example PRIVATE cxx_std_23)
+
+    # MrDocs looks for plugins in the `plugins` subdirectory of an addon root,
+    # so the library is built into one. The generator expression keeps CMake
+    # from appending a per-configuration subdirectory, which would put the
+    # library somewhere the test cannot name.
+    set(MRDOCS_STATS_ADDONS "${CMAKE_CURRENT_BINARY_DIR}/native/stats/addons")
```
alandefreitas, line 52: examples/generators/native? We never see this term "native" anywhere else, and the other generators in examples/generators are probably more "native" than this one. Maybe example/plugins/stats?

docs/modules/ROOT/pages/extensions/plugins.adoc
```diff
@@ -0,0 +1,121 @@
...
+include::example$examples/generators/native/stats/plugin.cpp[tag=build]
+----
+
+== Transforming the corpus
+
+A plugin can also install a cpp:Transform[]: a pass that runs once, after the corpus is built and finalized and before any generator runs, so what it changes is what every output format sees. Where a generator is handed the corpus as `const`, a transform is handed it mutable, and may read it, change the symbols it finds, or both. What it cannot do is create a symbol or destroy one, since a corpus keeps its storage to itself. Everything a symbol holds is in reach, though, including the lists of members that the generators walk to decide what to write.
+
+[source,cpp]
+----
+class BriefFiller
```
alandefreitas, line 64: Hard-coded example 👎

examples/generators/native/stats/mrdocs.yml
```diff
@@ -0,0 +1,4 @@
+generator: stats
```
alandefreitas, line 1: I don't understand this example. Where's the project being documented with the plugin doing something with it? Is this the project that defines the plugin or the example that uses the plugin? In either case, where's the other project?

docs/modules/ROOT/pages/extensions/plugins.adoc
```diff
@@ -0,0 +1,121 @@
...
+        mrdocs::Config const& config) const override;
+};
+
+MRDOCS_PLUGIN_MAIN(context)
+{
+    return context.installTransform(std::make_unique<BriefFiller>());
+}
+----
+
+The transforms a plugin installs run before the ones an extension script registers with `mrdocs.register_transform`, and among themselves in the order they were installed. An error from a transform stops the run before any generator sees the corpus, and the diagnostic names the transform through cpp:Transform::id[].
```
alandefreitas, line 86: The example is missing the implementation.

docs/modules/ROOT/pages/extensions/plugins.adoc
```diff
@@ -0,0 +1,121 @@
...
+The macro reports the toolchain as well, and Mr.Docs checks that too, for the reason the section on building a plugin explains.
+
+== The generator
+
+A plugin's generator is an ordinary cpp:Generator[] subclass: the same interface the built-in formats implement, and the same one the xref:extensions/as-library.adoc[library] examples use.
+
+.`plugin.cpp`
+[source,cpp]
+----
+include::example$examples/generators/native/stats/plugin.cpp[tag=generator]
```
alandefreitas, line 47: A more realistic example would be great. Although I know that's hard.

alandefreitas (commented):
This is really nice. Especially that we can just use the same interface as the extensions.

What was the design process? How is the design compared to, for instance, clang plugins? 

It's also really nice that it has a system to check whether the versions are incompatible. How do we avoid extreme incompatibility between versions? At the very least, every time we update LLVM, the plugins will have an API breaking change. In practice, it seems like almost new Mr.Docs commit will imply in a breaking change that invalidates other plugins. A way to fix that would be maybe a more stable class that reconstructs the objects on the other side but I'm not sure if that's feasible. The macros delegating to an intermediary C API can also help some more but the structs can still break anyway.

Were all of these MRDOCS_DECL wrong? What happened down there?

CMakeLists.txt
```diff
@@ -63,6 +63,16 @@ set_ternary(MRDOCS_DO_BUNDLE_DEPENDENCIES "MRDOCS_BUNDLE_DEPENDENCIES AND NOT MR
...
 set_ternary(MRDOCS_GCC   "CMAKE_CXX_COMPILER_ID STREQUAL \"GNU\"" ON OFF)
 set_ternary(MRDOCS_CLANG "CMAKE_CXX_COMPILER_ID MATCHES \"Clang$\"" ON OFF)
+
+if (CMAKE_CXX_FLAGS MATCHES "(^| )-static($| )" OR
+    CMAKE_EXE_LINKER_FLAGS MATCHES "(^| )-static($| )")
+    set(MRDOCS_PLUGIN_MODULES_DEFAULT OFF)
+else ()
+    set(MRDOCS_PLUGIN_MODULES_DEFAULT ON)
+endif ()
+option(MRDOCS_BUILD_PLUGIN_MODULES "Build the plugin example and its link test"
```
alandefreitas, line 73: So our releases won't support plugins anyway? What's the point?

libs/handlebars/include/mrdocs/Handlebars/Platform.hpp
```diff
@@ -13,19 +14,26 @@
...
+// expands to nothing either way, and the symbols are reachable because the tool
+// is linked with its exports enabled. A static consumer of the library, which
+// defines MRDOCS_STATIC_LINK and not MRDOCS_TOOL, still sees plain declarations.
 
-#if defined(MRDOCS_STATIC_LINK)
-# define MRDOCS_HANDLEBARS_DECL
-#elif defined(_MSC_VER)
-# if defined(MRDOCS_TOOL)
+#if defined(_MSC_VER)
+# if defined(MRDOCS_TOOL) // building MrDocs
```
alandefreitas, line 28: Isn't this breaking encapsulation?

docs/mrdocs.yml
```diff
@@ -17,16 +17,8 @@ file-patterns:
...
-# (INFO/LOG/F) and compat shims used across the headers. Keep them out of
-# MrDocs's own reference.
-exclude-macros:
-  - 'MRDOCS_*'
-  - 'INFO'
-  - 'LOG'
-  - 'F'
-  - 'FMT_CONSTEVAL'
+include-macros:
+  - 'MRDOCS_PLUGIN_*'
```
alandefreitas, line 21: Mmm... Our convention for macros is bad. It was bad and it remains bad. It's more than excluding MRDOCS_* and it's more than includeing MRDOCS_PLUGIN_* but it's less than includeing MRDOCS_*.

alandefreitas (commented):
Sorry. I clicked on submit review by mistake before finishing it. 

A plugin must match the compiler major, the standard library, and on MSVC the iterator debug level and CRT, and so on and so on. Do we also have a system in place to check all of that? That's the C boundary thing I was suggesting maybe could work. I'm not sure how other systems do it. In any case, nothing in the docs or the release notes says what that is. The person downloading mrdocs after this commit won't have access to plugins because cmakelists.txt disables it for -static builds (removing -static breaks the release across OS versions by the way - this is there for a reason). 

MRDOCS_DECL now means dllexport whenever MRDOCS_TOOL is defined, even in a static build? Is this correct?

Can't the tests easily check the path where MRDOCS_PLUGIN_API_VERSION isn't accepted?

docs/modules/ROOT/pages/extensions/plugins.adoc
gennaroprota: Yes, mentioned that as an example.

docs/modules/ROOT/pages/extensions/plugins.adoc
gennaroprota: Fixed.

docs/modules/ROOT/pages/extensions/plugins.adoc
gennaroprota: Done.

docs/modules/ROOT/pages/extensions/plugins.adoc
gennaroprota: > Shouldn't we mention the macro for the plugin version be mentioned and linked here?

Yes, done now.

> And how does the versioning system work, etc?

This is explained in the new section at the end of the page. The macro's value is compiled into the plugin, MrDocs compares it with its own before calling the plugin, and refuses on a mismatch. The value changes whenever the plugin API changes. The section also says what the plugin API is: the context, the generator and transform interfaces, and the corpus and configuration they are handed. A change elsewhere in the headers does not count.

> How do we developers ensure that this MRDOCS_PLUGIN_API_VERSION is always correct?

Good question. At the moment, it's up to us to raise the number when we change what a plugin sees. A better way would be to encode the MrDocs version into the build tag, so a plugin built against another release is refused. But that still leaves the rolling releases that we publish through CI. Or builds from source that someone might do.

examples/generators/CMakeLists.txt
gennaroprota: Yeah. Moved to examples/plugins/stats now.

docs/modules/ROOT/pages/extensions/plugins.adoc
gennaroprota: Fixed.

examples/plugins/stats/sample-project/mrdocs.yml
```diff
@@ -0,0 +1,4 @@
+generator: stats
```
gennaroprota: Hmm, yes, it was one directory holding two different things with nothing saying so. It is laid out now:

```
examples/plugins/stats/
├── CMakeLists.txt
├── README.md
├── plugin.cpp            <- the plugin
└── sample-project/       <- the project it is run on
    ├── mrdocs.yml
    ├── simple.cpp
    └── stats.txt
 ```

docs/modules/ROOT/pages/extensions/plugins.adoc
gennaroprota: Oops, the page shows the implementation now.

docs/modules/ROOT/pages/extensions/plugins.adoc
gennaroprota: What about a plugin listing every symbol carrying `[[deprecated]]`, with the replacement text taken from the attribute?

CMakeLists.txt
gennaroprota: Only the Linux releases, actually.

Now, they build in an ubuntu:26.04 container, so a dynamically linked artifact would take a 2026 glibc as its floor, and would not start on Ubuntu 24.04 and many other current distributions.

I think we should build those releases with an old glibc and a new GCC, with `-static-libstdc++ -static-libgcc` but without `-static`. If you agree, I'd do this in this PR but with a separate commit.

gennaroprota:
> Sorry. I clicked on submit review by mistake before finishing it.
> 
> A plugin must match the compiler major, the standard library, and on MSVC the iterator debug level and CRT, and so on and so on. Do we also have a system in place to check all of that?

Yes, and it checks the list you wrote, almost item by item. Each plugin reports a build tag, and MrDocs refuses a tag that doesn't match its own. The tag depends on the compiler and its major version, the standard library and some settings: on MSVC, the STL update, the iterator debug level and whether the CRT is linked statically or dynamically; on libc++, the version and the ABI version; on libstdc++, the release and whether debug mode is on.

> That's the C boundary thing I was suggesting maybe could work.

The tag check is done by `extern "C"` functions.

> I'm not sure how other systems do it. In any case, nothing in the docs or the release notes says what that is. The person downloading mrdocs after this commit won't have access to plugins because cmakelists.txt disables it for -static builds (removing -static breaks the release across OS versions by the way - this is there for a reason).

Yes. See my comment above about removing `-static`. About the docs not saying any of this, there's now a closing section on the Plugins page covering what has to match and what happens when it does not.

> MRDOCS_DECL now means dllexport whenever MRDOCS_TOOL is defined, even in a static build? Is this correct?

Yes. `MRDOCS_TOOL` means "this translation unit is being compiled into MrDocs itself". It does not mean that MrDocs is a shared library.

> Can't the tests easily check the path where MRDOCS_PLUGIN_API_VERSION isn't accepted?

Oh yes, they do now (see tests/plugin-api/ForeignVersion.cpp).

libs/handlebars/include/mrdocs/Handlebars/Platform.hpp
gennaroprota: Yes, although that was a pre-existing issue. I'll fix it.

docs/mrdocs.yml
gennaroprota: Yeah.

PR #1267, opened 2026-08-28
feat: link references to symbols documented elsewhere (input tagfiles)

### LINKED ISSUE #1072: support for reading tagfiles
#670 included support for generating tagfiles in the adoc generators. 

We should also implement support for reading tag files, so that `@ref` commands, etc., can also refer to elements outside the corpus.

gennaroprota:
MrDocs can now read external tagfiles and, thus, handle a doc comment that references a symbol not extracted by MrDocs itself. A new option, `input-tagfiles`, specifies the external tagfile(s) and maps each one to the base URL its documentation set is published under. A reference that resolves to nothing in the corpus takes its URL from there instead of rendering as plain code. The name is resolved from the scope it is written in outward, as any other reference is, and is never shortened to reach an entry. Every generator takes the URL from one place.

Tagfiles are read with a small ad-hoc parser, rather than by LibXml2: the format is a small, machine-generated subset of XML, and linking LibXml2 into mrdocs-core would make it a dependency of every build and of every consumer of the installed package.

A path that cannot be read, or a file that is not a tagfile, fails the build.

The option MrDocs already had for writing a tagfile (named `tagfile`) becomes `output-tagfile`, so that the pair of options says which direction each of them goes; `tagfile` remains as a deprecated alias. Note that all deprecated options warn now, including `report`, which didn't before.

## Changes

- **Source**: the tagfile index and its reader; the `input-tagfiles` option and loading it into the corpus; `Corpus::externalUrl` beside `Corpus::lookup`; `href` on `ReferenceInline` and the branch that renders it; the `tagfile` -> `output-tagfile` rename with a deprecated alias; generic, deferred reporting of deprecated options.
- **Tests**: unit suites for the index and for the reader (`clang.mrdocs.TagfileIndex`, `clang.mrdocs.TagfileReader`).
- **Golden tests**: a `config/input-tagfiles` fixture (xml, adoc, html) covering a fully qualified name, a name relative to its enclosing scope, a namespace page, a member with an anchor, a member on its own page, and names that resolve to nothing; the `config/tagfile` fixture is renamed to `config/output-tagfile` and regenerated for its new file names.
- **Build**: the new options land in `ConfigOptions.json`, so the generated
  config schema and the published option reference change; the new `href` field regenerates the reflected DOM schema (`mrdocs.rng`, `generators/mrdocs.schema.json`, `dom-schema.adoc`).
- **Tooling**: the in-repo callers that passed `--tagfile` now pass pass `--output-tagfile`.
- **Breaking changes**: `tagfile` still works but warns. The long-deprecated `report` option starts warning too. Under `warn-as-error`, either fails the build.

## Testing

The unit suites test the index and the reader, including the inputs it must reject; the golden fixture tests resolution end to end in XML, AsciiDoc and HTML.

Beyond that, the reader was validated against the real Cppreference tagfile, and the possibility to repeat  `--input-tagfiles=<path>=<base-url>`, needed by the Antora extension, was checked by hand.

## Documentation

The "External docs" section of configuration/output.adoc documents the new option with an example. commands/inlines.adoc cross-references it from `@ref`, and the option reference is generated from `ConfigOptions.json`. No new page is added.

Closes #1072.


docs/modules/ROOT/pages/configuration/output.adoc
```diff
@@ -125,11 +125,23 @@ include::example$snippets/options/base-url/base-url.adoc[tags=!footer]
...
 
 TIP: A few projects use the tag file in a less obvious way: they read it from their own site-generation scripts to build a navigation tree or a search index from the symbol list, rather than maintaining one by hand.
 
-NOTE: The inverse direction (consuming tag files emitted by other projects so that your pages link to *their* symbols) is handled by the xref:extensions/antora.adoc#antora-cpp-tagfiles-extension[`antora-cpp-tagfiles-extension`]. Point it at one or more tag files in your Antora playbook, and references like `boost::asio::io_context` resolve to the Boost.Asio site automatically.
+xref:configuration/reference.adoc#input-tagfiles_option[`input-tagfiles`] is the other direction: it reads the tag files of documentation sets you do not build, so a reference to a symbol MrDocs never extracted still becomes a link. Each key is a tag file, each value the base URL that set is published under:
```
alandefreitas, line 132: Is **input**-tagfiles that go in **the other direction** really an "output" option?

docs/modules/ROOT/pages/configuration/output.adoc
```diff
@@ -125,11 +125,23 @@ include::example$snippets/options/base-url/base-url.adoc[tags=!footer]
...
+
+.`mrdocs.yml`
+[source,yaml]
+----
+include::example$configuration/input-tagfiles.yml[]
+----
+
+With that in place, `@ref std::vector` in a doc comment links to Cppreference instead of rendering as plain code. A name one of these sets documents resolves the way every other reference does, from the scope it is written in outward, so a name relative to its enclosing scope reaches an external symbol just as it would reach one of yours. What never happens is shortening: `vector` reaches `std::vector` only from inside `std`, because the shorter the name the likelier it is to collide with something unrelated in a foreign set, and a reference that silently points at the wrong page is worse than one that points nowhere. A name no scope accounts for is left as code.
```
alandefreitas, line 140: cppreference or C++ Reference with link?

alandefreitas, line 140: This has to be followed by output example, like all other options explained in this section. Without an example in form of narrative, it becomes redundant with the reference section.

alandefreitas, line 140: > What never happens is shortening

These negatives are very AI-ish. I'm sure this didn't cross the reader's mind at this point. There's no reason to negate this or provide any details about edge cases at this point because we haven't even provided the example of the general case yet.

docs/modules/ROOT/pages/configuration/output.adoc
```diff
@@ -125,11 +125,23 @@ include::example$snippets/options/base-url/base-url.adoc[tags=!footer]
...
+
+NOTE: A tag file only says where the pages of that set are, so MrDocs can resolve no more than the set publishes. Cppreference, for one, gives `namespace std` itself no page, so `@ref std::chrono` stays plain text while `@ref std::chrono::duration` links. Either way, it is a reference that resolved to nothing, and xref:configuration/reference.adoc#warn-broken-ref_option[`warn-broken-ref`] reports it as one.
+
+NOTE: For an Antora site there is also the xref:extensions/antora.adoc#antora-cpp-tagfiles-extension[`antora-cpp-tagfiles-extension`], which adds a `cpp:` macro and bundles the Cppreference tag file. The two are complementary, and reach different text: the extension resolves the names you write in your own pages, `input-tagfiles` resolves the ones inside doc comments, for every generator rather than for Asciidoctor alone.
```
alandefreitas, line 144: Out of scope (even as a note)

docs/modules/ROOT/pages/configuration/output.adoc
```diff
@@ -125,11 +125,23 @@ include::example$snippets/options/base-url/base-url.adoc[tags=!footer]
...
+
+.`mrdocs.yml`
+[source,yaml]
+----
+include::example$configuration/input-tagfiles.yml[]
+----
+
+With that in place, `@ref std::vector` in a doc comment links to Cppreference instead of rendering as plain code. A name one of these sets documents resolves the way every other reference does, from the scope it is written in outward, so a name relative to its enclosing scope reaches an external symbol just as it would reach one of yours. What never happens is shortening: `vector` reaches `std::vector` only from inside `std`, because the shorter the name the likelier it is to collide with something unrelated in a foreign set, and a reference that silently points at the wrong page is worse than one that points nowhere. A name no scope accounts for is left as code.
+
+NOTE: A tag file only says where the pages of that set are, so MrDocs can resolve no more than the set publishes. Cppreference, for one, gives `namespace std` itself no page, so `@ref std::chrono` stays plain text while `@ref std::chrono::duration` links. Either way, it is a reference that resolved to nothing, and xref:configuration/reference.adoc#warn-broken-ref_option[`warn-broken-ref`] reports it as one.
```
alandefreitas, line 142: Out of scope (even as a note)

include/mrdocs/Corpus.hpp
```diff
@@ -135,6 +137,26 @@ class MRDOCS_VISIBLE
...
+
+        A name no enclosing scope accounts for is not matched against
+        a longer one: `vector` reaches `std::vector` only from inside
+        `std`, exactly as it would for a symbol of this corpus.
+
+        @param context The context the name is written in.
+        @param name The name of the symbol to look up.
+        @return The URL, or nothing if no tagfile documents the name.
+    */
+    std::optional<std::string>
```
alandefreitas, line 157: Shouldn't this use Expected to match the convention for `lookup`?

src/mrdocs/Corpus.cpp
```diff
@@ -788,6 +813,42 @@ lookup(SymbolID const& context, std::string_view name)
     return lookupImpl(*this, context, name);
 }
 
+std::optional<std::string>
+Corpus::
+externalUrl(SymbolID const& context, std::string_view name) const
+{
+    std::optional<std::string> result;
+    std::string_view const scopeQualifier = "::";
+    if (name.starts_with("scopeQualifier"))
```
alandefreitas, line 822: comparing with a literal string instead of the variable? this works?

src/mrdocs/Metadata/Finalizers/DocCommentFinalizer.cpp
alandefreitas, line 721: This is intentional and makes the comment about `warn-broken-ref` in the documentation incorrect. The reason we skip this is that users tend to know what they're doing in this case and they typically just want code formatting here. Doxygen libraries come with this a lot where people @ref they're own library and std freely just for the formatting.

include/mrdocs/Config.hpp
```diff
@@ -139,12 +139,43 @@ class MRDOCS_DECL
...
     */
     std::vector<std::string> unknownConfigKeys;
+
+    /** A deprecated option this configuration sets, and its advice.
+
+        Populated during @ref normalize, which is the pass that sees each
+        option's properties, and surfaced by
+        @ref reportDeprecatedOptions.
+    */
+    struct DeprecatedOption
```
alandefreitas, line 164: This is not our first deprecated option. What was the model we had before, and why was it insufficient?

tests/golden/fixtures/config/input-tagfiles/other.tag.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
alandefreitas, line 1: Is this generated by doxygen or us? We need to test both cases to be sure.

alandefreitas (commented):
This feature is amazing. This is a feature I've wanted for a while, and the layout is right. Resolution happens in the finalizer, which is good, and the reference node has the href.

Is this Doxygen tagfile format good enough? Do we need our own format?

We should load the tagfiles before extraction so that a typo in a tagfile path fails the build before extracting the whole AST, which could take one hour for a large library such as LLVM.

The `tagfile` to `output-tagfile` transfer depends on declaration order. This can be solved by doing it once in `Config::normalize`. Then the special case in `normalizeString` and the comment can go.

docs/modules/ROOT/pages/configuration/output.adoc
gennaroprota: Oops, moved to the Inputs page.

docs/modules/ROOT/pages/configuration/output.adoc
gennaroprota: IIRC, the site name was mentioned once, on the main page, as "Cppreference", but I can't find that now, so I'm switching to "cppreference.com", with a link.

gennaroprota: Done (I assume you meant "input example").

gennaroprota: Removed. The section now leads with what the option does and the example; the sentence after it on how a name resolves is stated positively.

docs/modules/ROOT/pages/configuration/output.adoc
gennaroprota: Removed.

include/mrdocs/Corpus.hpp
gennaroprota: Yes, better.

src/mrdocs/Corpus.cpp
gennaroprota: Great catch! Thanks! It was a string literal initially, then I introduced the variable but forgot to remove the quote characters.

src/mrdocs/Metadata/Finalizers/DocCommentFinalizer.cpp
gennaroprota: Yes. The skip stays. And I removed the documentation note that contradicted it.

include/mrdocs/Config.hpp
gennaroprota: The previous model was erroneous, in that it tried to warn inside `normalizeInteger` but that never printed, because option parsing runs with the reporting level forced to errors-only. Deprecated options warn now for the first time.

tests/golden/fixtures/config/input-tagfiles/other.tag.xml
gennaroprota: By us. I noticed that we write `encoding="UTF-8"`, with double quotes, but Doxygen writes `encoding='UTF-8'`, with a space before `?>`. All forms are covered now.

PR #1304, opened 2026-09-16
UI refresh follow-ups: title centering, copy, mobile contents

julioest:
Follow-ups to the comic design refresh (`be06d44fb`), found by measuring the built site against the Figma.

## What changed

**`fix(ui): centre the docs page title`** — `margin: 1.5rem 0` on `h1.page` is (0,3,1) and outranked `.doc > * { margin-inline: auto }` at (0,1,0), so the title kept the 779 measure but anchored left, off the content band's axis by (content width − 779) / 2. 15.5px at 1440, 255.5px at 1920, and 0 below 1169. The losing rule's own comment says the title should be "centred, which keeps the text under the band's converging rays".

**`fix(website): restore Figma copy`** and **`fix(website): keep maintainer wording on the cards`** — together these put the row 2 and row 4 panel subtitles on the Figma text and leave the two card descriptions alone. The first commit initially restored all four; the two card strings were maintainer edits (`e9f847d8a`, `315232bea`) carrying C++ terminology, so the second narrows it back. The mockup carries the older wording for those two cards and needs updating.

**`fix(website): code-style noexcept, name the hero snippet`** — `noexcept` sat as body text beside a code-styled `[[noreturn]]` in the Attributes & Specifiers subtitle; both are C++ keywords. The hero panel is the `is_prime` snippet but its tab read "C++" while row 3 shows the same file as `is_prime.cpp`. Both raised in review. The Figma marks only `[[noreturn]]` as code and uses the same snippet twice, so it needs updating on the first point; the duplication itself is left as a design decision.

**`fix(website): hero and principles rhythm`** — `.principles` inherited the default 80px section padding against Figma's 164, pulling the whole block up ~82px. The hero's desktop rhythm now follows the frame's uniform gap32 stack. Landmarks land within 2px of the frame.

**`feat(ui): clip long mobile contents lists`** — the embedded contents list ran full length above the page title: `generators/reference` opened with 3303px of links and put the `h1` at y3588 against a 285 baseline.

**`fix(website): name LaTeX math in the SFINAE row`** — the sqrt example renders two LaTeX fragments (`\f$…\f$` in the brief, `$…$` in the description) but nothing said math was a feature. Measured, the two formulas are 71×17 and 81×16 in a 638×1194 panel, and `O(log value)` renders as italic serif that passes for ordinary emphasis. Math now sits in the row's list of claims; the subtitle wraps to three lines where it took two, so the row is ~52px taller.

**`docs(ui): note the section mark is unobservable`** — comment only.

## Before / after

Docs page title at 1920. The band and body copy are identical in both; the guides mark the band's centre and the title's, so the 256px offset and its correction are measured rather than eyeballed.

docs page title, before and after

Mobile contents at 402, `generators/reference`.

mobile contents, before and after

Landing hero and principles rhythm at 1440, with the gap from the hero band to the heading measured against the frame's 164.

landing hero and principles rhythm, before and after

## Mobile contents, in detail

Measured across the built site first: 3745 pages carry a contents list and the median is 2 entries. Six exceed the threshold. The always-open block the frame draws (316:334668, 402×206) is correct for the rest and is untouched.

The threshold is derived rather than picked — the list is ~88px of chrome plus ~30px per entry with the title directly beneath, so the title lands at roughly 285 + 88 + 30n against ~700px of usable viewport on a 402×874 phone, clearing it only to n = 10.

From 11 entries the list shows its first five top-level rows, with a chevron to reveal everything and a per-section caret to open one section in place. Both controls reuse existing theme idioms: `chevron.svg` at panel level as the nav drawer's section chips use it, `caret.svg` at item level as the nav tree does. Labels still navigate.

## Verification

- `gulp lint` clean.
- Desktop ≥1024 unchanged: embedded list hidden, rail 240, `h1` top 269, 0px to the band.
- Geometry diffed against the design at 402 / 768 / 992 / 1200 / 1440 / 1920 in both themes, against the built theme rather than injected CSS.
- Mobile hero byte-identical before and after the landing change — 0 changed nodes inside `<header>` at 402 and 768.

Regression expectations live in a rig outside this repo, so they do not run in CI.


alandefreitas:
A few glitches or things that look off:

* The menus on mobile

image

* The Y in the landing page

image

* The readability of typewriter font in long paragraphs

image

* Diagrams using different fonts

image

* Different headings in previews

image

* Borders of tabs and opacity of tabs options

image

* Previews in configuration/reference have different heading styles and code blocks are broken (dark grey bg / black font / no highlighting)


image

* Tables in generators/reference and extensions/library reference are weird. The figma template has complementary colors but these are typically used in HTML to represent broken links so it looks like the links in the table are broken.

image

* The prev/next bar is misaligned with the text

image

julioest:
Pushed 67e966b2a for the unstyled code blocks in the config reference previews.

Root cause was in `docs/extensions/config-options-reference.js`, not the theme.
That file renders previews with a standalone Asciidoctor instance, which inherits
no attributes from the playbook. Without `source-highlighter` it emits

```html
<pre class="highlight"><code class="language-cpp" data-lang="cpp">
```

where the theme's highlight bundle selects on `pre code.hljs[data-lang]`, so those
blocks never got highlighted. The hand-built blocks in the same function already
hardcode `highlightjs`/`hljs`; the converted path just didn't match. Fix is that
one attribute.

Scope: 116 of the 216 blocks on `configuration/reference.html` were affected. No
other page was — the remaining 13 pages with previews had none. After the fix,
a sweep of all 3,760 built pages finds 5,081 `data-lang` blocks and zero missing
`hljs` (5,081 = 4,965 already working + the 116), so nothing outside those 116
changed.

PR title widened accordingly; the rest of the review points are still open.

julioest:
@alandefreitas 8 of 9 fixed. Couple of question for you, on the Y and readability of typewriter font.

**The Y.** CAN'T REPRODUCE. Tried it across a range of viewports, in both themes, on live as well as the preview, and it renders whole every time. What browser and viewport size were you on?

**Readability of typewriter font in long paragraphs.** OPEN. Is this because of the background behind certain long paragraphs? Or does the font need to be styled a little better (line-height, spacing, etc)?

**The menus on mobile.** FIXED [ae7a01305](https://github.com/cppalliance/mrdocs/commit/ae7a01305) [ebde211f1](https://github.com/cppalliance/mrdocs/commit/ebde211f1). Panel is full height at the frame's sizes, and the burger swaps to a cross so it can be closed. The docs menu was undersized too; both match now.

**Different headings in previews**, and the heading styles in configuration/reference. FIXED [72ede9a7f](https://github.com/cppalliance/mrdocs/commit/72ede9a7f) [1cad1d426](https://github.com/cppalliance/mrdocs/commit/1cad1d426). Previews rendered a level too shallow, so sections took the styling meant for symbol names, and the card's own type scale collided with the display face's stroke. Both follow the landing spec now.

**Borders of tabs and opacity of tabs options.** FIXED [871dd769c](https://github.com/cppalliance/mrdocs/commit/871dd769c). Rebuilt to the component; the opacity had faded the label along with the surface, where the frame tints only the background. One divergence: dark inactive tabs take the page surface, not the frame's solid blue, which left the label too faint.

**Code blocks broken in configuration/reference.** FIXED [2edc19567](https://github.com/cppalliance/mrdocs/commit/2edc19567). The generator, not the theme: it renders previews with a standalone Asciidoctor instance that wasn't told to highlight. Same cause for the grey background and black text.

**Diagrams using different fonts.** FIXED [dfac0063b](https://github.com/cppalliance/mrdocs/commit/dfac0063b) [11eb3b5ad](https://github.com/cppalliance/mrdocs/commit/11eb3b5ad). Mermaid was initialised with Roboto, the fallback at the end of the body stack, so diagrams rendered in the face the prose falls back to. Dark mode was worse than the font, so that went too: node colours were authored as `classDef`, which Mermaid emits inline and no stylesheet can reach, and dropping those lets the theme own them. Adds `--mr-green`.

**Tables in generators/reference and extensions/library.** FIXED [41abcbaaa](https://github.com/cppalliance/mrdocs/commit/41abcbaaa). The inline code was the problem, not the links: every code span in a cell was painted the label colour, and none are links. The links were separately too faint and differed from adjacent code by colour alone, so they're darker and underlined now, and the cell-wide hover is gone.

**The prev/next bar is misaligned with the text.** FIXED [f445ad7d0](https://github.com/cppalliance/mrdocs/commit/f445ad7d0). Explicit side margins outranked the rule centring `.doc` children, so it anchored left instead. Aligned at every width now.

PR #1306, opened 2026-09-17
fix: user errors do not invite a bug report

### LINKED ISSUE #1116: MrDocs error message confusing between internal and user errors
When there is any kind of error in the mrdocs tool execution, it produces:
```
  An issue occurred during execution.
  If you believe this is a bug, please report it at https://github.com/cppalliance/mrdocs/issues
  with the following details:
      MrDocs Version: 0.8.0 (Build: ad5dbb241f9bfa2088e0163a7396f45ad586f9fa)
      Reported From: `src/lib/Metadata/Finalizers/DocCommentFinalizer.cpp` at line 1721
```

This happens regardless if this was an internal error, or if this is a user error, for example an undocumented entity warning, with warnings as errors enabled.

Ideally this kind of message should only be produced for internal errors, and we should trust that user errors are not mrdocs' problem.

gennaroprota:
Every error-level message in MrDocs printed the internal source location and an invitation to file a bug report. For instance, a missing configuration file produced one, and with `--warn-as-error` so did every undocumented symbol, thus a reader was asked to report a MrDocs defect for a mistake in their own input.

`report::bug` is now the only function that does that: it prints the message followed by the version and the source location a bug report needs, while `error`, `fatal`, `warn` and the rest print the message alone.

Other defects surfaced while going through the reporting code and are fixed in their own commits.

## Changes

- **Source**: `report::bug` now reports MrDocs defects, with the bug-report details suppressed everywhere else; quoted errors carry their reason without the raising location; `source-root`'s bare-placeholder default resolves to the configuration file's directory; the release-build exception report is compiled in again; the overload-set diagnostic names overload sets rather than enums; the unused `getLevel` is gone.
- **Tests**: a unit test for the `source-root` default.

## Testing

tests/unit/Config.cpp covers the `source-root` default: it writes a minimal mrdocs.yml into a temporary directory and asserts the value of `source-root` comes back as that file's directory, which the old resolver could not do.

The diagnostics themselves are not tested, as that seemed overkill.

## Documentation

No page describes the diagnostics or reproduces their wording, so nothing needed updating. `report::bug` is public API and carries its doc-comments, which include when to reach for it instead of `error`.

Closes #1116.


alandefreitas:
The historical motivation is that when MrDocs was a new tool, errors were usually not user errors. They were actually MrDocs failing to extract the symbols. That's why we invited a bug report. If you notice, the message doesn't say this is a user error or a MrDocs error. The message used to say that if the user **thinks** this is a MrDocs error, they could report it (because, at least at the time, there was a high probability it was a MrDocs error, not a user error).

This historical rationale matters because the whole point of the previous behavior was that we didn't know whether it was MrDocs or a user. There's no point in a new command to report bugs because we don't know if something is a bug or a user error. 

If we **know** there's a bug, we shouldn't be reporting it. We should fix the bug. If fixing the bug is too hard, we should never take this path, and the feature is simply unsupported. Thus, there's no need to `report::bug`.

And as far as user errors that are actually MrDocs errors are concerned (the original motivation for these messages), if we now think that, for an individual error, the probability that it's the user's fault is much higher than the probability that this is MrDocs' fault, then we can simply remove the part of the message that suggests opening a pull request. Then the user will figure it out in the rarer case when it's not really his fault.

gennaroprota:
> If we **know** there's a bug, we shouldn't be reporting it. We should fix the bug. If fixing the bug is too hard, we should never take this path, and the feature is simply unsupported. Thus, there's no need to `report::bug`.

It's not that we know in advance that there's a bug, just that, sometimes, the code can detect a bug. For instance, if you get past the `if` here:

```
void
ASTVisitor::
addMember(EnumSymbol& I, Symbol const& Member) const
{
    if (auto const* U = Member.asEnumConstantPtr())
    {
        addMember(I.Constants, *U);
        return;
    }
    report::bug("Cannot push {} of type {} into members of enum {}",
        Member.Name,
        mrdocs::toString(Member.Kind),
        I.Name);
}
```

(we might as well use an assertion for such cases, as long as it yields a clear error message).

